PR #181, opened 2018-02-07
Made requested UI changes for landing page

metincansiper:


sacdallago:
I agree with @d2fong , while comments are good practice, if the variable names and code structure are good, they may slow down reading the code. In this case, I think it's safe to remove them :) other than that: looks good to me.

Issue #176, opened 2018-02-02
Design Discussion - Factoid Landing Page


d2fong:
## General Ideas
Vertical split of text
A description of what Factoid
Links to databases (pathway commons)/other projects
People who worked on it (different labs)
More distinguishable form of choosing mode (easy to miss that there are different modes)
About page
Github links
Links to Reach, cytoscape
General Idea - More context as to what Factoid is

## Goals and Key Concepts
- What will entice a user to actually use Factoid if they go the landing page?
- Can we come up with a call to action to get them to use Factoid?
- Who can we reasonably expect to visit the Factoid landing page?
    - publishers?
    - random people?
    - scientists?
- Knowing the target audience would help us focus on the things we need for this page

d2fong:
https://www.landingfolio.com/category/gallery/application/ for landing page ideas

d2fong:
Resources to read:

Landing Page Optimization: The Definitive Guide to Testing and Tuning for Conversions, 2nd Edition

Will help gives us context as to what a landing page should do

d2fong:
[new.doc.2018-02-02.13.24.40-20180202134850.dragged.pdf](https://github.com/PathwayCommons/factoid/files/1690466/new.doc.2018-02-02.13.24.40-20180202134850.dragged.pdf)
[new doc 2018-02-02 13.24.40-20180202134850 (dragged) 2.pdf](https://github.com/PathwayCommons/factoid/files/1690467/new.doc.2018-02-02.13.24.40-20180202134850.dragged.2.pdf)

sacdallago:
More mockups: 
[Project - Drawing 1.pdf](https://github.com/PathwayCommons/factoid/files/1705528/Project.-.Drawing.1.pdf)
[Project - Drawing 1_1.pdf](https://github.com/PathwayCommons/factoid/files/1705529/Project.-.Drawing.1_1.pdf)


The proportions are a bit off, apologies for that!

The idea is:

- Have users land on a page where there is a sentence with `"Create document" [from text] // [blank]`, where `[]` are buttons, whereas `""` is just normal text put nicely in-line with the buttons. - The buttons are "color-coded", as you see from the other screen: this gives the users a clue about what they are doing when the text area expands, and that they still have the option to create a blank document.
- Giving color to the buttons and leaving the rest black/white should attract the user's attention there, which is what we want.

Opinions welcome :)

metincansiper:
[...]

PR #184, opened 2018-02-16
Landing page UI updates

metincansiper:
Made the changes suggested in #176. You can see the results in the following screencast (http://recordit.co/g5b0Q6tiAh). Seems like the quality of screencast is not good sorry for that.


src/client/components/multi-form-editor/landing-page.js
```diff
@@ -1,36 +1,224 @@
...
 
 class LandingPage extends React.Component {
+  constructor( props ){
+    super( props );
+
+    this.state = {
+      textAreaEnabled: false,
+      submitting: false,
+      textAreaAnimating: false,
+      textAreaHeight: '20em'
```
sacdallago: Please don't mix styles with JS. Use CSS instead.

src/client/components/multi-form-editor/landing-page.js
```diff
@@ -1,36 +1,224 @@
...
+    };
+  }
+
+  createDoc(){
+    let text = ReactDom.findDOMNode(this).querySelector('.landing-page-text').value;
+
+    let makeRequest = () => fetch('/api/document', {
```
sacdallago: @d2fong @maxkfranz  are we using brokers? I would argue that having direct calls to endpoints in components is long-term not too maintainable. Opinions?

src/client/components/multi-form-editor/landing-page.js
```diff
@@ -1,36 +1,224 @@
...
+          h('button.landing-page-fill-text-btn', {
+            className: makeClassList({
+              'landing-page-left': this.state.textAreaEnabled,
+              'landing-page-no-right-border': !this.state.textAreaEnabled,
+              'landing-page-no-bottom-border': this.state.textAreaEnabled,
+              'landing-page-round-tl': true,
+              'landing-page-round-bl': !this.state.textAreaEnabled,
+              'landing-page-round-tr': this.state.textAreaEnabled
+            }),
+            onClick: () => {
```
sacdallago: this is quite a big function. I'd rather have it at class level (a.k.a.: at the same level of `render`). And then call it here via `onClick={this.something}`.

src/client/components/multi-form-editor/landing-page.js
```diff
@@ -1,36 +1,224 @@
...
+
+              // if there is an animation for text area do not allow click event
+              if (this.state.textAreaAnimating) {
+                return;
+              }
+
+              let textArea = ReactDom.findDOMNode(this).querySelector('.landing-page-text');
+
+              let toggleTextAreaState = (() => {
```
sacdallago: Define at class level

src/client/components/multi-form-editor/landing-page.js
metincansiper: I think I just moved this code segment from old ```document-filler.js``` page (https://github.com/PathwayCommons/factoid/blob/alpha/src/client/components/document-filler.js#L27)

src/client/components/multi-form-editor/landing-page.js
metincansiper: I define CSS values in CSS files and use adding/removing classes for changing them dynamically. However, this is a different case. As you most probably realized the text area is opened and closed by animating through its height. Therefore I need to store its height before it is closed to use when it is opened back (Note that the height of text area can be changed by users). Here is the code line where how the height is supposed the change during animation is defined (https://github.com/metincansiper/factoid/blob/landing-page-ui/src/client/components/multi-form-editor/landing-page.js#L107). While opening the text area it should animate between 0 and the height of text area that is saved before closing it. And the initial value for the open height (I mean the height that it has when it is open) of it is defined at the code segment that you highlighted. If I defined the initial height of text area through CSS, it would be overridden during the animation. Therefore, this is the way that sounds best to me. If you suggest a better way I will be glad to hear and discuss.

src/client/components/multi-form-editor/landing-page.js
sacdallago: Thanks @metincansiper, let's see what @d2fong @maxkfranz think

src/client/components/multi-form-editor/landing-page.js
sacdallago: MMhh... I understand but can't you toggle-add a css class with the height when opened and use animation on height with ease on the parent css class? A small most likely not working, offline written example to illustrate the idea:

```javascript
<element className={this.state.textAreaActive===false ? "textArea" : "textArea open" }
```

```css
.textArea {
   ....;
   max-height: 0;
   transition: height 1s;
  overflow: hidden;
}
.textArea > .open {
   max-height: 6em;
}
```

Or: https://codepen.io/LFeh/pen/ICkwe

d2fong: Yes @sacdallago your approach is the convention that this code base uses for state based css

src/client/components/multi-form-editor/landing-page.js
d2fong: I think we use direct calls in the code base so I assume this is fine...

src/client/components/multi-form-editor/landing-page.js
metincansiper: @sacdallago Thank you for the suggestion. However, I have 2 main concerns about it. One is more important while the other one would not be that much important.

The important concern of me is determining a value for ```max-height``` for the text area when it is open (I mean the exact value to be written instead of ```max-height: 6em;``` in your example). As I mentioned users can increase the height of text area. Therefore, we do not know what would be the maximum height that it should have when it is open. In this case one would think of setting an arbitrary large value for the ```max-height``` however, in that case the animation would be destroyed. I mean in case of ```max-height``` is much more than ```height``` users will observe that the accordion will open immediately with nearly no transition effect while opening the text area. While closing it they will observe that there will be no effect for a while (until the ```max-height``` become equal to ```height```) then again it will be closed immediately with nearly no transition effect. Also, even the ```max-height``` that we preset would not be enough. 

The minor concern is that I need to do something before the animation in case text area is being opened and do it after the animation is completed while it is being closed (You can see https://github.com/metincansiper/factoid/blob/landing-page-ui/src/client/components/multi-form-editor/landing-page.js#L116). For this reason ```promise``` support of ```animejs``` library was a good facility. I can try to use ```transitionend``` event instead but I suspect that it would decrease the readability and would require using some not preferred practices. 

As a final comment, your concern about the initial implementation of me was that I was keeping a property in state object that is assigned to```20em``` initially and you considered it as mixing JS with styles. However, from a different perspective it just keeps a state (saying what should the height of text area be) to be used as a parameter while having a call of ```animejs``` library.

What do you think about all of these?

@d2fong I do not agree that approach of @sacdallago was strictly on use in this code base for similar animations. You can see https://github.com/PathwayCommons/factoid/blob/alpha/src/client/components/document-filler.js#L45. Actually, I needed to move this code segment to landing page since document filler's content is moved to there. That code segment would be totally removed according to the discussion on https://github.com/PathwayCommons/factoid/issues/176. However, it would needed to be updated accordingly in case it will not be removed.

metincansiper:
@d2fong how should the new URL's for creating a document from text be? I mean should 'document' be replaced by 'edit' in URL's?

And currently "New document" button is not performing successfully. What should be the path to follow when it is clicked? Currently, it navigates to ```/debug/new-document``` and redirected to ```/edit0/${id}/${secret}``` from there.

Thank you

metincansiper:
@d2fong @sacdallago @maxkfranz BTW I removed the extra interface and directly navigated to the document when "Next" button is clicked. Made the other minor changes that are discussed in scope of this PR as well. The only thing remaining for this PR is updating the URLs and will do it once the last question of me is replied.  You can review if you have time.

src/client/components/document-filler.js
```diff
@@ -1,100 +0,0 @@
-const React = require('react');
-const h = require('react-hyperscript');
-const { Link } = require('react-router-dom');
```
d2fong: Can you please un-delete this file.  It will cause more merge conflicts down the road.

src/client/router.js
```diff
@@ -6,7 +6,6 @@ const uuid = require('uuid');
 const Home = require('./components/home');
 const Editor = require('./components/editor');
 const Debug = require('./components/debug');
-const DocumentFiller = require('./components/document-filler');
```
d2fong: Un-delete this line to avoid merge conflicts down the road.

src/client/router.js
```diff
@@ -83,12 +82,6 @@ let routes = [
       } );
     }
   },
-  {
-    path: '/debug/new-document/fill',
```
d2fong: Un-delete this line to avoid merge conflicts down the road.

src/styles/document-filler.css
```diff
@@ -1,57 +0,0 @@
-.document-filler {
-  width: 40em;
```
d2fong: Un-delete this line to avoid merge conflicts down the road.

src/styles/index.css
```diff
@@ -12,7 +12,6 @@
 @import "./home.css";
 @import "./debug.css";
 @import "./document-linkout.css";
-@import "./document-filler.css";
```
d2fong: Un-delete this line to avoid merge conflicts down the road.

metincansiper:
@d2fong Is the only reason why you want me to undelete document filler and and undo all other related changes is preventing conflicts with #189? If that is the only reason I think it is a better approach to solve merge conflicts while merging #189.

I think the main conflict will be adout deleting ```document-filler.js``` because I deleted it and you modified it in #189. However, we can think that the changes you made on that file will not be needed because that file is no more used. Either you can remove that file from that branch directly or you can prefer to delete it as a way of solving conflict during the merge. 

For the other ones I am not sure if they will cause conflicts because I suppose on your branch you did not changed that files and the changes on them would be automatically merged (Maybe I am wrong, not sure). But I even it gives conflict they must be small conflicts and would be better to solve them during merging.

On the other hand if I do the changes suggested here undoing them again will be needed and so I think that the other way would be better. 

What do you think?

d2fong:
ok fine lets just merge it.

PR #198, opened 2018-03-05
Landing page

d2fong:
various sections:

screen shot 2018-03-05 at 5 10 22 pm
screen shot 2018-03-05 at 5 10 15 pm
screen shot 2018-03-05 at 5 09 50 pm
screen shot 2018-03-05 at 5 09 40 pm


maxkfranz:
So the skeleton is done but we just need to fill in the content and maybe make some minor styling tweaks.

I'm merging this now so we can all look at it on the unstable CI server: unstable.factoid.baderlab.org

We can have another PR once we've iterated a bit more in the landing-page branch.

sacdallago:
Thanks @maxkfranz

Issue #201, opened 2018-03-06
Use PubChem results for grounding small molecules in entity tooltip search  


jvwong:
## Description of new feature
Add PubChem as service for entity grounding in tooltips (small molecules). 

## Motivation for new feature
Currently, ChEBI is used as one of the services for small molecule grounding in the entity tooltip. This is because of the conciseness and utility of the synonyms they provide. However, REACH text mining returns PubChem IDs and Factoid attempts to perform a mapping between the two. Just use PubChem alone.  

Replacing #195

IgorRodchenkov:
Please, let's be clear whether we're talking about "PubChem Compound" or "PunChem Substance" IDs.
These are different databases, which, unfortunately, both have integer IDs (no official prefix; though unofficially some people use CID: or SID: prefixes...). An ID usually point to unrelated, different molecules in those databases.

maxkfranz:
After this issue is addressed, we need to address this one: 

Pubchem has too many synonyms #202

maxkfranz:
@IgorRodchenkov That's a good point, but we also store InChi -- so we can still uniquely identify the compound, right?

IgorRodchenkov:
InChiKey, you mean? Yes, this is much better ID.

metincansiper:
@jvwong @fdurupinar @maxkfranz @IgorRodchenkov 

I tried to debug the code. I see that in this [code segment](https://github.com/PathwayCommons/factoid/blob/master/src/server/routes/api/document/reach.js#L112) the desired things are done if ```ground.namespace``` is equal to ```'chemical'``` but as far as I see that value is eqaul to ```'pubchem'``` not ```'chemical'```. 

I experimented changing it as ```'pubchem'``` and see that the buggy behaviour is fixed, I got the following result:

<img width="903" alt="pubchem" src="https://user-images.githubusercontent.com/6534865/37065180-ab36a8b6-2154-11e8-8c69-d56504093753.png">

Should the ```case``` in that line be updated as ```case 'pubchem':```? 

I think that the original issue (#195) was about this bug. Should ChEBI layer still be removed and just PubChem be used alone as it is addressed in this issue if this change solves the bug safely?

metincansiper:
Then I suppose that the only thing needed for this issue is changing the case that I mentioned in above comment as:

```js
case 'pubchem':
              return pubchem.get( q );
```

I am attaching the results to compare the old (ChEBI is used as well) and new (just PubChem) implementation for the example in #195
[...]

PR #203, opened 2018-03-07
Use PubChem results for grounding small molecules in entity tooltip search

metincansiper:
Created to solve #201


maxkfranz:
I'm going to cherry-pick this into unstable, which is the dev branch.  The alpha branch is pretty much for some experimentation.  It's probably not well named and should be removed eventually...

PR #205, opened 2018-03-08
log error message when the reach provider fails

d2fong:
- clean up require statements

- not sure what to return from fillDoc if there is an error, should I just return an empty doc?

### DIFF
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -82,6 +82,7 @@ http.post('/', function( req, res ){
     .catch( e => {
       logger.error('Could not fill doc from text: ', `text: ${text}`, e);
       res.sendStatus(500);
+      throw e;
     } )
   );
 });
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -78,15 +78,15 @@ http.post('/', function( req, res ){
     .then( doc => fillDoc( doc, text ) )
     .then( runLayout )
     .then( getDocJson )
-    .then( json => res.json( json ) )
+    .then( json => {
+      logger.info(`Created new doc ${json.id}`);
+      return res.json( json );
+    } )
     .catch( e => {
+      logger.error(`Could not fill doc from text: ${text}`);
+      logger.error('REACH exception thrown :', e.message);
+      res.sendStatus(500);
 
-      if( e instanceof Promise.TimeoutError ){
-        res.sendStatus(504);
-      } else {
-        logger.error('Could not fill doc from text: ', `text: ${text}`, e);
-        res.sendStatus(500);  
-      }
       throw e;
     } )
   );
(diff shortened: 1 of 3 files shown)


maxkfranz:
I don't think it should return an empty doc.  We could return an error json, but I think it would be best to just send an http error for now: https://expressjs.com/en/4x/api.html#res.sendStatus

maxkfranz:
Probably 500?

sacdallago:
Isn't this error generated IF there is an unexpected response from REACH? Aka: the response is generated, it's just not what we expect? And if so: isn't it impossible for the user to know what she/he has to change in order to _make it work_? If that is the case, not allowing the user to just _move on_ even if on an empty document might be frustrating to the user.

d2fong:
Ok I send a 500 error when there there is a problem with creating a doc from reach.

sacdallago:
I think this is a good start, but maybe we want to merge this in with all the other cases listed here: #207

src/server/routes/api/document/index.js
```diff
@@ -73,6 +79,10 @@ http.post('/', function( req, res ){
     .then( runLayout )
     .then( getDocJson )
     .then( json => res.json( json ) )
+    .catch( e => {
+      logger.error('Could not fill doc from text: ', `text: ${text}`, e);
+      res.sendStatus(500);
```
maxkfranz: Need to rethrow also

d2fong:
Ok the build CI check has passed.  Weird that it failed before, but succeeded after I restarted the build.

maxkfranz:
Travis doesn't seem to work well with RethinkDB.  Sometimes it just crashes.

d2fong:
So can I merge this?

src/server/routes/api/document/index.js
```diff
@@ -73,6 +79,16 @@ http.post('/', function( req, res ){
     .then( runLayout )
     .then( getDocJson )
     .then( json => res.json( json ) )
+    .catch( e => {
+
+      if( e instanceof Promise.TimeoutError ){
+        res.sendStatus(504);
+      } else {
+        logger.error('Could not fill doc from text: ', `text: ${text}`, e);
```
maxkfranz: Did you text the output of this?  I'm not sure that Winston logs objects with a prototype like this properly.  Maybe it's best to use a single template string.  You could use `${e.toString()}` to guarantee you get a printable representation.

maxkfranz:
Looks fine but I'm not sure about line 87.  I would have merged yesterday but didn't have time to investigate 87

src/server/routes/api/document/index.js
maxkfranz: Maybe better to have the error as a separate logger.error() call

maxkfranz:
I've turned off Travis for older versions of Node.  It seems like 6 hangs sometimes on npm install.  This may help with some of the Travis noise.

We use Node 8 now anyway.

src/server/routes/api/document/index.js
sacdallago: I agree with the second call to error with just the response, first point needs testing, javascript is all but an exact science.

PR #210, opened 2018-03-13
Allow the Highligher component to highlight multiple terms

### LINKED ISSUE #206: Allow the Highligher component to highlight multiple terms
## Description of new feature

Change the multiplicity of the `term` prop in the Highlighter component.

Before: `term: 'foo bar'`
After: `terms: ['foo bar']` (highlight exactly 'foo bar') or `terms: ['foo', 'bar']` (highlight any 'foo' and any 'bar')

Each term should be highlighted in the search text, just as the single term is currently.

## Motivation for new feature

The main usecase of the highlighter is to show how well names match in the entity grounding UI.  When the user-typed name (term) is a space-separated, compound word, you can have very few highlights.  It would be useful to have partial matches, and allowing multiple terms in the Highlighter allows for this.

See also : Pubchem has too many synonyms #202

metincansiper:
Created to solve #206

### DIFF
--- a/src/client/components/element-info/entity-info.js
+++ b/src/client/components/element-info/entity-info.js
@@ -516,7 +516,6 @@ class EntityInfo extends React.Component {
     }
 
     let proteinFromAssoc = (m, searchTerms) => {
-
       return [
         h('div.entity-info-section', [
           h('span.entity-info-title', 'Organism'),
@@ -553,7 +552,6 @@ class EntityInfo extends React.Component {
     };
 
     let chemFromAssoc = (m, searchTerms) => {
-
       return [
         h('div.entity-info-section', [
           h('span.entity-info-title', 'Formulae'),
--- a/src/client/components/element-info/entity-info.js
+++ b/src/client/components/element-info/entity-info.js
@@ -578,7 +578,7 @@ class EntityInfo extends React.Component {
       let complete = stage === STAGES.COMPLETED;
       let highlight = !complete;
       let searchStr = highlight ? s.name : null;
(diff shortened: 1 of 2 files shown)


src/client/components/element-info/entity-info.js
```diff
@@ -577,12 +577,13 @@ class EntityInfo extends React.Component {
     let targetFromAssoc = (m) => {
       let complete = stage === STAGES.COMPLETED;
       let highlight = !complete;
-      let searchTerm = highlight ? s.name : null;
+      let searchStr = highlight ? s.name : null;
+      let searchTerms = searchStr.split(' ');
```
maxkfranz: This is probably OK, but users sometimes accidentally type two spaces so `\s+` is safer.

metincansiper:
Sorry I missed that point. Now I updated the regex to split. Also, I realized that ```highlighter``` should skip the non-truthy values (I think they would only be empty strings in our case). It would be problematic when user put extra spaces to end of the input (e.g. "iron oxide   "). Removing the empty string from the end of ```terms``` array as well would solve this problem but as far as I see from the previous implementation ```highlighter``` is supposed to handle such cases. Therefore I made this change (https://github.com/PathwayCommons/factoid/pull/210/commits/dad53f3db99b4e07ec9cff73b5c9d7dc7c97b7f2#diff-2253f654ecb3903f42c03e1e4709367fR77) as well.

maxkfranz:
An alternative for that case is to trim the input string, but what you have is more general.  Thanks

Issue #202, opened 2018-03-06
Pubchem has too many synonyms


maxkfranz:
## Environment info

- Browser version (client) : N/A
- Node.js version (server) : N/A

## Current (buggy) behaviour

Results from Pubchem return many synonyms.  Here's an example: https://pubchem.ncbi.nlm.nih.gov/compound/24393#section=Depositor-Supplied-Synonyms&fullscreen=true

We need to display at least some synonyms on the search stage of the entity wizard, because the user needs to have feedback about matches within synonyms (indicated by a yellow highlight in the UI).

When the list of synonyms is very long, there is too much noise.  Many synonyms are not relevant and the user has to scroll a lot.  One entity can take up the whole height of the screen.

## Desired behaviour

Filter the synonyms list somehow so the overall per-entity list is concise and contains relevant names.

One possibility is to only show synonyms that are highly-ranked matches, i.e. ones that are close to what the user typed.  Those are ones that would tend to have highlights (i.e. exact substring matches).  There may need to be a hard cut-off in case many synonyms are highly-ranked.  We may also need to remove synonym display completely from the entity info shown for grounded chemicals.  Maybe we could just show the IUPAC name under the name the user typed instead of the whole list of synonyms.

## Minimum steps to reproduce

You can take a look at Pubchem data outside of Factoid.  Just do a search on the Pubchem site, e.g. https://www.ncbi.nlm.nih.gov/pccompound?term=iron

Then choose a chemical and look at its synonyms list.  The list is very long a lot of the time.

sacdallago:
To expand on 
> One possibility is to only show synonyms that are highly-ranked matches, i.e. ones that are close to what the user typed. Those are ones that would tend to have highlights (i.e. exact substring matches). There may need to be a hard cut-off in case many synonyms are highly-ranked. We may also need to remove synonym display completely from the entity info shown for grounded chemicals. Maybe we could just show the IUPAC name under the name the user typed instead of the whole list of synonyms.

we could assign weights to synonyms, for now just like this (higher value, more important):

- `2` for every synonym that matches the search query from the beginning of any word (Sort of regex's $ but word-wise)
- `1` if search term is matched as substring
- `0` for anything else

Then we could say: 10 picks from the sorted list? This will always prioritize the ones with `2`, but eventually we could come with other weights to add and then normalize

maxkfranz:
[...]

PR #213, opened 2018-03-14
Use a special shortening method for displaying synonyms

metincansiper:
Implemented the second solution for displaying long number of synonyms from issue #202

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "form-data": "^2.3.1",
     "fready": "^1.0.0",
     "hash.js": "^1.1.3",
+    "heap": "^0.2.6",
     "hyperscript": "^2.0.2",
     "hyperscript-helpers": "^3.0.2",
     "leven": "^2.1.0",
--- a/src/client/components/element-info/entity-info.js
+++ b/src/client/components/element-info/entity-info.js
@@ -305,6 +305,34 @@ class EntityInfo extends React.Component {
         .then( matches => {
           if( this._unmounted ){ return; }
 
+          let getShortSynonyms = (synonyms, cmpStr = name) => {
+            if (synonyms.length <= MAX_SYNONYMS_SHOWN) {
+              return synonyms;
+            }
+
+            // use a memoized distance function to avoid re-calculating the same distances inside nsmallest function
+            let distance = _.memoize( s => {
+              return distanceMetric(s, cmpStr);
+            } );
+
+            let cmp = (s1, s2) => {
+              return distance(s1) - distance(s2);
+            };
+
+            // fix the first constant number of synonyms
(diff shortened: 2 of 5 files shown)


metincansiper:
I just realized that I will need to update this PR because memoizing the results of comparison function will not be enough. I need to memoize the distances between synonyms and the search text instead.

maxkfranz:
> I need to memoize the distances between synonyms and the search text instead.

Yes, you need a memoized version of `distanceMetric()` rather than `cmp()`.

metincansiper:
I updated it

maxkfranz:
That's great.  It would be a bit better to remove the `shortenSynonyms()` from `render()` entirely and instead add `ent.shortenSynonyms` inside `updateMatches()`.  It's good practice to keep `render()` as cheap and simple as possible since it can be called many times.

`updateMatches()`, on the other hand, is called only when needed.

metincansiper:
I got it. I renamed ```shortenSynonyms()``` as ```getShortSynonyms()``` (I thought my previous naming would be confusing because shorten would imply directly changing synonyms array) then I moved it to ```updateMatches()``` to calculate ```match.shortSynonyms```. Then inside ```render()``` directly used ```match.shortSynonyms```.

maxkfranz:
It looks good.  The only thing I'd change (re. code organisation) at this point would be to move the distance metric definition to util so that it's defined only in one place.  Right now, if the server changes its string distance metric, the UI would be out of sync.

I'll try running the code on my machine tomorrow, if possible, and then I'll merge.

maxkfranz:
Something like `util.stringDistanceMetric()`

metincansiper:
I totally agree but seems like none of the existing files under ```util``` is suitable to put such a function and I will need to create a new file for this purpose. What could be a good name for the new file? It would be dependent on whether any similar functionalities would be added to that file later. 

I think if the file will not have a more general purpose then we can give it the name of this function that is  ```stringDistanceMetric.js```. However, otherwise it would be dependent on what will be expected from the functions inside the new file.

maxkfranz:
Maybe `strings.js` for string util functions.  We have some formatting stuff for entity data, for example, that could also go there eventually.

metincansiper:
Okay, I introduced new function inside ```strings.js``` and used it from both client and server sides.

maxkfranz:
Great.  I'll merge a bit later today.

sacdallago:
Just checking: can this be merged @maxkfranz ? :)

maxkfranz:
I've got a lot on my plate, so I can't always test a PR right away.  In future after I've reviewed the code itself, you're welcome to test PR to make sure (1) that the code changes are effective and (2) the main usecase isn't broken (i.e. textmining to doc, create a node and ground it, create an edge and add arrows).

PR #223, opened 2018-03-20
Form editor mvp cont.

d2fong:
Add some new features to the mvp

This is still mostly a fake interface, the data doesn't link up fully, (entities aren't completed, interactions don't have entities hooked up).  This is mostly to allow:

1. users the ability to add new rows 
2. let users choose from 3 interaction types (phosphorylation, enzyme reaction, and the generic 'interacts with' interaction)
3. renamed the title from 'Form Editor' to 'Enter Pathway As Text'

### DIFF
--- a/src/client/components/form-editor/index.js
+++ b/src/client/components/form-editor/index.js
@@ -2,27 +2,70 @@ const { Component } = require('react');
 const h = require('react-hyperscript');
 const DocumentWizardStepper = require('../document-wizard-stepper');
 
+
+class EntityForm extends Component {
+  constructor(props){
+    super(props);
+    this.state = {
+      entityName: ''
+    };
+  }
+  render(){
+    return h('input[type="text"].form-entity', {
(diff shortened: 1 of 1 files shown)


maxkfranz:
The code looks fine.  One thing you may want to do is to use `data` instead of `state` for all setting.  That way, you can avoid async issues with React not updating `state` when you specify.  You just need a `setData()` wrapper function for `setState()`.

Since React updates `state` in batches for `render()`s, it probably makes sense to use `state` for `render()`.  But I'd use `data` for everything else -- e.g. when you have to check flags async but need to have a fresh value.

For simple components, you don't have to worry about this and can generally just use `state`.  But because so much is going to be async for these components, I think it makes sense to use `data`.

d2fong:
screen shot 2018-03-22 at 11 50 37 am

mvp screenshort

d2fong:
I'm just gonna merge this

Issue #167, opened 2018-01-12
Reach text highlighting


sacdallago:
- Re-engineer current text input ( here: http://factoid.baderlab.org/new-document/fill )
- Get answer from reach
- Highlight text in text input with Reach annotations

d2fong:
- highlight gene names that are recognized by reach
- highlight interactions and their types recognized by reach with a different colour

d2fong:
https://github.com/PathwayCommons/factoid/commit/b8c5dcc02dd82fa6d6de69623b7bde1fa52ac4e5

sacdallago:
Update: this goes back to https://github.com/PathwayCommons/factoid/issues/182#issuecomment-369708008

metincansiper:
I have a few questions and thoughts considering this issue:

>Highlight text in text input with Reach annotations

Does this mean just highlighting specific parts of text while it is still editable? I mean will users still be able to edit the text when specific parts of it are highlighted? If it must still be editable I think it would bring some concerns to consider in terms of functionality and implementation. If that is the case I can explain these concerns. If not I think there could be a button of checkbox to toggle the availability of highlights. The text would become editable when the highlights are off and it would become non-editable when highlights are on.

I think that reusing ```highlighter``` component for this feature would be nice but it will require updating ```highlighter``` since this issue requires using different colors for highlighting references to  entities and interactions while ```highlighter``` does not support different colors (according to answer of my first question ```highligter``` would even need to be updated more).

Currently highlighter gets a list of terms to be highlighted (e.g. ```['term1', 'term2']```). To support this feature it would need to get different lists for different groups like the following:

```js
[
   { // entities
      list: ['ent1', 'ent2',...],
      class: ['highlighted-entity']
   },
   { // interactions
      list: ['int1', 'int2',...],
      class: ['highlighted-interaction']
   },
   ...
]
```
In this case, for example, the ```span``` that represents an entity of the text would have class ```highlighted-entity``` instead of ```highlighter-term``` and users can adjust style for different groups in that way.

Does this make sense or what would be an alternative way?

Also, should the entities or the interactions be prioritized? I mean for example an entity name would be ```abc``` and an interaction name would be ```abc interacts xyz```, in that case highlighter could highlight either ```abc``` or ```abc interacts xyz``` when ```abc interacts xyz``` is encountered. 

[...]

PR #232, opened 2018-03-24
Reach text highlighting on seed page

metincansiper:
To solve #167 

Adopted a few lines of code from b8c5dcc02dd82fa6d6de69623b7bde1fa52ac4e5

The followings are screenshots before and after highlighting:

before highlight
after highlight

### DIFF
--- a/src/client/components/document-seeder.js
+++ b/src/client/components/document-seeder.js
@@ -100,63 +100,63 @@ class DocumentSeeder extends React.Component {
     let retVal = [];
 
     while ( entityIndex < entityIntervals.length && interactionIndex < interactionIntervals.length ) {
-    	let entityInt = entityIntervals[entityIndex];
-    	let interactionInt = interactionIntervals[interactionIndex];
+      let entityInt = entityIntervals[entityIndex];
+      let interactionInt = interactionIntervals[interactionIndex];
 
       // push an interaction object for the part of interaction that intersects no entity
-    	if ( entityInt.start > interactionInt.start ) {
-    		let interactionObj = {
-    			start: interactionInt.start,
+      if ( entityInt.start > interactionInt.start ) {
+        let interactionObj = {
+          start: interactionInt.start,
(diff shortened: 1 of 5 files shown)


sacdallago:
I think this will be a good starting point to give an example of what it'd look like next monday, although the specifics of the final implementation are still up to debate.

maxkfranz:
This needs to be reconsidered from a services standpoint.  What you are showing is basically the same as the UArizona server UI.

In order to have an accurate portrayal of what is the 'end result' from Reach, you need to go through all the processing that the existing service does.  So it would make more sense to build on top of the existing service rather than to just proxy Reach directly.

Is this supposed to offer some kind of utility to the user?  Is this just re-inventing the wheel to colour text from Reach?  If you use Reach directly, this will not be able to evolve into a text-based UI.

metincansiper:
Do you mean [this function ](https://github.com/PathwayCommons/factoid/blob/unstable/src/server/routes/api/document/reach.js#L24) by existing service?

I have another questions assuming that you are referring to that function. 

From todays call I inferred that the main benefit you expect by using existing service is that existing service merges the entities with the same ground. 

I see that two grounds are considered the same if they have the same namespace and id. However, I want to learn which entities are supposed to or could have the same ground? The ones that exactly have the same name? Or would they have different names (maybe synonyms with different names)?

Since the entities with the same ground are merged only one of them will be included in resulting list. In this case should I only highlight the one that is ended up with being in the resulting list (I mean not the ones that are eliminated as a result of merging)? 

Also, one thing confusing me is that as far as I see existing service does not return the  start and end indices of the entities and interactions in the input text. In this PR I was highlighting the entities and interactions using their start and end offsets. If I do not have this information should I highlight any appearance of an entity or interaction in resulting list? If so the answer of my question about which entities could have the same ground would matter. Because if the entities with different names would be merged then we will not be highlighting some of them because they are merged with another name but we will still be highlighting some who are not in the resulting list after merging.

PR #240, opened 2018-04-04
landing page try factoid button goes to the sample doc route

d2fong:


sacdallago:
Not entirely sure `/example-document` works?

I get the following in chrome:

```js
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
bundle.js:1 Uncaught (in promise) SyntaxError: Unexpected token I in JSON at position 0
    at bundle.js:1
```

Hard to debug from the bundle, but maybe it's the reach instance being unreachable from here? What's the expected behaviour?

screen shot 2018-04-05 at 2 57 52 pm

Issue #242, opened 2018-04-06
Landing page section: How does Factoid work?


jvwong:
For the 'How does Factoid work' section, proposing two possible ideas: 

1. Super simple (author) 
![author_workflow_2](https://user-images.githubusercontent.com/4706307/38443431-a5614c5c-39b9-11e8-849d-8a641da1fc1e.png)

2. Greater detail (journals, stakeholder)

![factoid_ workflow _ scenarios](https://user-images.githubusercontent.com/4706307/38443437-ab6de952-39b9-11e8-8f1a-4ee11f14364e.png)

sacdallago:
While this looks great, I fear it's missing the future vision aspect. I'm not even talking about putting in something like "curator", but rather convey the message that the system is pretty modular and **can** be independent of the whole journal submission process if desired.

maxkfranz:
I think this addresses three major users well: journals, authors, and readers (researchers using Factoid data).  Those are the major points we need to touch on.

While other classes of users (e.g. curators, colleagues at a conference, webservices) may be good to consider, they are not the focus of the landing page.  Those concerns belong elsewhere, probably mostly outside of the landing page or at least outside of the subsections Jeff is discussing.  Maybe we could have a separate issue to discuss those other groups of users.

Back on topic:

I think the figures look great.  Here are some minor suggestions:

(1) Maybe we should update the photo in the edit stage to a more-recent version of the UI (i.e. without the blue edgehandles dot).  We could remove as much 'chrome' as possible from the screenshot to keep it simpler (i.e. the edit icons, blue dot, blue selection colour on Mdm2).

(2) I'm concerned that someone might, at first glance, misconstrue that researchers would only access Factoid directly and not also the journals.  I know that's not our intention, but someone giving a quick look at the figure might get a bad impression.

Maybe we should show something like a line back to the journal at the final step to make the data loop more explicit.  Alternatively, we could reword the 'Researchers' box to make it explicit that journals are involved.

d2fong:
Also do you have an svg version?

d2fong:
Can we close this issue?

jvwong:
Sure let's shut it down.

PR #246, opened 2018-04-09
add how it works figures to the landing page

d2fong:
refs #242


maxkfranz:
Merging.  Jeff is tweaking the images and we'll update the repo with the updated images.  Thanks

maxkfranz:
@d2fong FYI we've made some tweaks in 18846a7

Issue #262, opened 2018-05-08
more user friendly new document wizard stepper buttons 


d2fong:
## Description of new feature

Wizard buttons should be moved closer to the center of the screen or given colour accents to give hints to users that they exist.  

## Motivation for new feature

Sometimes users don't notice these buttons in the far corners of the screen.  E.g Christian didn't know where they were at first glance.

PR #263, opened 2018-05-08
wizard stepper buttons are closer to the center of the screen / content

d2fong:
- refs #262


maxkfranz:
Would you please correct the lint error?  It looks like it's a minor change

d2fong:
Ok the lint error has been fixed.

PR #269, opened 2018-05-09
Form components

fdurupinar:
Mocha and Lint tests are passing


maxkfranz:
@fdurupinar Do you want code feedback at this point, or would you prefer to do that later when we have more time?

Since the tests and linting are passing, I'm fine with squashing this in.  But we should eventually do a review.  Maybe after the model updates.

fdurupinar:
Let's do it after the model updates. I haven't finalized anything yet.

PR #270, opened 2018-05-10
My factoids local storage

d2fong:
- implements my factoids using local storage
- every time the user visits a document (via the graph editor only), the doc name, id, and secret will be stored locally
- when the users visits the route /documents, the json in local storage will be parsed and each entry will have a corresponding ui linking out to that particular document.

### DIFF
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -338,6 +338,11 @@ class Editor extends React.Component {
 
   componentDidMount(){
     this.data.mountDeferred.resolve();
+    let doc = this.data.document;
+
+    let myFactoids = JSON.parse(localStorage.getItem('my-factoids')) || [];
+    myFactoids.push({id: doc.id(), secret: doc.secret(), name: doc.name()});
+    localStorage.setItem('my-factoids', JSON.stringify(myFactoids));
   }
 
   componentWillUnmount(){
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -340,9 +340,12 @@ class Editor extends React.Component {
     this.data.mountDeferred.resolve();
     let doc = this.data.document;
 
-    let myFactoids = JSON.parse(localStorage.getItem('my-factoids')) || [];
-    myFactoids.push({id: doc.id(), secret: doc.secret(), name: doc.name()});
-    localStorage.setItem('my-factoids', JSON.stringify(myFactoids));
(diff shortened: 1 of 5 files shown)


src/client/components/editor/index.js
```diff
@@ -338,6 +338,14 @@ class Editor extends React.Component {
 
   componentDidMount(){
     this.data.mountDeferred.resolve();
+    let doc = this.data.document;
+
+    let docs = JSON.parse(localStorage.getItem('my-factoids')) || [];
+    let docData = { id: doc.id(), secret: doc.secret(), name: doc.name() };
+    if( _.find(docs,  docData) == null ){
```
maxkfranz: What about the case where a user first views the document publicly and then privately?  It might make sense to merge each time.

src/client/components/my-factoids.js
```diff
@@ -0,0 +1,32 @@
...
+const { Component } = require('react');
+
+
+class MyFactoids extends Component {
+  render(){
+    let factoids = this.props.factoids.map(factoid => {
+      return h('div.factoid-entry', [
+        h(Link, {
+          className: 'plain-link',
+          to: `/document/${factoid.id}/${factoid.secret}`
```
maxkfranz: If the secret is null, would this go to `/document/some-id/null`?

d2fong: it goes to ```document/id/read-only```

src/client/components/editor/index.js
d2fong: If they have visited both, why not give two links so that they know one is read only and one is write only?  The current non-debug seed flow doesn't mention any concept related to read/write links.

src/client/components/my-factoids.js
maxkfranz: It might be better to use `document/id` without any secret appended to the end at all

src/client/components/editor/index.js
maxkfranz: In that case, it would make sense to have different UI elements for each listed document to allow opening in public and (when available) private mode.  Having a doc listed twice could be confusing later on.

Anyway, real users probably won't be using this page for now.  It's useful for demos  though.

The merging option is a bit simpler, since you don't need any UI changes.

maxkfranz:
@d2fong I'm going to merge this for now.  We can create tickets for the edgecases I mentioned later, if needed.

PR #280, opened 2018-05-16
Help system

d2fong:
- basic help system implemented
- uses the current network to display tips for now
- uses the cytoscape-popper extension to display help tooltips on the nodes/edges

Areas of concern
- the help system is currently shoved into the editor file, maybe it can be modularized and cleaned up in the future


src/client/components/editor/buttons.js
```diff
@@ -7,22 +9,43 @@ const Linkout = require('../document-linkout');
...
+    tippy: {
+      zIndex: tippyTopZIndex
+    }
+  };
+
   if( document.editable() ){
     grs.push([
-      h(Tooltip, { description: 'Add entity', shortcut: 'e' }, [
-        h('button.editor-button.plain-button', { onClick: () => controller.addElement().then( el => bus.emit('opentip', el) ) }, [
+      h(Tooltip, { description: 'Help' }, [
```
d2fong: This only shows the help button if the doc is editable.  Thoughts on this?  In some ways it makes sense because it it emits some tooltips related to editing the nodes / edges, but maybe view mode specific tooltips for the nodes / edges would be useful?

maxkfranz: Makes sense

Issue #278, opened 2018-05-16
Basic Biopax export of the model


maxkfranz:
## Description of new feature

Export 

- [ ] Proteins
- [ ] Chemicals (small molecules)
- [ ] General interaction
- [ ] Expression interaction
- [ ] Phosphorylation interaction
- [ ] Methylation interaction
- [ ] Ubiquination interaction
- [ ] Protein and chemical interaction

## Motivation for new feature

Exporting to PC

metincansiper:
I made the following mappings between factoid interactions and type of binary relations [here](http://www.pathwaycommons.org/pc2/formats#sif):

For ``protein-chemical``
If the target is ``chemical`` then  ``controls-production-of`` or ``consumption-controled-by``
else ``chemical-affects``.

For ``expression``  - ``controls-expression-of``.

For general ``modification`` - ``controls-state-change-of``.

For ``modification`` subtypes including ``methylation``, ``phosphorylation`` and ``ubiquination`` - ``controls-phosphorylation-of`` by changing the effect on target protein.

For ``general`` - ``interacts-with``.

According to table at the [link]((http://www.pathwaycommons.org/pc2/formats#sif)) for ``interacts-with`` the participants are supposed to be proteins but in Factoid ``general`` interaction type would be added between not only proteins but it can be added between protein-chemical and chemical-chemical. Therefore, my mapping there would be wrong or incomplete. However, when I create molecular interactions between protein-chemical and chemical-chemical then export the graph to biopax and open it by chibe the interactions are rendered with no problem.

The following is visualization of a factoid graph where I created any kind of binary relations

<img width="1068" alt="factoid-to-export" src="https://user-images.githubusercontent.com/6534865/40379480-0f7a261c-5dab-11e8-831e-7477343aee77.png">

and the following is how the owl file exported from this graph is visualized by chibe

<img width="595" alt="chibe-viz" src="https://user-images.githubusercontent.com/6534865/40379546-3082b8e2-5dab-11e8-9206-547ab8d7ba4d.png">

Here are JSON representation of [factoid document](https://github.com/PathwayCommons/factoid/files/2027885/factoid-doc.txt) and [the exported owl file](https://github.com/PathwayCommons/factoid/files/2027880/311eb217-1832-4762-8491-7bc58b29ab17.txt).

While exporting to owl I am representing entities by their name accessible through ``entity.name()`` should it be replaced by something else that identifies them better?

[...]

PR #284, opened 2018-05-23
Export to biopax

metincansiper:
To accomplish basic Biopax export of the model (#278)

- I mentioned how it works in the associated issue.
- Works as if factoid to biopax converter webservice is running at port 9090 of localhost since I do not know where it will be deployed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "dice-coefficient": "^1.0.2",
     "eventemitter3": "^1.2.0",
     "express": "^4.13.4",
+    "file-saver": "^1.3.8",
     "form-data": "^2.3.1",
     "fready": "^1.0.0",
     "hash.js": "^1.1.3",
--- a/src/config.js
+++ b/src/config.js
@@ -34,7 +34,7 @@ let defaults = {
   PUBCHEM_CACHE_SIZE: DEFAULT_CACHE_SIZE,
   AGGREGATE_CACHE_SIZE: DEFAULT_CACHE_SIZE,
   MAX_SEARCH_SIZE: 50,
-  BIOPAX_CONVERTER_URL: 'http://localhost:9090/FactoidToBiopaxServer/ConvertToOwl'
+  BIOPAX_CONVERTER_URL: 'http://causalpath.org:9090/FactoidToBiopaxServer/ConvertToOwl'
 };
 
 let envVars = _.pick( process.env, Object.keys( defaults ) );
--- a/src/config.js
+++ b/src/config.js
@@ -33,7 +33,8 @@ let defaults = {
   PUBCHEM_LINK_BASE_URL: 'https://pubchem.ncbi.nlm.nih.gov/compound/',
   PUBCHEM_CACHE_SIZE: DEFAULT_CACHE_SIZE,
   AGGREGATE_CACHE_SIZE: DEFAULT_CACHE_SIZE,
-  MAX_SEARCH_SIZE: 50
+  MAX_SEARCH_SIZE: 50,
+  BIOPAX_CONVERTER_URL: 'http://localhost:9090/FactoidToBiopaxServer/ConvertToOwl'
 };
 
 let envVars = _.pick( process.env, Object.keys( defaults ) );
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -120,11 +120,15 @@ http.post('/query-reach', function( req, res ){
   .then( reachJson => res.json(reachJson) );
 });
 
-http.post('/convert-to-biopax', function( req, res ){
-  let templates = req.body;
-  getBiopaxFromTemplates( templates )
-  .then( result => result.text() )
-  .then( owl => res.send( owl ));
+http.get('/biopax/:id', function( req, res ){
+  let id = req.params.id;
+  Promise.try( loadTables )
+    .then( json => _.assign( {}, json, { id } ) )
+    .then( loadDoc )
+    .then( doc => doc.toBiopaxTemplates() )
+    .then( getBiopaxFromTemplates )
+    .then( result => result.text() )
+    .then( owl => res.send( owl ));
 });
 
 module.exports = http;
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -59,12 +59,20 @@ let runLayout = doc => {
 
 let getReachOutput = text => provider.getRawResponse( text );
 
+let handleResponseError = response => {
+  if (!response.ok) {
+    throw Error(response.statusText);
+  }
+  return response;
+};
+
 let getBiopaxFromTemplates = templates => {
   return fetch( BIOPAX_CONVERTER_URL, {
-      method: 'POST',
-      body: JSON.stringify(templates),
-      headers: { 'Content-Type': 'application/json' }
-  } );
+    method: 'POST',
+    body: JSON.stringify(templates),
+    headers: { 'Content-Type': 'application/json' }
+  } )
+  .then(handleResponseError);
 };
 
 // get existing doc
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -2,6 +2,7 @@ const http = require('express').Router();
 const Promise = require('bluebird');
 const _ = require('lodash');
 const uuid = require('uuid');
+const fetch = require('node-fetch');
 
 
 const Document = require('../../../../model/document');
@@ -9,6 +10,8 @@ const db = require('../../../db');
 const logger = require('../../../logger');
 
 const provider = require('./reach');
+
+const { BIOPAX_CONVERTER_URL } = require('../../../../config');
 
 let newDoc = ({ docDb, eleDb, id, secret }) => {
   return new Document( _.assign( {}, docDb, {
@@ -56,6 +59,14 @@ let runLayout = doc => {
 
 let getReachOutput = text => provider.getRawResponse( text );
 
+let getBiopaxFromTemplates = templates => {
+  return fetch( BIOPAX_CONVERTER_URL, {
+      method: 'POST',
+      body: JSON.stringify(templates),
+      headers: { 'Content-Type': 'application/json' }
+  } );
+};
+
 // get existing doc
 http.get('/:id', function( req, res ){
   let id = req.params.id;
@@ -99,6 +110,13 @@ http.post('/query-reach', function( req, res ){
   getReachOutput( text )
   .then( reachRes => reachRes.json() )
   .then( reachJson => res.json(reachJson) );
+});
+
+http.post('/convert-to-biopax', function( req, res ){
+  let templates = req.body;
+  getBiopaxFromTemplates( templates )
+  .then( result => result.text() )
+  .then( owl => res.send( owl ));
 });
 
 module.exports = http;
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
-      "integrity": "sha1-+PLIh60Qv2f2NPAFtph/7TF5qsg=",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
       "dev": true
     },
     "accepts": {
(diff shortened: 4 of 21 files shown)


src/config.js
```diff
@@ -33,7 +33,8 @@ let defaults = {
   PUBCHEM_LINK_BASE_URL: 'https://pubchem.ncbi.nlm.nih.gov/compound/',
   PUBCHEM_CACHE_SIZE: DEFAULT_CACHE_SIZE,
   AGGREGATE_CACHE_SIZE: DEFAULT_CACHE_SIZE,
-  MAX_SEARCH_SIZE: 50
+  MAX_SEARCH_SIZE: 50,
+  BIOPAX_CONVERTER_URL: 'http://localhost:9090/FactoidToBiopaxServer/ConvertToOwl'
```
maxkfranz: @metincansiper @jvwong Could the BIOPAX_CONVERTER server be hosted somewhere with a public URL so everyone doesn't have to host locally to test?

package.json
```diff
@@ -45,6 +45,7 @@
     "dice-coefficient": "^1.0.2",
     "eventemitter3": "^1.2.0",
     "express": "^4.13.4",
+    "file-saver": "^1.3.8",
```
maxkfranz: How big are the BioPAX XML files typically?

metincansiper: It is 25KB for a Factoid document that includes 11 different interaction types and 2 participants per interaction. It became 49KB when I made it 22 interactions by adding new elements having the same of existing interaction and participant types with different names.

src/config.js
metincansiper: Where would it be hosted? If needed we have a google cloud machine hosted at http://causalpath.org/.

maxkfranz: @metincansiper If possible, it might be better for you guys to host it -- at least for now.

metincansiper:
@maxkfranz the converter webservice is publicly available at http://causalpath.org:9090/FactoidToBiopaxServer/ConvertToOwl.

I updated this PR to use it as well. However, after updating the url travis checks failed. The following is travis failure message:

``The command "sudo apt-get install rethinkdb" failed and exited with 100 during .``

I suppose it must be related to something different.

maxkfranz:
The travis logs say the apt repo was unavailable.  If travis fails but the log isn't indicating a factoid problem, you can restart the build.

I just restarted it, so we'll see if it works...

metincansiper:
Seems like it worked, thanks.

IgorRodchenkov:
Does anyone validates the result of the converting to biopax, addresses warnings/errors if any?
Why to run the converter service at such a weird URL, publicly? Running it internally on the same host with factoid and rethinkdb would be enough and easier to deploy and maintain, no?
@maxkfranz @metincansiper

src/server/routes/api/document/index.js
```diff
@@ -101,4 +120,11 @@ http.post('/query-reach', function( req, res ){
   .then( reachJson => res.json(reachJson) );
 });
 
+http.post('/convert-to-biopax', function( req, res ){
```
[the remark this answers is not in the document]
metincansiper: Actually my initial idea was to avoid being have to fetch the document from database since we already have access the document from where this route would be called. However, it must be a trivial concern and changing it as a GET would be better as you mentioned. It will also give the flexibility of calling the route where document is not ready, if needed in the future.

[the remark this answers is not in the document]
metincansiper: Do you prefer ``biopax`` or ``convert-to-biopax``? I think ``biopax`` would be too general if some other Biopax related api would be added in the future.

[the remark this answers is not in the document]
maxkfranz: I wouldn't be concerned about performance, as it's fast on the server side to query the DB.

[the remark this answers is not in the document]
maxkfranz: > Do you prefer biopax or convert-to-biopax? I think biopax would be too general if some other Biopax related api would be added in the future.

I think it's better to keep it simple and just use 'biopax' for now

metincansiper:
>Does anyone validate (with the biopax validator) the result of the converting to biopax, addresses warnings/errors if any?

@IgorRodchenkov I did not know about that service before. I just did it. There are a few warnings and most importantly an error. Error is labeled as ``no.xrefs``. Explanation of the error is "EntityReference does not have any xrefs".

@maxkfranz does Factoid provide any api to get xrefs of entities?

metincansiper:
I updated the route for the converter service as ``GET /biopax/:id`` as it is discussed.

maxkfranz:
The xref should be to be constructed from the `entity.association()` data.

IgorRodchenkov:
When there are no stanard xrefs attached to entity references and genes in a biopax model, it is barely useful for pathway analysis, IMO.

If in the JSON model there are UniProt, HGNC Symbol (gene name) of a Protein/Gene/Dna, etc. (or ChEBI, Punchem Compound ID), then the biopax converter/exporter must generate RelationshipXref for each such ID and attach to corresponding EntityReference (but don't use UnificationXref please, and never - with PhysicalEntity class and sub-classes, unless it's about a uniprot accession number of a ProteinReference, some standard chemical db ID of a SmallMoleculeReference, or a gene symbol of a Gene object). If there is no EntityReference (e.g, in case of Gene or Complex without components), then attach xref(s) to the corresponding physical entity (or gene).

For example, if you have a TP53 Protein p, then create a new (or re-use existing) RelationshipXref x: x.db='HGNC Symbol', x.id='TP53', and add it to the corresponding ProteinReference r, p.entityReference=r; add x to r.xref ;

metincansiper:
Thanks @maxkfranz 

@IgorRodchenkov 
The following is an example xref/association for TP53

```
{ distance: 0,
  geneNames: [ 'TP53', 'P53' ],
  id: 'P04637',
  name: 'Cellular tumor antigen p53',
  namespace: 'uniprot',
  organism: 9606,
  proteinNames: 
   [ 'Cellular tumor antigen p53',
     'Antigen NY-CO-13',
     'Phosphoprotein p53',
     'Tumor suppressor p53' ],
  type: 'protein' }
```

The xref itself contain an id field. Should I use that id field (P04637 in this case) or TP53 as xref.id?

metincansiper:
@maxkfranz @IgorRodchenkov 
I think that there are 2 different complications about xrefs,

1. During export factoid entities would be incomplete and so would not have any xref.
1. Since Factoid represents binary relations biopax converter would need to create some entities that does not directly included in Factoid model. For example, consider ``controls-production-of`` [here](http://www.pathwaycommons.org/pc2/formats). For that relation Factoid has ``A`` and ``y`` as the participants but Biopax converter needs to create an entity for ``x`` as well.

For the first case what can be done when an entity does not have xref?
For the second case what can be done for ``x``?

Is not creating an entity reference for ``x`` in second case acceptable? And for the first case would it be users responsibility to provide xref for any entity before exporting to Biopax (And a related Biopax error would be expected in case xref is not defined)?

IgorRodchenkov:
Wait, wait... do not panic ;)
Xrefs (for all entity references and entities that have no entity
references for some reason) must be present in our biopax export, at least
we can do our best.

Comments below -

On Thu, May 31, 2018 at 4:39 PM, metincansiper <notifications@github.com>
wrote:

> @maxkfranz <https://github.com/maxkfranz> @IgorRodchenkov
> <https://github.com/IgorRodchenkov>
> I think that there are 2 different complications about xrefs,During export
> factoid entities would be incomplete and so would not have any xref.
>
Well, in such case we can ignore the biopax best practice of having xrefs;
however, our goal is to ground/resolve all the entities in factoid, right?

>
>    1. Since Factoid represents binary relations biopax converter would
>    need to create some entities that does not directly included in Factoid
>    model. For example, consider controls-production-of here
>    <http://www.pathwaycommons.org/pc2/formats>. For that relation Factoid
>    has A and y as the participants but Biopax converter needs to create
>    an entity for x as well.
>
> I think you mix concepts of biopax entity (PhysicalEntity or Gene, i.e.,
state) and EntityReference (i.e., protein or chemical class/type).
We were talking about having at least RelationshipXref (one or several) for
each known EntityReference (ER, e.g., ProteinReference) that we produce;
i.e, when we know uniprot ID and/or HGNC Symbol, etc.
But if we, for a reason, ever generate entity without ER, then we can add
xref to the entity as well.


>
> For the first case what can be done when an entity does not have xref?
> For the second case what can be done for x?Is not creating an entity
> reference for x in second case acceptable? And for the first case would
> it be users responsibility to provide xref for any entity before exporting
> to Biopax (And a related Biopax error would be expected in case xref is not
> defined)?
>
I think so.
In controls-production-of we (factoid model) do not know anything about x;
so simply do not create that;
why to confuse biopax users with some trivial no-name no-id PhysicalEntity
object on the left side of the reaction?..


IR.

maxkfranz:
For this

```
{ distance: 0,
  geneNames: [ 'TP53', 'P53' ],
  id: 'P04637',
  name: 'Cellular tumor antigen p53',
  namespace: 'uniprot',
  organism: 9606,
  proteinNames: 
   [ 'Cellular tumor antigen p53',
     'Antigen NY-CO-13',
     'Phosphoprotein p53',
     'Tumor suppressor p53' ],
  type: 'protein' }
```

the output would be something like

```
xref = {
  id: 'P04637',
  db: 'UniProt'
}
```

We don't have a `gene` entity type or HGNC groundings.  So `HGNC Symbol` support would be speculative and error-prone at this point.

maxkfranz:
BTW, the `distance` should probably be removed from the stored association...

metincansiper:
>I think you mix concepts of biopax entity (PhysicalEntity or Gene, i.e.,
state) and EntityReference (i.e., protein or chemical class/type).
We were talking about having at least RelationshipXref (one or several) for
each known EntityReference (ER, e.g., ProteinReference) that we produce;
i.e, when we know uniprot ID and/or HGNC Symbol, etc.
But if we, for a reason, ever generate entity without ER, then we can add
xref to the entity as well.

Currently, I am associating each entity with and entity reference (since Factoid does not provide any type that would not have an entity reference like complexes) and so I can just add the xrefs to entity references.

BTW while creating entities and entity references I was following this pattern (based on my previous conversations with Ozgun and Emek, but I suppose Ozgun replied some of my questions assuming that the conversion would be from form editor not directly the factoid model):

- If there is already an entity reference with the given name and class (currently Protein or SmallMolecule) then does not create the entity reference use the existing one instead.
- If the entity reference already have a physical entity that has the given location (it is trivial for now) and modification set then does not create the physical entity use the existing one instead.

I think now I will update the first condition and will check if there is an existing entity reference with the same name, class and xref.

>I think so.
In controls-production-of we (factoid model) do not know anything about x;
so simply do not create that;
why to confuse biopax users with some trivial no-name no-id PhysicalEntity
object on the left side of the reaction?..

Actually, what I mean was not creating an entity reference for ``x`` but creating an entity for it. I thought that it would be an example of your case about some entities that do not have entity reference (if possible). However, I think as you mentioned not creating an entity for ``x`` and omitting it since it is trivial makes sense and sounds safer.

metincansiper:
> then the biopax converter/exporter must generate RelationshipXref for each such ID and attach to corresponding EntityReference (but don't use UnificationXref please, and never - with PhysicalEntity class and sub-classes, unless it's about a uniprot accession number of a ProteinReference, some standard chemical db ID of a SmallMoleculeReference, or a gene symbol of a Gene object)

@IgorRodchenkov 
First, I implemented using RelationshipXref and then tried to validate the result with biopax validator. It gave error basically complaining that there was no UnificationXref. Then I tried to use UnificationXref instead of RelationshipXref and validated again. Validation had no error this time. Why you mentioned that I should generate RelationshipXref and should avoid using UnificationXref? Because, I suspect I would be getting the same error unless I do not generate UnificationXref.

IgorRodchenkov:
Well you didn’t provide details and examples at first, and I haven’t reviewed your code or biopax results yet; so, I guessed, and tried to express several concerns regarding xrefs. Looks, we’d better be more specific and clear.
Sure, use unification xref if it’s a ProteinReference and UniProt ID (or SmallMoleculeReference and ChEBI/PubChem ID) case, but be careful in other cases, because unification xref means identity (and in an ideal world, DnaReference cannot have the same db:id unification xref as some ProteinReference or Gene has, etc.)

PS:
Please go ahead, for I now think you know how. I’ll test and try it later myself and report, if anything at all, in a separate ticket. Just don’t mess-up entities and entity references and which has which, like it sounds  in previous comments.

metincansiper:
> Please go ahead, for I now think you know how. I’ll test and try it later myself and report, if anything at all, in a separate ticket. Just don’t mess-up entities and entity references and which has which, like it sounds  in previous comments.

Okay, sounds good. I think I could not express what I mean in that previous comment so you thought that I would be messing up entities and entity references.

metincansiper:
Service handles xrefs now and I updated this PR accordingly.

maxkfranz:
@metincansiper If you change the package.json, it's also necessary to commit the changes to package-lock.json.  It looks like that's why the CI is failing on node@10

metincansiper:
@maxkfranz Okay, I updated it. Tests are successful now.

maxkfranz:
FYI we'll have to create another ticket for moving out the files in util/file

That's client-only code, so it can't go in util.  It should go in client/

Issue #273, opened 2018-05-14
Issues with touch device support


maxkfranz:
## Environment info

- Browser version (client) : Should affect all mobile browsers, tested on Safari iOS 11
- Node.js version (server) : N/A

## Current (buggy) behaviour

Recent iterations have worsened mobile support.

 

## Desired behaviour

Issues are mostly related to uses of Tippy.js:

- [x] Large popovers should not obscure the entire screen.  They need to be resized and re-oriented depending on the screen size.
- [x] The UI for the interaction wizard in the arrowhead stage may need to be redesigned a bit.  The popovers aren't visible, as they overlap each other and the participants.
- [x] The tooltips show at odd times and never disappear.  This might be fixed by updating the version of Tippy.js used.
- [x] The page can sometimes scroll, even though it's supposed to use 100% of the screen space -- no more, no less.

## Minimum steps to reproduce

Open a doc on an iPhone or Android phone.  An iPhone SE is a good test device, because the screen is relatively small.  All of these issues are at their worst on phones, but they should also affect tablets to varying degrees.

maxkfranz:
Downgrading to medium priority, focus on issues that would affect larger screen mobile devices (like iPads) rather than smaller screen devices (like iPhones).

maxkfranz:
PR #286 should resolve issues on both tablets and smartphones

PR #286, opened 2018-05-25
Issues with touch device support

maxkfranz:
Ref :  Issues with touch device support #273 

@d2fong Would you please review this?  It affects the help system.  It works as far as I can tell, but it would be good to have a second person try it out.

@jvwong This PR contains the same sort of fixes that should be applied to pc/app-ui.  (Some fixes don't apply to app-ui, but several do.)  We should add this task to one of the student's queue.

### DIFF
--- a/src/styles/editor/editor.css
+++ b/src/styles/editor/editor.css
@@ -1,7 +1,7 @@
 .editor {
-  height: 100vh;
-  width: 100vw;
-  position: relative;
+  height: 100%;
+  width: 100%;
+  position: absolute;
   opacity: 0.01;
   transition-duration: 300ms;
   transition-property: opacity;
--- a/src/styles/notification.css
+++ b/src/styles/notification.css
@@ -95,10 +95,22 @@
   max-width: calc(100vw - 2em);
 }
 
+@media (min-width: 500px){
+  .corner-notification {
+    max-width: calc(100vw - 4em);
+  }
+}
+
 .corner-notification-active {
   display: block;
 }
 
 .incomplete-entity-notification .notification-open {
   display: none;
+}
+
+@media (max-width: 500px) {
+  .notification-message {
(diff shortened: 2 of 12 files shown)


d2fong:
Here is what the help system looks like on chromes mobile view:
The interaction tooltips still get cut off and they dont seem to point to the interaction.

screen shot 2018-05-25 at 1 41 01 pm

I am guessing some of the tooltip changes need to be applied to these tooltips too?

d2fong:
I think it would be useful to apply these changes to the help system like calculating the max/min width for the tooltips

src/styles/notification.css
```diff
@@ -95,10 +95,22 @@
   max-width: calc(100vw - 2em);
 }
 
+@media (min-width: 500px) {
```
d2fong: You use 500px here but in cy/tippy.js you use 650 to determine whether there is a small screen or not.  Should these two numbers be the same or not?

d2fong:
Ok in that case, I didn't notice anything wrong with the help system in this branch.

src/styles/editor/editor.css
```diff
@@ -1,7 +1,7 @@
 .editor {
-  height: 100vh;
-  width: 100vw;
-  position: relative;
+  height: 100%;
```
d2fong: Why do these need to be changed to percentages?

maxkfranz: I've found that `100vh` isn't what someone would normally expect on iOS -- it includes the bottom toolbar such that anything with `100vh` will overflow the screen.

So then, the only safe way to use `vh` units is for something much smaller than the full screen, like `50vh`.

d2fong: Ok that sounds like something I may have to consider for the help system overlay.  Anyways I don't see anything wrong with the code so I approve of the PR.

Issue #252, opened 2018-04-23
Add interaction and participant type to Reach-created elements


maxkfranz:
## Description of new feature

In the document JSON specified by the Reach service, currently interactions and participants have no type.  That results in all interactions being of type 'general' and participants being undirected.

Specify, where possible, the interaction type (`association`) and the participant sign (`group: positive | negative | unsigned`).  For example, for 'A phosphorylates B':

(1) The interaction should be of type `modification`.

(2) B should be marked as a `positive` participant.

(3) B should be marked with modification state `phosphorylated`.

Note that (3) is entity state.  As I recall, Reach does not specify entity state but it could be inferred based on the interaction type that Reach provides.

## Motivation for new feature

The interaction UI (#199) allows a user to specify all this information.  Where possible, we should capture the same information from Reach.

See also #245.  We may consider populating the interaction with the type data, but the interaction may be marked as incomplete.

ozgunbabur:
We've just discussed a few things related to this issue. Consider the sentence

A phosphorylated B at S123

Current design models this by saying the interaction is modification, and marks node B with phosphorylation:

A   [modifies]   B-p

But what if there are other forms of B in the model? Assume unphosphorylated B is binding to C. There has to be two B nodes in the model that are visually identical, which can be confusing to users.

The solution is to move modifications to the interactions and leave the molecules featureless.

A   [phosphorylates at S123]   B

fdurupinar:
In order to cover the interaction types represented in the form-based editor, the following should be added:
-TRANSLOCATION: A translocates B 
-MOLECULAR INTERACTION:   This will cover physical interaction, complex association and dissociation. Takes a list of molecules.
-ACTIVATION INHIBITION: Refers to protein-protein interaction

maxkfranz:
Counterexample mentioned during the call : p-A interacts with B, where a precondition of the interaction is that A is phosphorylated.

ozgunbabur:
You are right, if you are concerned about representing "p-A interacts with B", I agree that having protein **states** as distinct nodes is the best option. But having states without having processes is very problematic. If we have states, then we should be able to model how these states transform into each other. Let's think we are modeling the below sentence.

A phosphorylates B at S123

If we do this by saying the relation is "modification" and use the phosphorylated B as target (B-pS123), then the structure will be:

A   ----modifies--->   B-pS123

[...]

PR #287, opened 2018-05-26
Add interaction type to reach created elements

metincansiper:
To re-solve #252 

I send queries to Reach server and examined the results to get a mapping between Reach event types/subtypes and Factoid interaction associations.

From my observations: ``expression`` is represented by ``transcription`` in Reach results, ``phosphorylation`` and ``methylation`` are represented as is and any combination of``protein-chemical`` in Factoid are represented as ``activation`` (I was able to try it for "chemical activates/inhibits protein" and "protein produces chemical" but "chemical consumes protein" returns no event frame from Reach).

A query like "TP53 ubiquinates MDM2" returned no event frame, so I did not covered ``ubiquination``. But normally I would expect it to be represented as is just like other modification types (``phosphorylation`` and ``methylation``).

Since ``modification`` subtype cannot be added from network editor I did not cover it. Any interaction I did not cover (e.g. ``modification``) or does not fit the cases I described above are classified as ``General`` in Factoid.

I just defined the associations of interactions by the rules above so interactions comes incomplete (I am not sure if something other than it was expected for this issue or possible to do).

The following is what we initially have for "TP53 phosphorylates MDM2" in the network editor:

reach-phosphorylation

### DIFF
--- a/src/server/routes/api/document/reach.js
+++ b/src/server/routes/api/document/reach.js
@@ -5,6 +5,7 @@ const toJson = res => res.json();
 const fetch = require('node-fetch');
 const FormData = require('form-data');
 const Organism = require('../../../../model/organism');
+const { INTERACTION_TYPE } = require('../../../../model/element/interaction-type/enum');
 const uniprot = require('../element-association/uniprot');
 
 // TODO re-enable once a more stable solution for pubchem xrefs is found
@@ -20,6 +21,13 @@ const REMOVE_DISCONNECTED_ENTS = true;
 const REMOVE_UNGROUNDED_ENTS = false;
 const APPLY_GROUND = true;
 const REMOVE_GROUND_FOR_OTHER_SPECIES = false;
+
+const REACH_EVENT_TYPE = Object.freeze({
+  TRANSCRIPTION: 'transcription',
+  PHOSPHORYLATION: 'phosphorylation',
+  METHYLATION: 'methylation',
+  ACTIVATION: 'activation'
+});
 
 module.exports = {
   // TODO remove this function as reach should never need to be exposed directly
@@ -218,6 +226,7 @@ module.exports = {
           let evtArg = frame.arguments.find( arg => isSingleArgEvt( getFrame( getArgId(arg) ) ) );
           let haveEvtArg = evtArg != null;
           let isBinaryCollapsible = argsAreEntAndEvt && argsAreControllerControlled && haveEvtArg;
+          let isProtein = ele => ele.type == 'protein';
 
           if( argsAreEnts || isBinaryCollapsible ){
             let intn = {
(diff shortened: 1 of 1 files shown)


maxkfranz:
Does Reach indicate direction (which participant is the target) and sign (positive versus negative -- activation versus inhibition)?  If so, that information can be stored by specifying the `group`, e.g.

`
{
  id: 'some-interaction-uuid',
  association: 'phosphorylation',
  entries: [
    { id: 'some-ppt-uuid-for-tp53' },
    { id: 'some-ppt-uuid-for-mdm2', group: 'positive' }
  ]
}
`

src/server/routes/api/document/reach.js
```diff
@@ -226,21 +236,107 @@ module.exports = {
               description: getSentenceText( frame.sentence )
             };
 
-            if( argsAreEnts ){
-              intn.entries = argFrames.map( getElFromFrame ).map( getEntryFromEl );
-            } else if( isBinaryCollapsible ){
```
metincansiper: ``getEntityFrame()`` was being called only if ``isBinaryCollapsible`` is true. I made a change to call it anyways since extra effort caused by it is only a cheap lookup check and we can avoid some code duplication by removing ``isBinaryCollapsible`` and ``argsAreEnts`` checks.

I can keep making ``isBinaryCollapsible`` and ``argsAreEnts`` checks if this is not okay. However, even in that case using these checks only to get something like ``entityFrames`` and making the remaining operations on ``entityFrames`` would make sense to avoid code duplication.

metincansiper:
>Does Reach indicate direction (which participant is the target) and sign (positive versus negative -- activation versus inhibition)? If so, that information can be stored by specifying the group, e.g.

Yes, Reach provides that information. I updated PR to specify the group using that information.

src/server/routes/api/document/reach.js
maxkfranz: @jvwong Do you remember why we restricted reach interactions to only controller-controlled.? I suspect it's to filter out interaction types we don't support, like complex formation.  If that's the case, we may need to keep isBinaryCollapsible in order to filter out unsupported types.

metincansiper: @maxkfranz this code segment is inside ``if( argsAreEnts || isBinaryCollapsible )``. I think for the existing version of the code ``else if( isBinaryCollapsible )`` is functioning as if ``else``.

maxkfranz: If I recall correctly, it's important to keep it as we don't have a spec of the reach output.  We only support cases that we've manually analysed to make sure we convert accurately (even if we don't convert everything).

maxkfranz: Basically, `argsAreEnts || isBinaryCollapsible` means "can the reach output be converted to a binary interaction" -- either directly or indirectly.

maxkfranz:
Assigning to self for testing

PR #288, opened 2018-05-29
Notification list

d2fong:
Implements the todo list as a notification list

screen shot 2018-05-29 at 1 41 56 pm

screen shot 2018-05-29 at 1 27 34 pm

## Features
- new button in the menu that opens the task list.  Also has an indicator showing how many tasks (incomplete entities) are left
- task list that uses notification styles for each todo is shown when the button is clicked

## Major Changes
- Added TaskList component that is used by the main Editor Component
- implemented controller functions in the Editor component to control whether the task list is shown or not
- it didn't make sense to me to make task list notifications dismissible, so I added a notification option to configure whether notifications have a dismiss button or not.

## Controversial Changes
- additional css for the menu buttons and editor graph to handle the cases when the task list is closed or open
- to make the task list more compatible with the help list, the task list is closed before the help mode is activated
- I changed the editor buttons function to an actual react component and inject a className prop to help shift the buttons over when the task list is open

## Limitations
- Clicking the 'show me' for any todo only toggles the element info for that node.  e.g. clicking show me twice will open the node info and then close the node info.
- I also get this error when I toggle node info i.e. multiple calls to 
```js 
bus.emit('opentip', node)
```

screen shot 2018-05-29 at 1 41 21 pm

### DIFF
--- a/src/client/components/editor/help.js
+++ b/src/client/components/editor/help.js
@@ -76,10 +76,13 @@ class Help extends React.Component {
     let showHelp = this.data.showHelp;
     let editable = this.props.document.editable();
     let bus = this.props.bus;
-    let cy = this.props.controller.data.cy;
+    let controller = this.props.controller;
+    let cy = controller.data.cy;
 
     if( !showHelp ){
       bus.emit('showtips');
+
+      controller.resetMenuState();
 
       let ent = cy.nodes(node => !isInteractionNode(node)).first();
       if( !ent.empty() ){
--- a/src/client/components/editor/help.js
+++ b/src/client/components/editor/help.js
@@ -77,13 +77,10 @@ class Help extends React.Component {
     let editable = this.props.document.editable();
     let bus = this.props.bus;
     let controller = this.props.controller;
-    let cy = controller.data.cy;
+    let cy = this.props.cy;
 
-    if( !showHelp ){
+    let showTips = () => {
       bus.emit('showtips');
-
-      controller.resetMenuState();
-
       let ent = cy.nodes(node => !isInteractionNode(node)).first();
       if( !ent.empty() ){
         this.data.entTip = this.makeHelpTooltip( {
@@ -109,12 +106,19 @@ class Help extends React.Component {
         this.data.intnTip.show();
       }
       this.setData({ showHelp: true });
+    };
 
-    } else {
+    let hideTips = () => {
       bus.emit('hidetips');
       if( this.data.entTip ){ this.data.entTip.hide(); }
       if( this.data.intnTip ){ this.data.intnTip.hide(); }
       this.setData({ showHelp: false });
+    };
+
+    if( !showHelp ){
+      controller.resetMenuState().then( showTips );
+    } else {
+      hideTips();
     }
   }
 
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -342,6 +342,13 @@ class Editor extends React.Component {
     }
   }
 
+  resetMenuState(){
+    this.setData({
+      drawMode: false,
+      taskListMode: false
+    });
+  }
+
   render(){
     let { document, bus, incompleteNotification } = this.data;
     let controller = this;
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -346,13 +346,15 @@ class Editor extends React.Component {
     let { document, bus, incompleteNotification } = this.data;
     let controller = this;
 
+    let showTaskList = this.data.taskListMode;
+
     let editorContent = this.state.initted ? [
-      h(Buttons, { controller, document, bus }),
+      h(Buttons, { className: showTaskList ? 'editor-buttons.editor-buttons-shifted' : 'editor-buttons', controller, document, bus }),
       incompleteNotification ? h(CornerNotification, { notification: incompleteNotification }) : h('span'),
       h(UndoRemove, { controller, document, bus }),
-      h('div.editor-graph#editor-graph'),
+      h(`div.${showTaskList ? 'editor-graph-shifted#editor-graph' : 'editor-graph#editor-graph'}`),
       h(Help, { document, bus, controller }),
-      h(TaskList, { document, bus, controller, show: this.data.taskListMode })
+      h(TaskList, { document, bus, controller, show: showTaskList })
     ] : [];
 
     return h('div.editor' + ( this.state.initted ? '.editor-initted' : '' ), editorContent);
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -20,6 +20,7 @@ const defs = require('./defs');
 const Buttons = require('./buttons');
 const UndoRemove = require('./undo-remove');
 const Help = require('./help');
+const { TaskList } = require('./task-list');
 
 const RM_DEBOUNCE_TIME = 500;
 const RM_AVAIL_DURATION = 5000;
@@ -116,6 +117,7 @@ class Editor extends React.Component {
       bus: bus,
       document: doc,
       drawMode: false,
+      taskListMode: false,
       newElementShift: 0,
       mountDeferred: defer(),
       initted: false,
@@ -211,6 +213,17 @@ class Editor extends React.Component {
 
   drawMode(){
     return this.data.drawMode;
+  }
+
+  toggleTaskListMode( ){
+    this.data.bus.emit('toggletasklist');
+    this.setData({
+      taskListMode: !this.data.taskListMode
+    });
+  }
+
+  taskListMode(){
+    return this.data.taskListMode;
   }
 
   addElement( data = {} ){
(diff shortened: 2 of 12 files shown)


maxkfranz:
The UI looks great.  I'll send code comments tomorrow.

Some initial thoughts:

> it didn't make sense to me to make task list notifications dismissible, so I added a notification option to configure whether notifications have a dismiss button or not.

Wasn't this already configurable in the component side?

> additional css for the menu buttons and editor graph to handle the cases when the task list is closed or open

Probably unavoidable if we want the task list to be allowed to remain open rather than auto-close on any interaction.

> to make the task list more compatible with the help list, the task list is closed before the help mode is activated

Makes sense

> I changed the editor buttons function to an actual react component and inject a className prop to help shift the buttons over when the task list is open

I'd have to look at the code to be certain, but wouldn't this mean we'd be using inter-component references?

> I also get this error when I toggle node info i.e. multiple calls to

That's a problem with the version of tippy we're using, it seems.  I've seen the error elsewhere sporadically.  I tried upgrading, but it messes up the entity popovers.  Probably just have to ignore it for now

maxkfranz:
Sorry, I haven't had time to review the code in-depth yet.  I'll put it on the top of my list for tomorrow.

src/client/components/editor/help.js
```diff
@@ -76,11 +76,14 @@ class Help extends React.Component {
     let showHelp = this.data.showHelp;
     let editable = this.props.document.editable();
     let bus = this.props.bus;
-    let cy = this.props.controller.data.cy;
+    let controller = this.props.controller;
+    let cy = controller.data.cy;
```
maxkfranz: It would be preferable for `cy` to be sent as a prop to `Help` rather than knowing about controller internals

src/client/components/editor/index.js
```diff
@@ -329,16 +342,26 @@ class Editor extends React.Component {
     }
   }
 
+  resetMenuState(){
+    this.setData({
+      drawMode: false,
```
maxkfranz: This won't work (no events)

You need toggleDrawMode(false)

src/client/components/editor/index.js
```diff
@@ -213,6 +215,17 @@ class Editor extends React.Component {
     return this.data.drawMode;
   }
 
+  toggleTaskListMode( ){
```
maxkfranz: May be useful to have a bool param to force on/off

src/client/components/editor/task-list.js
```diff
@@ -0,0 +1,164 @@
...
+const Notification = require('../notification');
+const NotificationList = require('../notification/list');
+const NotificationPanel = require('../notification/panel');
+
+const getIncompleteEntities = doc => doc.entities().filter(ent => !ent.completed());
+
+
+const eleEvts = [ 'rename', 'redescribe' ];
+const entEvts = [ 'modify', 'associated', 'unassociated', 'complete', 'uncomplete'];
+const intnEvts = [ 'retype' ];
```
maxkfranz: You may want to simplify the events a bit.  Probably just eleEvts is fine.

You only really need 

- rename
- complete
- uncomplete

maxkfranz: Aside : associated is a typo : should be associate

src/client/components/editor/task-list.js
```diff
@@ -0,0 +1,164 @@
...
+
+  render(){
+    let doc = this.props.document;
+    let ntfns = doc.entities().filter(ent => !ent.completed()).map(ent => {
+      let ntfn = new Notification({
+        openable: true,
+        dismissable: false,
+        openText: 'Show me',
+        active: true,
+        message: `Provide more information for ${ent.name() === '' ? 'unnamed entity' : ent.name()}.`
```
maxkfranz: Not sure if this would be confusing to users or not: The displayed name won't match up in all cases.  If you have an entity named 'pcna' and then you quit the wizard, your message will show 'pcna' but the graph will show '?'.

maxkfranz: Maybe

> Provide more information for pcna (?)

src/client/components/editor/task-list.js
d2fong: There is an associated event https://github.com/PathwayCommons/factoid/blob/unstable/src/model/element/entity.js#L52 and I don't remember what my line of thinking was at the time but I vaguely remember associated working better than associate.

src/client/components/editor/help.js
d2fong: If I pass cy as props I run into some react rendering issues.  I can try to fiddle around some more though.

src/client/components/editor/index.js
d2fong: Okay I changed that, but I also notice that the tooltips are activated before the close task list animation ends, leading to the button tooltips shifted and floating in the middle of the editor.  Thoughts on how I might solve this?  I was thinking I could try to use setInterval.

src/client/components/editor/task-list.js
maxkfranz: associate => as soon as the local action happens (optimistic)

associated => when the association is stored in the server (acknowledgement)

There may be some exceptions, but almost all UI code should only ever need the associate event.

d2fong: Okay good to know for the future.  In either case, I don't listen to those events anymore like you said.

src/client/components/editor/index.js
maxkfranz: I'm not noticing that when I pull in your changes and test locally.  What are the steps to reproduce the issue?

src/client/components/editor/task-list.js
```diff
@@ -10,9 +10,9 @@ const NotificationPanel = require('../notification/panel');
 const getIncompleteEntities = doc => doc.entities().filter(ent => !ent.completed());
 
 
-const eleEvts = [ 'rename', 'redescribe' ];
-const entEvts = [ 'modify', 'associated', 'unassociated', 'complete', 'uncomplete'];
-const intnEvts = [ 'retype' ];
+const eleEvts = [ 'rename' ];
+const entEvts = [ 'complete', 'uncomplete' ];
```
maxkfranz: Shouldn't these be in eleEvts?  Any ele can be completed.  For interactions, it means the same thing in the UI: A user must complete the wizard to get 'complete'.

src/client/components/editor/index.js
```diff
@@ -343,9 +346,9 @@ class Editor extends React.Component {
   }
 
   resetMenuState(){
-    this.setData({
-      drawMode: false,
-      taskListMode: false
+    return Promise.all([this.toggleTaskListMode(false),  this.toggleDrawMode(false)]).then( () => {
+      // wait for menu animations to complete
+      return new Promise( resolve => setTimeout(resolve, 250));
```
maxkfranz: There's a Promise.delay(250) in bluebird you can use that does the exact same thing.  May be useful in future, but not absolutely necessary to change now

src/client/components/editor/task-list.js
maxkfranz: That may need a separate case for the notification message, as the interaction doesn't have a name() I think.

It does have a toString() though, which describes the interaction concisely, e.g. "A phosphorylates B"

maxkfranz: For incomplete interactions, though, the toString() may not be very meaningful.  It could even be something like "(?) interacts with (?)", so maybe a different message would be needed

d2fong: I wasn't aware of the changes made a month ago that allowed all elements to have a completed state.  Right now the task list only focuses on incomplete entities, but I can extend it to support interactions too.

d2fong:
Ok I support interactions now.  Here is an example notification message:

`Provide more information for the interaction between unnamed entity and p53 (?)`

I also use Promise.delay(250) now.

I also cleaned up the events to only listen to eleEvts.

maxkfranz:
Looks perfect.  The only thing we might consider tweaking is whether to sort the elements in render() to show the entities before the interactions.  It's probably better for the user to make associations in that order.

Either way, I think we can merge this before the call on Monday.

d2fong:
Another controversial change but I emit 'hidetips' and 'closetip' in resetMenuState to handle this case:
screen shot 2018-06-01 at 5 52 01 pm

sacdallago:
Sorry to jump in on this so late, but a minor-major cosmetic change: what about opening the notification bar still on the right, rather then on the left?

Not super important, the rationale here being that that's the _standard_ and people will implicitly associate something popping up in the right to notifications. Right cols popping up are usually menu's. It's remarkable how our brains react differently to positioning of UI elements...

d2fong:
It will eventually be on the right.

maxkfranz:
Rationale: Cheaper to integrate on LHS, esp. while integrating with multiple other PRs.  The original design uses the RHS.

d2fong:
@maxkfranz I remembered why I used this.props.controller.data.cy instead of passing cy as props.  

For some reason, this.props.cy is undefined in read only mode.  Any ideas as to why this may be the case?

maxkfranz:
It looks like it has to do with mountDeferred and initting cy in the constructor.  It looks like a general race condition that is triggered often for the editable:false case.

Did we move the cy init outside of componentDidMount for some reason?

d2fong:
To my knowledge it has always been in the constructor, but I can try moving it to componentDidMount and seeing if it solves the issue.

maxkfranz:
Actually, I don't think it would help as render() is called before componentDidMount().  We'd probably need to wait on cytoscape@3.3 for cy.mount() for this to work easily.  Maybe let's just use the workaround you had before reading directly from controller.data

PR #292, opened 2018-06-01
form-components

fdurupinar:


src/client/components/form-editor/entity-form.js
```diff
@@ -15,70 +15,110 @@ class EntityForm extends DirtyComponent {
...
-      }
+
+    this.state.bus.on('hideEntityInfo', ()=>{
+      this.hideEntityInfo();
     });
-    this.dirty();
-  }
 
-  hideEntityInfo(){
+    this.state.entity.on('associated', ()=>{
```
d2fong: Does this code mean that when the entity is associated, change the style of this component?

fdurupinar: If a clone of this entity's association is updated, its information needs to be updated as well. I could only get the EntityInfo component to refresh when its style is updated.

maxkfranz: This seems like it should just be using a flag for info visibility that should be checked in render()

fdurupinar: That's not about visibility though. It's about re-rendering a component inside the render function.

maxkfranz: Either:

(1) You need to pass props to the child component that changes that component itself.

(2) The subcomponent itself needs to know when to change internally.

This seems like a form-only issue, so it's probably (1).

More importantly, though, why would the form need to modify an EntityInfo?  Each EntityInfo is stand-alone.

fdurupinar: Ideally, you would expect (2), but it doesn't happen. EntityInfo takes Entity as its main prop, and whenever an entity's association is updated, EntityInfo is supposed to update its rendering. 
As I explained above, it is the clones that cause the problem. Say we have two interactions "A phosphorylates B" and "A activates expression of C", which are already grounded. Suppose the user decides to change the grounding of A, by interacting with the A in the phosphorylation interaction.  The second A's grounding information needs to be concurrently updated, which requires a re-rendering of its EntityInfo, which doesn't happen on its own.

fdurupinar: I mean, yes these suggestions are great in theory and they are the obvious explanations, but they don't help solve the problem. 
Of course we only observe this in the form because clone entities are shown as separate entity boxes on the form as opposed to the graph editor where the same node can take place in more than one interactions.

fdurupinar: I don't know if there is a value; maybe we can ask about this in the user tests. I would prefer to be able to see them all at once as a user, but then I'm not a biologist.

Showing a single EntityInfo may simplify things. Still, I think this is an issue that remains to be discussed, even if at a later iteration.

PR #293, opened 2018-06-01
form-components

fdurupinar:


fdurupinar:
A single entity info is allowed in the form, but still not in the document. I.e. when the user opens both the graph and the form editor, entity info components can be open in both  as their buses don't communicate with each other. 

Plus, I think the form looks ugly. I'll try to make it look better before Monday if I have the time -- perhaps adding a popover on top of entity boxes. But for now this is complete.

PR #294, opened 2018-06-04
Split editor and app buttons

d2fong:
- Editor buttons are on the left, app buttons placed on the right.  Each implemented as a separate react component
- Task list comes in from the right
- Show ```50+``` for the task list indicator once the number of todos passes 50
- Use props.controller.data.cy as a temp fix for the view mode only cy undefined error
- Add a minimal WIP app bar for the form editor (mostly the same as the network editor more button)

screen shot 2018-06-04 at 5 31 35 pm

### DIFF
--- a/src/client/components/editor/buttons.js
+++ b/src/client/components/editor/buttons.js
@@ -148,6 +148,55 @@ class EditorButtons extends React.Component {
       ])
     ]);
 
+    if( document.editable() ){
+      grs.push([
+        h(Popover, {
+          tippy: {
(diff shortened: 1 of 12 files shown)


maxkfranz:
@d2fong This looks OK code-wise, but maybe we shouldn't use the alpha branch anymore.

d2fong:
Ok makes sense.  Alpha was force updated to be a branch off of unstable so I could deploy it for Christian but ill make and use descriptive branches in the future.

PR #305, opened 2018-06-06
Doc title seeder

d2fong:
Adds document title input
- sends title text as well as body text to the doc creation api

Thoughts about title input constraints?
Should we set a title length limit or anything like that?


d2fong:
screen shot 2018-06-08 at 11 41 37 am

maxkfranz:
The code looks fine, but it may be a good idea to store the master value for the text and document name in `state` rather than in the DOM.  I'm not sure why this pattern was used previously for the "paper text" value.

Issue #295, opened 2018-06-04
Merging unstable into form-components breaks editor/graph synchronization


fdurupinar:
## Environment info

- Browser version (client) : Chrome  66.0.3359.181 (Official Build) (64-bit)
- Node.js version (server) :v8.2.1

## Description 

Synchronization between the graph and form editors is broken after merging unstable into form-components.

I.e. commit 7a391a98b8314b3c128c1211c301818fb02272fe was working fine, but  1f9be760ef48ac0667646f2b47c25bd81586ead8 is buggy due to the recent updates.

## Current (buggy) behaviour
- Interactions do not appear correctly on the other side when added through graph or form.
- Completing an entity on the form side would automatically complete the entity on the graph side in the previous commit. This is not the case anymore.

d2fong:
@fdurupinar Do you have the merged version of the branch somewhere?

metincansiper:
@d2fong it is merged to form-components branch. Merging is done by 1f9be76 so 1f9be76 and later are buggy. The bug comes from [this line](https://github.com/PathwayCommons/factoid/commit/1f9be760ef48ac0667646f2b47c25bd81586ead8#diff-be250af46404c5fa3538c8fa311ac7e6R132). Some ``ele`` there are ``undefined``, so the following error is thrown:

"Unhandled rejection TypeError: Cannot read property 'completed' of undefined"

metincansiper:
I suppose [the line](https://github.com/PathwayCommons/factoid/commit/1f9be760ef48ac0667646f2b47c25bd81586ead8#diff-be250af46404c5fa3538c8fa311ac7e6R132) I mentioned above was not the cause of problem itself but it was one of the effects of the problem. Browser shows that target of edge could not be created before that error:

"Can not create edge `031d1d45-d8af-470a-ade8-481692e3a8e9` with nonexistant target `260efa21-1781-490c-9e5e-81ddaa52e34d`"

maxkfranz:
Does the error happen only once or many times per session?

maxkfranz:
In the form branch, there was also changes to the model made.  If those changes are reverted, what is the effect on this?

metincansiper:
>Does the error happen only once or many times per session?

It happens in network editor any time a new interaction is created from form editor.

metincansiper:
@maxkfranz @d2fong @fdurupinar 
I found out the lines of codes that causes the bug or at least the bug disappears when I comment out that lines.

The lines I mention are 

```
[...]

PR #306, opened 2018-06-06
Form fixes

maxkfranz:
Addresses issues related to form sync

Ref : Merging unstable into form-components breaks editor/graph synchronization #295

### DIFF
--- a/src/model/element/interaction.js
+++ b/src/model/element/interaction.js
@@ -12,6 +12,11 @@ const TYPE = 'interaction';
 const makeAssociation = ( type, intn ) => {
   let Type = _.isString(type) ? getIntnTypeByVal(type) : type;
 
+  // fall back on default type if garbage data is passed as `type` arg
+  if( Type == null || type == null ){
+    return INTERACTION_TYPE.INTERACTION.value;
+  }
+
   return new Type( intn );
 };
 
--- a/src/client/components/editor/cy/doc.js
+++ b/src/client/components/editor/cy/doc.js
@@ -33,6 +33,10 @@ function listenToDoc({ bus, cy, document, controller }){
       oldAni.stop();
     }
 
+    if( el.removed() ){
+      return;
+    }
+
     el.style('overlay-color', defs.editAnimationColor);
 
     let ani = el.animation({
@@ -498,7 +502,6 @@ function listenToDoc({ bus, cy, document, controller }){
   document.on('add', onAddEle);
   document.on('remoteadd', docEl => applyEditAnimation( getCyEl( docEl ) ));
   document.on('remove', onRmEle);
-  document.on('remoteremove', docEl => applyEditAnimation( getCyEl( docEl ) ));
   cy.on('layoutstop', onLayout);
   bus.on('renamedebounce', onRenameDebounce);
   bus.on('removeselected', removeSelected);
--- a/src/client/components/element-info/entity-info.js
+++ b/src/client/components/element-info/entity-info.js
@@ -101,9 +101,11 @@ class EntityInfo extends DataComponent {
       progression.goToStage( s.stage );
     }
 
-    this.onRemoteRename = () => {
+    this.onRename = () => {
       this.setData({ name: this.data.element.name() });
+    };
 
+    this.onRemoteRename = () => {
       if( this.remRenameAni ){
         this.remRenameAni.pause();
       }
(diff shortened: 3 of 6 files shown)


maxkfranz:
In particular, see https://github.com/PathwayCommons/factoid/issues/295#issuecomment-395179861

d2fong:
Looks fine to me.

src/model/element/interaction.js
```diff
@@ -12,6 +12,11 @@ const TYPE = 'interaction';
 const makeAssociation = ( type, intn ) => {
   let Type = _.isString(type) ? getIntnTypeByVal(type) : type;
 
+  // fall back on default type if garbage data is passed as `type` arg
+  if( Type == null || type == null ){
```
metincansiper: Changes looks fine to me but I only stuck around this line. Is not ``Type`` already suppose to be assigned to ``null`` if ``type`` is ``null``?

maxkfranz: Strictly speaking, the extra check on `Type` is not necessary.  I thought it might be useful in case `getIntnTypeByVal()` ever changes -- as it's in another file -- or if we later wanted to use reflection at the end of the `if` to validate the class of `Type`.

Basically, my goal was to make the checks as conservative as possible to help prevent future problems when the code changes.

fdurupinar:
@maxkfranz:
1. Why is remoteRename a mis-use? That was precisely what this code is trying to do : when an entity is renamed from the graph editor or any other window, catch this and force this page to re-render. At least, that is what I understand from *remote*.
2.   getSource was not working at the time getInputParticipant was written. Hence the awkward code. This was a few weeks ago. I don't know if that's fixed now. I didn't have the time to check -- perhaps @metincansiper can do. 
3.  Overall, the code and developers would highly benefit from commenting and documentation for the workflow. Otherwise, development will be based on guessing and trying to make sense of bits and pieces to produce some working code,  which is not the best use of our time.

src/model/element/interaction.js
metincansiper: Actually, what I exactly mean was that ``_.isString(type)`` returns ``false`` if type is ``null`` and ``Type`` is assigned to ``null`` in that case. Therefore, ``Type == null || type == null`` must be equal to ``Type == null`` for the current status of the code. However, I think you would not want to change the code if how ``Type`` is calculated changes in a way that the rule I mentioned gets lost. 

That makes sense, I just wanted to make sure that you did it on purpose.

metincansiper:
>getSource was not working at the time getInputParticipant was written. Hence the awkward code. This was a few weeks ago. I don't know if that's fixed now. I didn't have the time to check -- perhaps @metincansiper can do.

As far as I remember ``getSource()`` and ``getTarget()`` was throwing error when interaction is not complete. However, now it does not.

maxkfranz:
(1) I think you've understood the events themselves fine.  The issue is that you've used the same handler for the general event (local+remote) and remote-only events.  It doesn't cause any serious errors now, but that pattern could be problematic for future changes.

(2) That's a minor thing.  I just revised it because it was simple and inexpensive to address while I was looking at it.

(3) That's a good point about docs, but the model isn't actually the problem with `addInteractionRow()`.  The issue was with structure w.r.t. promises.  We should make tickets for future changes -- jsdoc comments, jsdoc generation, and form review.  

I don't think we should do form refactoring now unless it addresses a bug or helps with the planned tweaks, as it could risk creating other bugs.  However, we should create a ticket for a more thorough code review of the form.  (The code in the project, apart from the form, has been reviewed before merging.  The reviews can also help if there are confusing parts of the API.)

fdurupinar:
I completely agree with you about code reviews. My point is not specifically about the addInteractionRow etc. The implementation of form turned out to be unnecessarily time-consuming and frustrating. In order to avoid these issues in the future, we should go more systematically and make sure everyone is informed about the underlying structure, workflow and design choices. Docs and comments are one good way of doing this ahead of time, unlike code reviews, which are definitely necessary for the sake of good coding practices, but only helpful retrospectively. 

@metincansiper, could you please test the form and graph synchronization and merge the request?

metincansiper:
@metincansiper, could you please test the form and graph synchronization and merge the request?

I think it is working just like 7a391a9.

However, when an interaction is added from network editor it seems like the following in form editor (the one where only the delete icon is visible is the one added from network editor. It was the same in 7a391a9):

screen shot 2018-06-07 at 11 39 46 am

I merged PR since I think it must be a separate issue.

metincansiper:
>Yes, I think it's a separate issue based on the way that the form determines whether to render a line. If there is no source/target distinction, then the form ignores the line.

It renders fine when the page is reloaded.

Issue #299, opened 2018-06-05
PC entity linkouts


maxkfranz:
## Description of new feature

Support multiple linkouts:

- Proteins
  - Uniprot
  - Pathway Commons
- Chemicals / small molecules
  - Chebi
  - Pathway Commons

## Motivation for new feature

Having the PC link shows intent and a proof-of-concept for PC-Factoid integration w.r.t. "pathways related to my factoid".

PR #309, opened 2018-06-08
Add support for Pathway Commons linkouts on entities

maxkfranz:
- PC search uses `q=${fullOfficialName}`, since PC UI does not yet support `q=${ns:id}` queries for `uniprot` and `chebi` namespaces.
- Use PubChem linkouts for chemicals, even though we use Chebi grounding for now.  It works OK, and the PubChem page is a bit nicer.

Ref : PC entity linkouts #299

### DIFF
--- a/src/client/components/element-info/entity-assoc-display.js
+++ b/src/client/components/element-info/entity-assoc-display.js
@@ -51,20 +51,28 @@ let chemical = (m, searchTerms) => {
 };
 
 let link = m => {
-  let url;
+  let pcQ = encodeURIComponent(m.name);
+  let url, nsName;
 
   switch( m.type ){
     case 'protein':
       url = UNIPROT_LINK_BASE_URL + m.id;
+      nsName = 'UniProt';
       break;
     case 'chemical':
       url = PUBCHEM_LINK_BASE_URL + m.id;
+      nsName = 'PubChem';
       break;
   }
 
-  return h('div.entity-info-section', [
-    h('a.plain-link', { href: url, target: '_blank' }, [
-      'More information ',
+  return h('div.entity-info-section.entity-info-linkouts', [
+    h('span.entity-info-title', 'More information'),
+    h('a.plain-link.entity-info-linkout', { href: url, target: '_blank' }, [
+      nsName + ' ',
+      h('i.material-icons', 'open_in_new')
+    ]),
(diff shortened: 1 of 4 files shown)


maxkfranz:
@d2fong @jvwong Not sure if there is a better way to do the linkouts from factoid to app-ui.  I don't think we support q=ns:id style queries on prod yet, right?

maxkfranz:
Screenshots:

screen shot 2018-06-08 at 4 58 49 pm

screen shot 2018-06-08 at 4 58 42 pm

jvwong:
Not yet on the `ns:id`... but that's really just for the 'infobox'

maxkfranz:
OK, here's what the PC linkout for p53 looks like:

screen shot 2018-06-08 at 5 03 24 pm

maxkfranz:
And for ASA:

screen shot 2018-06-08 at 5 04 50 pm

jvwong:
Ideally you'd wanna send us the uniprot accession (or later on the hgnc symbol)

jvwong:
Same goes for ChEBI ID...

maxkfranz:
As soon as it's supported in App-UI, it's easy to change over here.  Probably not worth rushing the next App-UI just for this though.

src/client/components/element-info/entity-assoc-display.js
```diff
@@ -51,20 +51,28 @@ let chemical = (m, searchTerms) => {
...
-  return h('div.entity-info-section', [
-    h('a.plain-link', { href: url, target: '_blank' }, [
-      'More information ',
+  return h('div.entity-info-section.entity-info-linkouts', [
+    h('span.entity-info-title', 'More information'),
+    h('a.plain-link.entity-info-linkout', { href: url, target: '_blank' }, [
+      nsName + ' ',
+      h('i.material-icons', 'open_in_new')
+    ]),
+    h('a.plain-link.entity-info-linkout', { href: `http://apps.pathwaycommons.org/search?q=${pcQ}`, target: '_blank' }, [
```
d2fong: Should apps.pathwaycommons.org/ be a option in config.js?

maxkfranz: Added PC_LINK_BASE_URL

Issue #298, opened 2018-06-05
Tweaks to document creation wizard


maxkfranz:
## Description of new feature

- [ ] Buttons can overflow the scroll (100% height for entire page)
- [ ] Choose editor images can be too large on small screens

...

## Motivation for new feature

Minor tweaks to improve the wizard's demoability

maxkfranz:
Yes, ideally all of the content would be positioned and sized relative to the window size.  For sizing, that probably means using vh, vw, or % units.

PR #314, opened 2018-06-12
Seed ui changes

d2fong:
I had to make the document stepper component more configurable to allow options like labels as props, and which material icon to use.  I also changed the component to not render the forward/back buttons if they are disabled.

screen shot 2018-06-12 at 11 34 12 am
screen shot 2018-06-12 at 11 33 57 am
screen shot 2018-06-12 at 11 33 41 am

### DIFF
--- a/src/client/components/document-seeder.js
+++ b/src/client/components/document-seeder.js
@@ -288,20 +288,24 @@ class DocumentSeeder extends React.Component {
       },[
         h(IntervalHighlighter, { text: this.state.reachHighlightInput, intervals: this.state.reachHighlightIntervals })
       ]),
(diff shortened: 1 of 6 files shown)


d2fong:
#298

d2fong:
Temporary solution to the images being too big - wrap the images so that the user has to scroll vertically to choose them.  Not ideal but with priority shifting to the form, it should be enough for now.

maxkfranz:
@d2fong This looks OK to merge to me.  Is there anything else you wanted to include, or can we good to go ahead?

d2fong:
No I think thats all for now.

PR #316, opened 2018-06-15
Updated form subtypes other than protein modification

metincansiper:
In sync-fix branch protein modification is already updated with little changes. Made the similar changes to other form subtypes as well.

### DIFF
--- a/src/client/components/form-editor/activation-inhibition-form.js
+++ b/src/client/components/form-editor/activation-inhibition-form.js
@@ -6,9 +6,9 @@ class ActivationInhibitionForm extends InteractionForm{
 
 
   render(){
-    const intn = this.state.interaction;
-      const lEnt = this.getInputParticipant();
-      const rEnt = this.getOutputParticipant();
+    let intn = this.data.interaction;
+    let lEnt = this.getInputParticipant();
+    let rEnt = this.getOutputParticipant();
 
 
     let actVal =  intn.association().isInhibition()? "inhibits" : "activates" ;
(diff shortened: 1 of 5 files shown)


maxkfranz:
It was mistakenly left in.  It was used for debugging and easy enabling/disabling of logging.

I must have missed it when merging, as I had lots of individual branches that were merged together to become synch-fix.

It would be great if you would remove the line in this PR.

metincansiper:
Okay, I just wanted to make sure that it is totally okay to remove it before doing so. Removed it now.

d2fong:
The build keeps failing for some reason.

maxkfranz:
I think it's mostly to do with squashing.  The commits I pushed seem to resolve it on my machine, though it was only reproducible with `squash >= 1`.

metincansiper:
Tests look successful now. @maxkfranz @d2fong Have you changed something for Travis or restarted the build?

metincansiper:
>but the molecular interaction section needs to be modified to be binary like the others.

I understand that form editor should not allow adding the third participant for molecular interactions. However, what to do when a molecular interaction has more than 2 participants in a way? (Normally this must not be possible but for an unexpected case where it happens. For example it would happen for docs created before disabling adding the third participant from the form editor.) 

I think that things to do in that case would be asserting that number of participants are 2 and/or just displaying the first 2 participants. Other than that currently we are enabling buttons to delete participants if there are more than 2 participants. One thing to do would be removing that buttons as well. 

However, network editor just tries to display any participant and removing them independent of number of participants. Therefore, I will just remove the add participant button for now if it looks okay.

maxkfranz:
Maybe a simple solution is to just use a two participant template, like `[A] interacts with [B]`.  In this case, no dropdown is needed.  Just plain text could be used between A and B.

metincansiper:
>Maybe a simple solution is to just use a two participant template, like [A] interacts with [B]. In this case, no dropdown is needed. Just plain text could be used between A and B.

Sound good to me I think in that case we will be using the first two participants.

metincansiper:
>Maybe a simple solution is to just use a two participant template, like [A] interacts with [B]. In this case, no dropdown is needed. Just plain text could be used between A and B.

I changed in this way. While deciding participants of molecular interaction I used first 2 non-nil participants directly (https://github.com/PathwayCommons/factoid/pull/316/commits/db45d002c8037b946bb4b83676ece2c9ca24dbab#diff-600c019947d94da34113ff929a40e880R43). 

Actually for the current state of code it won't work if I want to use ``source`` and ``target`` instead. It is something that is different in form and network editors. In network editor users have to decide source and target sides but in form editor both sides are set as ``UNSIGNED``. 

I am not sure if deciding source and target is a need for molecular interaction but maybe setting one of the participants as target would be good to have a better compatibility with the network editor.

maxkfranz:
I agree that source and target shouldn't matter for MI.  I'll review the changes tomorrow

PR #320, opened 2018-06-20
Merge unstable into master for 0.1.0

maxkfranz:
This is an initial merge for 0.1.0.  After this merge, we can go back to to merging in feature branches to unstable.

If unstable is not ready by the end of the week, then master only receives bug fixes (bumping to 0.1.1).  Everything in unstable would then wait until 0.2.0.

If unstable if ready by the end of the week, then we can merge into master again for 0.1.0.

Either way, we tag master as 0.1.0 at the end of this week.

@d2fong & @metincansiper , the merge should be fine but I'm assigning you to review just so someone else can double check.  It would be good to have at least one other person double check before we hit the merge button.


d2fong:
Tested it on my machine, seems to be fine.

metincansiper:
@maxkfranz @d2fong @sacdallago 
This branch works fine fine for me. However, I am having problems with both deployed master and unstable. I was working on creating an issue for it but maybe talking about here would be better (I am even not sure if it is related to code or server connection/availability). 

My browser is Chrome version 67.0.3396.87. I am not sure if it is related but my operating system is MacOS Sierra 10.12.6. 

When I create an interaction from form editor (also the same document is open on a network editor) it sometimes works fine. However, sometimes at least one of the participants is not attached to the interaction properly in network editor. Adding interaction from form editor takes more time than it takes doing so when I run factoid on my localhost. Sometimes it even takes much more longer. Sometimes opening/creating a document takes a lot of time (I suppose what takes time would be mostly connecting server rather than creating document itself). When it does not work it gives the following error in both form and network editors (I cannot make sure if the same error is thrown when I suppose it is working fine because I am not able to make it work again for a while): 

```
deps.js:53 WebSocket connection to 'ws://factoid.baderlab.org/socket.io/?EIO=3&transport=websocket&sid=WoXeGah3ypzt4CQqAAAD' failed: Error during WebSocket handshake: 'Upgrade' header is missing
```

Also I sometimes get server is unavailable error but it usually does not last long. Is it working fine for you when you add interaction from form editor when network editor is open and check the network editor after adding interaction (And it would require to try the same thing a few times). Or are you having slowness issues like me?

maxkfranz:
It sounds like an network issue.  I don't see any slowdown at the moment on my home machine (500Mb/s up/down).

Or it could be an issue on the Apache instance in front of Factoid on the cluster.

sacdallago:
https://github.com/PathwayCommons/factoid/pull/320#issuecomment-398911161 : I've noticed this behavior, too, but not the handshake problem. For me, the problem of the nodes not correctly connecting to one another happens every time (not just occasionally). In turn, this behavior manifests in two ways:

1. The nodes are independent and not connected by an edge on the graph. But some strange UI item appears on the graph (a little node). I have a recording for this.
2. One of the nodes has an edge to the origin point of the node's creation, but the two nodes taking part in an interaction aren't actually connected.

Both these problems gets solved by refreshing the graph editor's browser, so I assume the faulty code is either in the form part signaling the graph that new entities are being constructed, or in the graph part where the UI elements get rendered before their metadata has been fully populated.

Video: https://1drv.ms/v/s!AmKzZBtJc5Beo3O4Tb6FB9Udd6XZ

d2fong:
@sacdallago can you click on that file not found error?  
1. Does that say anything?
2. I noticed that you have quite a few extensions installed, could it be possible that one of them is blocking you from receiving this file?

d2fong:
Could you try again in incognito mode or with all of your extensions disabled?

d2fong:
Also, I noticed in the console that the following message appeared:
```[Random User Agent] User-Agent Javascript Protection disabled```
That doesn't appear on my machine.

**Comment by sacdallago** [comment]
Yep, I was thinking about the extensions, too. Plus: that is a blob which doesn't get found

**Comment by sacdallago** [comment]
> [Random User Agent] User-Agent Javascript Protection disabled
is simply setting the User-Agent to random, e.g. 
```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/65.0.3325.395 Safari/537.36
```

sacdallago:
Okay, the problem seems to be bound to one extension only, that being: ad-blocker (https://chrome.google.com/webstore/detail/adblock/gighmmpiobklfepjocnamgkkbiglidom/).
This causes also the `Failed to load resource: net::ERR_FILE_NOT_FOUND` blob error .

 It's sort of bad news, because a lot of people use this extension. But since it's external to us, for now, I'd ignore it.

sacdallago:
@metincansiper , could it be that you might also have adblocker installed? :)

metincansiper:
>@metincansiper , could it be that you might also have adblocker installed? :)

@sacdallago, no the only chrome extension I have is the one for react debugging.

However, I suspected that my problem would be related to OHSU network restrictions. Therefore, now I tried to run it on my home network and it worked fine. Therefore, my problem must be related to OHSU network.

Issue #328, opened 2018-06-22
Lock already existing elements on the layout performed after adding a new interaction from form editor


metincansiper:
## Description of new feature

Lock already existing elements on the layout performed after adding an interaction from form editor.

## Motivation for new feature

To preserve users mental map by keeping the already existing elements at their previous positions as discussed in the last developers call.

PR #329, opened 2018-06-22
Lock already existing elements on the layout performed after adding a new interaction from form editor

metincansiper:
To re-solve #328

**Changes**

- Updated ``document.applyLayout()`` by adding an option to specify the elements to lock before the layout.
- While calling ``document.applyLayout()`` from form editor used the already existing entities as ``elesToLock`` option. 
- Needed to change the layout that is specified while creating a ``cy`` instance inside ``document.applyLayout()`` from ``grid`` to ``preset`` as having a grid layout there updates cy nodes positions before locking the already existing ones. I suppose it must not break some feature since the function already performs a cose layout with random positions after cy is created. Is there a specific reason to have a grid layout there?

**Results**

From my debugging I can say that already existing entities are locked successfully and ``docEl.reposition()`` is called with correct parameters here (https://github.com/metincansiper/factoid/blob/a0dab526c0328e19c295c4fa6593d978c99e8b29/src/model/document/document.js#L325). However, network editor is not synched with exactly the same positions and that causes ugly layouts as long as network editor is not refreshed after adding interactions from form editor.

For example the following is how the network editor looks like after adding a few interactions from form editor (Positions are not synched correctly. Looks like this if I switch to network editor after adding each interaction. Does not look that terrible but still has wrong positions if I switch to network editor only once after adding all of the interactions.)

screen shot 2018-06-22 at 10 19 39 am

The following is how it looks after refreshing (has the updated positions):

screen shot 2018-06-22 at 10 19 52 am

I decided to open the PR independant of the problem in network editor since I beleive that it is a different problem (#330).

### DIFF
--- a/src/client/components/form-editor/index.js
+++ b/src/client/components/form-editor/index.js
@@ -214,6 +214,7 @@ class FormEditor extends DataComponent {
         h('button.form-interaction-adder-btn', {
           onClick: () => {
             let doc = this.data.document;
+            let existingEntities = doc.entities();
 
             let addInteractionRow = () => this.addInteractionRow( {
               name: formType.type,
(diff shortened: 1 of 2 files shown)


metincansiper:
Closing see #354.

Issue #319, opened 2018-06-19
v0.1 - known shortcomings


sacdallago:
High priority:

- [x] From+Graph: when creating entities in form, they don't layout well in graph. Fix is on the way #311
- [ ] Form+Graph+NLP: Get a list of papers to test against and use these to improve form+graph

Medium priority:

- [x] From: Creation of new entries should be unified (instead of the hard classification as is now) --> single button. Fix on the way #311 
- [x] Form: Branding on the form 
- [x]  From: Can not represent general interaction (but can do so in graph editor).  See also  #297
- [x] From: task list (checks over or beside entity boxes) needs to be integrated.

Low priority:

- [x] From: UX improvements on interaction creation (parallelize promise execution) [code speedup]
- [x] From: focus on text input when creating a new interaction or selecting an ungrounded entity
- [ ] From: No links to current document -- Next sprint #337

metincansiper:
> From: Can not represent general interaction (but can do so in graph editor).

I tried possible combinations on this.

- I created ``chemical interacts with protein`` and ``chemical interacts with chemical`` from form editor and the interactions were available as expected in the network editor. 
- I did the same thing from the network editor and interactions were successfully listed under molecular interaction in form editor. 

Is the problem about ``general`` interaction type or ``protein and chemical`` interaction type? I suppose protein and chemical is the one that is not supported by the form editor.

sacdallago:
From https://github.com/PathwayCommons/factoid/issues/221#issuecomment-398861630 :

- [ ] My factoids makes use of user's localstorage. This might need be extended with user auth stuff on the server side, but most likely this is heavily influenced by later collaborations w/ editors

maxkfranz:
> My factoids makes use of user's localstorage. 

I don't think that's something that should be changed at this point.

sacdallago:
@maxkfranz not changed, no. But made aware of.

sacdallago:
- [ ] NLP highlight is buggy https://github.com/PathwayCommons/factoid/issues/333

maxkfranz:
Remaining issues put into Sprint 5: https://github.com/PathwayCommons/factoid/milestone/9

Leaving this issue open to discuss tagging/merging for 0.1 on Tue call

maxkfranz:
Tag v0.1.0 -- the initial fixes and refactoring we did

Tag v0.2.0 -- the state at the last call

Closing

Issue #297, opened 2018-06-05
Re-evaluate interaction type list


maxkfranz:
## Description of new feature

Consider

- Renaming existing interaction types
- Adding new interaction types

## Motivation for new feature

- Ensure a certain level of BioPAX coverage
- Make sure that the names are unambiguous

maxkfranz:
TODO reserving comment for listing current type mappings

maxkfranz:
@jvwong I've filled in the above list.  Let me know if you notice that I've overlooked something, or feel free to edit the text directly.

jvwong:
Just some minor edits. I think based on the last call we'd have to explicitly declare some types in the UI using more tailored language. In particular, 

- Explicitly state  'Binding'
- Cases where 'Chemical' is involved

metincansiper:
I looked at the table here to see if I did everything correctly in terms of biopax export. It seems that I would need to update something for general interaction type since currently I am always creating molecular interaction for it.

Also for modification type I was only covering the case where the both sides are proteins and seems like current factoid code only allows proteins (https://github.com/PathwayCommons/factoid/blob/unstable/src/model/element/interaction-type/modification.js#L23). However, in the table C -> P  and C -| P are listed under modification. Will they be supported later?

@maxkfranz @jvwong 

- I could not fully understand what "SIF chemical-affects but no entity state info or superclass interaction" means. Should C -> P, C -| P be exported to the biopax in different way when interaction type is general (rather than protein chemical or modification)? 
- Super class interaction does not correspond to something from cPath2 Formats but it is directly the ``Interaction`` class from biopax level 3. Am I right?
- Looks like factoid enables adding general interaction between chemicals but C * C is classified as unsupported here. Should I just pass it when both participants of general interaction is chemicals?
- Is now a good time to make the changes that I mentioned above? Because this issue is still open and general modifications are always names as "A interacts with B" now independent of types of A and B.

maxkfranz:
> I could not fully understand what "SIF chemical-affects but no entity state info or superclass interaction" means. Should C -> P, C -| P be exported to the biopax in different way when interaction type is general (rather than protein chemical or modification)?

It means it's ambiguous.  It could be chemical-affects (but you don't know how it affects protein state) or it could be the top-level interaction in biopax.  Probably the second is safer.

> Super class interaction does not correspond to something from cPath2 Formats but it is directly the Interaction class from biopax level 3. Am I right?

Yes

[...]

PR #332, opened 2018-06-22
UI tweaks and fixes

maxkfranz:
- Makes several tweaks to the network editor & form
- See commit titles for list of changes
- See screenshots below for summary

screen shot 2018-06-22 at 3 17 31 pm

screen shot 2018-06-22 at 3 17 41 pm

screen shot 2018-06-22 at 3 23 46 pm

See #319


d2fong:
Looks fine to me but putting the form button types in a popover would lead to users having to click twice to create an interaction in comparison to the old way.

metincansiper:
Looks fine to me. However there is only one thing confusing me. In the latest edit of [interaction types table of #297](https://github.com/PathwayCommons/factoid/issues/297#issuecomment-394846756) any participant combinations other than Protein-Protein is removed from general interaction type of factoid.

screen shot 2018-06-22 at 4 36 24 pm

However, after this PR factoid still supports other combinations of Protein and Chemical under general interaction type. Therefore, there would be an inconsistency about that.

metincansiper:
Other than that I would propose to have some icons to represent template types and adding them to the left-most of template interaction lines in the form editor (would have content made up from names of template types e.g. "PM" for protein modification). 

Not sure if it looks nice but I am proposing in case you want to consider since having such icons would make distinguishing between different types of templates easier.

maxkfranz:
> any participant combinations other than Protein-Protein is removed from general interaction type of factoid.

The editor selects the type automatically for P*C.  Basically, the list is simplified to indicate only the types that the UI guides the user to use.  A user could manually select the general type (or even the wrong type).  The convertor would have to handle these edgecases either by inferring a more correct type or by just falling back on a generic interaction type.  The list just indicates the intended or "correct" selections now.

maxkfranz:
> Other than that I would propose to have some icons to represent template types and adding them to the left-most of template interaction lines in the form editor (would have content made up from names of template types e.g. "PM" for protein modification).

I could see how this would be useful for debugging, but I don't see how it adds value for the user.  I would assume that a biologist could correctly interpret what lines like "______ [activates] the [phosphorylation] of ______" mean.

maxkfranz:
> Looks fine to me but putting the form button types in a popover would lead to users having to click twice to create an interaction in comparison to the old way.

I agree.  However, I think that the old way looked confusing in that the buttons didn't look like they belonged to the "add interaction" button.

maxkfranz:
Merging for today's demo

sacdallago:
> https://github.com/PathwayCommons/factoid/pull/332#issuecomment-399598603
This is not a major concern to me. I think it's okay.

> Other than that I would propose to have some icons to represent template types and adding them to the left-most of template interaction lines in the form editor (would have content made up from names of template types e.g. "PM" for protein modification).

What I read as interesting from this is to _draw_ a visual comparison between what the form does and the graph editor. We can discuss tomorrow the details

Issue #335, opened 2018-06-25
Clicking one of the help tooltips does not close the help


maxkfranz:
## Environment info

- Browser version (client) : Firefox 60
- Node.js version (server) : Node 8

## Current (buggy) behaviour

The help does not close if you click the node or edge tooltips.

## Desired behaviour

Close the help no matter where you click.

## Minimum steps to reproduce

Click the 'Provide entity name' tooltip after enabling help.

PR #344, opened 2018-06-26
Tooltip overlay bug

d2fong:
refs #335


maxkfranz:
The change makes sense and works well.  The only downside is that the help button can feel a bit slow now.  Maybe we could experiment with slightly lower times, but that's not that important at the moment.

Issue #336, opened 2018-06-25
The textmining does not always complete an interaction


maxkfranz:
## Environment info

- Browser version (client) : Firefox 60
- Node.js version (server) : Node 8

## Current (buggy) behaviour

The interaction TGFB1->ALK5 is not marked as completed.

## Desired behaviour

The interaction should be complete if it comes from textmining.

## Minimum steps to reproduce

Use this text: 

> The transforming growth factor beta (TGFB) superfamily of cytokines are involved in a multitude of activities. The TGFB1 ligand can activate the cell-surface receptor ALK5. In the cytoplasm, activated ALK5 phosphorylates the SMAD2 protein. Phosphorylated SMAD2 is able to bind to the common mediator SMAD4, which regulates transcription.

Then click the edge between TGFB1 and ALK5.

sacdallago:
This sounds like a problem of the text mining? What I understand is that factoid doesn't pick up the interaction type, either because NLP is not returning it, or because factoid is not parsing the NLP response properly. Is this correct?

metincansiper:
See #349

PR #349, opened 2018-06-27
Updates on interactions created through text mining results

metincansiper:
Mainly addresses #336

The reason behind #336 was the following:

- Since ``General`` interaction type is the default for factoid I am classifying any interaction that is not classified as one of the other interaction types as general. However, while doing that I was letting ``getAssociation()`` to return ``null`` if interaction is not classified as any of the other types. I suppose while merging #287 Max added the last step to complete an interaction if it has an association and a direction. However, the problem is that even if  ``getAssociation()`` returns ``null`` I was expecting it to be a ``General`` type of interaction since factoid model use it as default. Returning ``null`` there was confusing and become problematic after the las change of Max. Therefore, I updated it so that ``getAssociation()`` returns ``General`` by default instead of ``null``.

Also since text mining should not bring an incomplete interaction I updated condition of adding interaction from text mining by checking if the interaction is completed as well. One important thing about this is that if an entity only exists as the participant of the interaction that is skipped because it is not complete that entity will be skipped as well. I can remove that extra check if it does not sound good.

### DIFF
--- a/src/server/routes/api/document/reach.js
+++ b/src/server/routes/api/document/reach.js
@@ -301,13 +301,6 @@ module.exports = {
                       return INTERACTION_TYPE.METHYLATION;
                   }
                 }
-                else if ( argsAreEnts ) {
-                  // more cases would be added in the future
-                  switch ( frame.type ) {
-                    case REACH_EVENT_TYPE.ACTIVATION:
-                      return INTERACTION_TYPE.MODIFICATION;
-                  }
-                }
               }
               // protein - chemical
               else if ( protCount == 1 ) {
@@ -320,7 +313,7 @@ module.exports = {
                 }
               }
 
-              return INTERACTION_TYPE.GENERAL;
+              return INTERACTION_TYPE.INTERACTION;
             };
 
             let getEntityFrame = frame => {
--- a/src/server/routes/api/document/reach.js
+++ b/src/server/routes/api/document/reach.js
@@ -299,8 +299,13 @@ module.exports = {
                       return INTERACTION_TYPE.PHOSPHORYLATION;
                     case REACH_EVENT_TYPE.METHYLATION:
                       return INTERACTION_TYPE.METHYLATION;
-                    default:
-                      return null;
+                  }
+                }
+                else if ( argsAreEnts ) {
+                  // more cases would be added in the future
+                  switch ( frame.type ) {
+                    case REACH_EVENT_TYPE.ACTIVATION:
+                      return INTERACTION_TYPE.MODIFICATION;
                   }
                 }
               }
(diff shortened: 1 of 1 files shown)


src/server/routes/api/document/reach.js
```diff
@@ -351,7 +354,7 @@ module.exports = {
               }
             }
 
-            if( intn.entries.length >= 2 ){
+            if( intn.entries.length >= 2 && intn.completed ){
```
metincansiper: This line is what I mentioned at the end of PR comment. I can remove the extra check if preferred.

metincansiper: One option would be removing this extra check from here but throwing error if the interaction is not complete.

maxkfranz: This is probably fine for now.  Maybe in future we could log these miss cases, but throwing an error might be too much

src/server/routes/api/document/reach.js
```diff
@@ -299,8 +299,13 @@ module.exports = {
...
                       return INTERACTION_TYPE.METHYLATION;
-                    default:
-                      return null;
+                  }
+                }
+                else if ( argsAreEnts ) {
+                  // more cases would be added in the future
+                  switch ( frame.type ) {
+                    case REACH_EVENT_TYPE.ACTIVATION:
+                      return INTERACTION_TYPE.MODIFICATION;
```
maxkfranz: Only the subclasses of modification are used in the UI.  If it's not something like phosphorylation, then it should probably be general.

metincansiper:
>Only the subclasses of modification are used in the UI. If it's not something like phosphorylation, then it should probably be general.

@maxkfranz I am writing here since the related change is outdated now.

I removed the related code segment. If an interaction is corresponding to ``Modification`` subtype from the reach results it is associated with ``General`` interaction type now since adding ``Modification`` subtype is not supported by network UI (This was also how it is done by #287 before being updated by this PR, now it is reverted back).

src/server/routes/api/document/reach.js
metincansiper: Yes, I agree it would be breaking.

metincansiper:
I think the "if statement" [here](https://github.com/metincansiper/factoid/blob/8009a2d9723404ba8dde288e0d76b0fa2455d02a/src/server/routes/api/document/reach.js#L339) became useless now. It would be functional if how ``getAssociation()`` works changes in the future. However, I am not sure if having it would be good in that case. It would be causing confusions again (Since even if ``getAssociation()`` returns ``null`` or ``undefined`` the interaction will have a default association by current logic of factoid).

Issue #330, opened 2018-06-22
Node positions in network editor are not synched correctly when docEl.reposition() is called


metincansiper:
## Environment info

- Browser version (client) : Chrome version 67.0.3396.87 
- Node.js version (server) : 8.11.3
- Npm version: 5.6.0
- Rethinkdb version: 2.3.6 (CLANG 8.1.0 (clang-802.0.42))

## Current (buggy) behaviour

Node positions in network editor are not synched correctly when ``docEl.reposition()`` is called from [here](https://github.com/metincansiper/factoid/blob/a0dab526c0328e19c295c4fa6593d978c99e8b29/src/model/document/document.js#L325). However, they are updated after refreshing the network editor (See #329 results).

## Desired behaviour

Node positions in network editor should be updated without needing to refresh the page.

## Minimum steps to reproduce

- Checkout [this branch](https://github.com/metincansiper/factoid/tree/form-layout). Currently it only has only one commit at the top of latest factoid unstable.
- Open the same document in both form and network editors. 
- Add a set of interactions from form editor (you can see the bug better if you switch between editors after adding each interaction).
- Switch to network editor to see unexpected layout.
- Refresh the page to see the expected layout.

maxkfranz:
I'm not seeing this in the unstable or master branches, so I suspect it could be something to do with the commits in the form-layout branch

metincansiper:
@maxkfranz 
Would be happening less frequently but I am able to see the same thing in unstable as well. The results below from my experiments in unstable branch.

The following is how network editor looks after adding some elements from form:

<img width="1429" alt="screen shot 2018-06-22 at 9 45 35 pm" src="https://user-images.githubusercontent.com/6534865/41805858-0a75e542-7666-11e8-8ca7-95a4381be3ae.png">

The followings are node positions for the graph above:

```
[{"x":36.781113061359974,"y":195.6263904197109},{"x":57.975886856398574,"y":30.71984839661434},{"x":272.90040685527856,"y":196.42354818615047},{"x":273.6856419851579,"y":31.970811132406197},{"x":58.645081666304804,"y":525.3828413978408},{"x":30.86630160898244,"y":361.91262959479866},{"x":0,"y":0},{"x":279.4523323113699,"y":360.48090089515745}]
```

The following is how the network editor of same document looks when I refresh the page:

[...]

PR #354, opened 2018-07-09
Form layout improvements

metincansiper:
- Locks the already existing elements before the layout. In this way re-solves #328, replaces #329.
- Merges adding interaction from form and performing the layout after it is done into a single operation (Most probably #330 can also be closed if it was only visible while trying to perform layout after adding interaction).
- Disrupts one form layout if another one is called before it is completed.
- Tried to test by calling ``addInteractionRow()`` twice instead of once when add interaction button is clicked. Two interactions are added successfully. However, this was not enough to be able to test the layout disruption case because even in that case the first layout have already finished before the second layout call. However, I suppose the second layout is called after first one finished but the the elements of first interaction are not created/added yet. Therefore, I was able to test that case.

### DIFF
--- a/src/client/components/form-editor/index.js
+++ b/src/client/components/form-editor/index.js
@@ -5,12 +5,14 @@ const _ = require('lodash');
 const EventEmitter = require('eventemitter3');
 const Promise = require('bluebird');
 const uuid = require('uuid');
+const Cytoscape = require('cytoscape');
 
 const logger = require('../../logger');
 const debug = require('../../debug');
 
 const Document = require('../../../model/document');
 const { exportDocumentToOwl } = require('../../util');
+const { makeCyEles, getCyLayoutOpts } = require('../../../util');
 
 const AppNav = require('../app-nav');
 const Tooltip = require('../popover/tooltip');
@@ -48,6 +50,15 @@ class FormEditor extends DataComponent {
     });
 
     let bus = new EventEmitter();
+
+    // list of new elements to add
+    this.elesToAdd = [];
+    // set of elements that are currently being added
+    this.beingAdded = new Set();
+    // whether a layout is just disrupted by the last one before completed
+    this.layoutDisrupted = false;
+    // the last layout performed repositions new elements
+    this.lastLayout = null;
 
     this.data = {
       document: doc,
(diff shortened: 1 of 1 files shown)


src/client/components/form-editor/index.js
```diff
@@ -140,8 +151,84 @@ class FormEditor extends DataComponent {
...
+
+          el.reposition( pos );
+        };
+
+        let cy = new Cytoscape({
+          headless: true,
+          elements: makeCyEles( doc.elements() ),
+          layout: { name: 'preset' },
+          styleEnabled: true
+        });
```
maxkfranz: Style is enabled, so you have things like animation loops running in the background.  If you don't `cy.destroy()`, then we'll have memory leaks.

src/client/components/form-editor/index.js
```diff
@@ -140,8 +151,84 @@ class FormEditor extends DataComponent {
         let handleElCreation = el => Promise.try( () => synch(el) ).then( () => create(el) );
         let els = [ intn, ...ppts ];
 
-        return Promise.all( els.map(handleElCreation) ).then( () => {
-          return Promise.all( ppts.map(add) ).then( () => add(intn) );
+        let runLayout = cy => {
+          let layout = cy.layout( _.assign( {}, getCyLayoutOpts(), {
+            animate: false,
+            randomize: true
```
maxkfranz: We probably shouldn't randomise, since this is more of an incremental layout usecase.  I think that's why the new nodes go so far away from the pre-existing graph.

src/client/components/form-editor/index.js
```diff
@@ -140,8 +151,84 @@ class FormEditor extends DataComponent {
...
+
+          return layoutDone;
+        };
+
+        let updatePos = el => {
+          let cyEle = cy.getElementById( el.id() );
+          let pos =  _.clone( cyEle.position() );
+
+          el.reposition( pos );
+        };
```
maxkfranz: At some point in the future, the reposition step could be coalesced into the creation step to reduce the number of ops for performance.  It's relatively fast right now, so let's consider this change later.

src/client/components/form-editor/index.js
```diff
@@ -140,8 +151,84 @@ class FormEditor extends DataComponent {
...
+          let elesToAdd = this.elesToAdd.filter( ( el ) => !this.beingAdded.has( el ) );
+          let pptsToAdd = elesToAdd.filter( isPpt );
+          let intnsToAdd = elesToAdd.filter( isIntn );
+
+          return Promise.all( elesToAdd.map( updatePos ) ).then( () => {
+            this.lastLayout = null;
+            els.forEach( el => this.beingAdded.add( el ) );
+
+            let postAdd = () => {
+              let toAddSet = new Set( elesToAdd );
```
maxkfranz: The main benefit of creating the set in this case would be performance, but the set is so small it's probably best to just use `elesToAdd.some()` in the `_.remove()` line to keep things simple and easy to read.

metincansiper: I think in that case ``some()`` will require a function to check so would not it be simpler to directly using [``_.includes()``](https://lodash.com/docs/4.17.10#includes)?

metincansiper:
@maxkfranz 
I made the requested changes other than promise cancelation. I could not do it because the promise that ``layout.promiseOn('layoutstop')`` returns has no ``cancel()`` function. Would it be a versioning issue for the related promise library?

I converted layout to incremental but I can say that after using incremental layout nodes started to overlap more. I think incremental layout would be tend to make nodes overlapping when we lock the already existing elements.

I compared incremental and randomized layout results by adding the same number of interactions.

Result for incremental layout:

incremental

Result for random layout:
random

maxkfranz:
Merged with extra cleanup commit 217c813.

The quality of the layout still isn't great.  I get the same results as your screenshot.  Maybe it's a COSE issue, or maybe COSE isn't meant to be used incrementally with locking.  We should look into this more.

It gives good results without locking and when running on all elements, so it's surprising that the result here is not nearly as good.

*An example of a result by running without locking on the whole graph:*


screen shot 2018-07-16 at 2 13 27 pm

metincansiper:
>Merged with extra cleanup commit 217c813.

I think it is not just a cleanup. Are not we stopping the last running layout now in case a new layout is called before it is done?

maxkfranz:
It's not needed if each run is independent -- only affecting the new elements of the row associated with the row.

Issue #369, opened 2018-09-27
Form 'add interaction' 


jvwong:
The form editor's add interaction button has no result upon click. Should enable user to fill in interaction details.

Reproduce by starting example text from homepage's 'Try Factoid', switching to form and clicking '+Add Interaction' button.

![image](https://user-images.githubusercontent.com/4706307/46156175-e3ca8600-c246-11e8-8d21-66c2fd2a7822.png)

PR #375, opened 2018-10-09
form bugfixes

d2fong:
- fixes an issue I had where the dev server would not start because the env vars used by the webpack environment plugin were not defined

- formEditor: tryPromise seemed to not work in relation to returning a promise after running the layout for the newly added form entities so the form component couldn't update properly.

- fixes issue where completing an interaction in the network would not be shown in the form

refs #369 #376

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -4,7 +4,7 @@ const isProd = env.NODE_ENV === 'production';
 const isProfile = env.PROFILE == 'true';
 const isNonNil = x => x != null;
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -43,7 +44,12 @@ let conf = {
   plugins: [
     isProfile ? new BundleAnalyzerPlugin() : null,
 
-    new webpack.EnvironmentPlugin(envVars),
+    new webpack.EnvironmentPlugin({
+      NODE_ENV: 'development',
```
maxkfranz: Shouldn't this be left to the actual env value?

d2fong: Yeah it should

Issue #361, opened 2018-09-11
Template tweaks re. 'top bar' concept


maxkfranz:
Implement:

- [x] Move the "..." menu to the very top-left of the screen and move several icons as text entries within it.
  - [x] Help
  - [x] Share
  - [x] Downloads
- [x] Put the logo on the top-left of the screen, to the right of the "..." menu.
- [x] Add the "Submit" button to the top-right of the screen.
  - [x] Add the button itself.
  - [x] Revise the task list component to be shown as a short summary in the submit popover.
- [x] Revise the toolbar to only contain the "Add entity" and "Add interaction" buttons.  Revise their icons to a circle and an angled line, respectively.
- [x] Add the button-group form-network switcher underneath the title.

When all of the above are complete:

- [ ] Port to form UI.
- [x] Port to publisher portal.

Mockup reference:

![factoid-123-re-bar page 1](https://user-images.githubusercontent.com/989043/45393241-0f1f6500-b5f9-11e8-91bf-0f1fcae5ab84.jpg)

![factoid-123-re-bar page 2](https://user-images.githubusercontent.com/989043/45393245-1181bf00-b5f9-11e8-85aa-3470d4b2966c.jpg)

![factoid-123-re-bar page 3](https://user-images.githubusercontent.com/989043/45393249-15addc80-b5f9-11e8-84f1-0085e47077cb.jpg)

![factoid-123-re-bar page 4](https://user-images.githubusercontent.com/989043/45393253-18103680-b5f9-11e8-8246-b47cdf3b2ffb.jpg)

![factoid-123-re-bar page 5](https://user-images.githubusercontent.com/989043/45393256-1b0b2700-b5f9-11e8-82c0-f9a2bf33092f.jpg)

maxkfranz:
Preview of latest changes:

![screen shot 2018-09-11 at 7 31 16 pm](https://user-images.githubusercontent.com/989043/45393299-4ee64c80-b5f9-11e8-934e-67ea306dd209.png)

![screen shot 2018-09-11 at 7 31 29 pm](https://user-images.githubusercontent.com/989043/45393306-53ab0080-b5f9-11e8-9b56-ee7f8540d9d9.png)

d2fong:
Okay I will work on the submit button

PR #379, opened 2018-11-05
Added menu bar to form based editor.

metincansiper:
The following is how form based editor looks after adding menu bar.

screen shot 2018-11-05 at 11 08 02 am

It looks nearly same of menu bar of network editor excluding the title part. I think giving the same content to title as well would be nice in terms of consistency. However, network editor title does not contain solely a text. It includes other parts too (https://github.com/PathwayCommons/factoid/blob/unstable/src/client/components/editor/index.js#L344).  If both network and form editors are requested to have the same title contents, how do you feel about creating a new small component like ``TitleContent`` that will have this content?

Other than that I moved ``tasks.js`` from ``src/client/components/editor`` to ``src/client/components`` because it is needed from the form editor as well.

### DIFF
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -20,7 +20,7 @@ const defs = require('./defs');
 const EditorButtons = require('./buttons');
 const MainMenu = require('../main-menu');
 const UndoRemove = require('./undo-remove');
-const { TaskView } = require('./tasks');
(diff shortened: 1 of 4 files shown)


fdurupinar:
They should have the same title content.

d2fong:
I think this refs #361

Issue #382, opened 2018-11-11
Update and test all the "biopax" json templates


IgorRodchenkov:
## Update and test all the "biopax" json templates.

Thanks to @maxkfranz there are "TODO BIOPAX" comments in the src/model modules.
We need to review and implement those methods according to [the interaction types doc](https://docs.google.com/document/d/1mN6_v9JfGJRW0lBLOncLZtZh_fEuakGQkAfG1Mu08QY/view#heading=h.7j1j758vqyn6).

Let's test each case and all of the toBiopaxTemplate() methods. The tests can be added either in test/document.js or better in test/interaction.js, and will not only make sure we get what we discussed and designed in the above document, but also generate a very good example JSON input for updating and testing the factoid-converters project (helps PathwayCommons/factoid-converters#2).

IgorRodchenkov:
@metincansiper, @maxkfranz , @d2fong , @jvwong 
I am already working on this...

IgorRodchenkov:
Done. Closing. (Open a new issue in case we need other changes to the model and biopax-ish templates)

PR #384, opened 2018-11-12
Refs #382 and fixed a couple of prev. refactoring bugs

IgorRodchenkov:
Refs #382 

There are 4 biopax template types now. 
In the Factoid model (interaction types), I re-purposed Interaction class to  what we call "Other" now, and Binding is what the Interaction used to be (MolecularInteraction), etc.

(Currently, you cannot export to BioPAX until factoid-converters is updated too, but you can see the intermediate JSON output in the log; see TODOs..)

See:
- [biopax-type.js](https://github.com/IgorRodchenkov/factoid/blob/model-bin-intns/src/model/element/interaction-type/biopax-type.js)
- [enum.js](https://github.com/IgorRodchenkov/factoid/blob/model-bin-intns/src/model/element/interaction-type/enum.js)
- an [example JSON output](https://github.com/PathwayCommons/factoid-converters/blob/master/src/test/resources/test2.json) (doc.toBiopaxtemplates()): 

I think the new intermediate JSON output (input for factoid-converters ws) is better and easier to convert to a biopax model than before.

I also fixed a bug (there was wrong/missing "chemicalprotein" edge type and interaction type), and now one can successfully draw an edge between a protein and chemical nodes ("other" by default).

If we all like the new json and model, I could also write tests in the test/interaction.js, for each specific interaction and participants case...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ warn.log
 error.log
 debug.log
 /.idea/
-
+*.iml
--- a/package-lock.json
+++ b/package-lock.json
@@ -13119,7 +13119,6 @@
               "version": "1.1.11",
               "bundled": true,
               "dev": true,
-              "optional": true,
               "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
@@ -13134,8 +13133,7 @@
             "code-point-at": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "concat-map": {
               "version": "0.0.1",
@@ -13146,8 +13144,7 @@
             "console-control-strings": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "core-util-is": {
               "version": "1.0.2",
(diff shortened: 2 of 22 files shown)


IgorRodchenkov:
Well, the travis-ci build fails for resons not related to my changes.

As to **UI**, @jvwong , you will notice that there is previously skipped Modification interaction type ("Protein modification/activity change", we could probably shorten it as "Modification/activity change") that now occurs in the list suggested for an edge type. That's because along with specific modifications (and activiti change), such as phosphorylation, we now want to support the case when a protein simply activates/inactivates the other protein without known modification (modification feature will be "active" or "inactive").

IgorRodchenkov:
Ok, please don't delete my fork branch yet :)

jvwong:
If you want to include that top-level, I would just say "Modification".

IgorRodchenkov:
Modification is superclass only for those more specific (De)Phosphorylation, (De)Ubiquitination, etc.

PS:
I was also thinking about Modulation, see the [biopax explanation](http://www.biopax.org/owldoc/Level3/classes/Modulation___-1321982139.html), i.e., when a chemical affects (usually via binding) catalytic activity/potential of a protein. Currently this goes to "Other". On the other hand, we have TransctiptionTranslation type that can have either chemical or protein as controller (source)... Modulation sounds similar but applies not only to the downstream expression processes. Modulation is also similar to the Modification (incl. phosphorylation etc.) as it also changes/modifies target's state (inactive/active) but does not specify protein modification as (biopax) feature...

Issue #388, opened 2018-11-20
REACH to factoid mapping


sacdallago:
Reminder:

@jvwong put together a gdoc on REACH output.

**TODO:**

- [ ] Identify entities that can exist in factoid (green highlight: can be converted, orange highlight: may be converted)
- [ ] Create logic to pour REACH output into factoid model

d2fong:
REACH to factoid notes
Intersection of Reach `simple events` and the interaction types supported by Factoid:
- phosphorylation
- dephosphorylation
- ubiquitination
- deubiquitination
- methylation
- demethylation
- transcription

## Google searches used to find test input:
- "* phosphorylates *", - "* methylates *", 
  - (CDK16 Phosphorylates and Degrades p53)[http://www.thno.org/v08p0650]
  - (EGFR phosphorylates FAM129B to promote Ras activation)[www.pnas.org/content/113/3/644]
  - (CARM1 Methylates GAPDH to Regulate Glucose)[https://www.cell.com/cell-reports/pdf/S2211-1247(18)31370-6.pdf]
 

## Findings
- while the REACH paper may say that phosphorylation, ubiquitination, and methylation may appear in simple events, my initial testing on our version of REACH ( which is not the latest ) leads me to believe that these interaction types appear as `nestedEvents` instead of `simpleEvents`.  

Possible reasons that the information in the paper differs from initial testing results:
- we are using an older version of REACH 
- we are using a different `grounding` from what the paper considers a `nestedEvent` and `simpleEvent`.  i.e. what we consider a `nestedEvent` is not what the paper considers is a `nestedEvent`

IgorRodchenkov:
Note: Factoid model currently only supports bio events for which there is a controller defined, except for binding (molecular interaction) and "other" undirected interaction type. It does not support e.g. transcription, transport, biochemical reaction alone, without a control entity, such as catalysys, modulation or transcription reaction regulation also defined. But REACH produces all sorst of events, often not even properly associating them  (I bet we cannot reliably link e.g., a transcription factor to protein expressed/activated, or a controller protein to the other one it affects via phosphorylation)

johnbachman:
@IgorRodchenkov Not sure exactly what you mean by this. REACH does extract events with their controllers from text. For example:

```
In [1]: from indra.sources import reach

In [2]: rp = reach.process_text('BRAF activation leads to the phosphorylation of MAPK1.')

In [3]: rp.statements
[...]

PR #391, opened 2018-11-22
update reach event type map

d2fong:
- add types according to Jeff's REACH output spec

refs #388 Part 1
https://github.com/PathwayCommons/factoid/issues/388#issuecomment-441089596

### DIFF
--- a/src/server/routes/api/document/reach.js
+++ b/src/server/routes/api/document/reach.js
@@ -262,10 +262,6 @@ module.exports = {
                 return null;
               }
 
-              if ( subtype === REACH_EVENT_TYPE.TRANSCRIPTION ){
-                return PARTICIPANT_TYPE.POSITIVE;
-              }
-
               if ( subtype.startsWith( 'positive' ) ) {
                 return PARTICIPANT_TYPE.POSITIVE;
               }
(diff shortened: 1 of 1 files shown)


src/server/routes/api/document/reach.js
```diff
@@ -298,8 +307,16 @@ module.exports = {
                       return INTERACTION_TYPE.TRANSCRIPTION_TRANSLATION;
```
IgorRodchenkov: I wonder whether Reach's 'transcription' events ever have a controller (e.g., TF protein)... If there are only template and product (gene/dna/rna -> rna or protein), then Factoid does not support such reactions at all.

maxkfranz: We're using an older version of reach, so maybe a newer version handles this case differently.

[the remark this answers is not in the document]
d2fong: @IgorRodchenkov I think REACH_EVENT_TYPE.ACTIVATION is already mapped to the generic interaction type;

https://github.com/PathwayCommons/factoid/pull/391/files#diff-a7b86c73370d994c0c73a46a5aea5480L307

Issue #392, opened 2018-11-22
Modify the form UI model


IgorRodchenkov:
Modify the form UI model to be consistent and work with the new Factoid interaction types and BioPAX templates.

PR #393, opened 2018-11-27
Implements #392 - simple/universal form editor

IgorRodchenkov:
Lot's of improvements and cleaner code (all tests and lint checks passed)
See the commit messages.
Enjoy.


IgorRodchenkov:
@maxkfranz 
OtherInteraction there in interaction-type/other-interaction.js is in fact "Other" (according to the design doc); it is not a super-class for anything else (unlike Modification or InteractionType and the Interaction in element/interaction.js).
What class and file name would you like instead?

IgorRodchenkov:
_Probably_ unrelated broken thing (perhaps never worked or not recently):
- menu action: Submit->(Submitted->)View in PathwayCommons (redirects to apps.pathwaycommons.org);
- there are still uncaught Promise exceptions when using the form ui;

Do we have open issues for these?

PR #397, opened 2018-12-03
Updates to the UI re. the 'Factoid binary interactions' specification

maxkfranz:
- Add custom icons for activation, inhibition, and unsigned arrowheads.
- Update the network context menu with the new arrowhead tools.
- Add more rotation CSS classes for icons.  The arrowhead buttons in the network toolbar use a slight angle.
- Add support for specifying the sign of an interaction in the network UI with three buttons in the toolbar -- one for each sign.
- Revise the interaction wizard:  The sign is chosen before the association.  The sign stage is optional, because the sign is preselected by the three buttons.  When selecting a sign that is incompatible with the current association, the default association (`Interaction` -- other) is set.
- Update edgehandles to v3.5.0.  This is necessary to use the `ghostEdgeParams` feature so that the ghost edge arrowhead matches the selected, edge-drawing button.  Enable snapping.
- Indent the subtypes of modification in the network UI.
- Model cleanup
  - Have a static `allowedParticipantTypes()` function on interaction types.  This allows for better lookup/matching -- before having an instance.
  - Remove `UNSIGNED_TARGET` participant type.  Funda wanted this type to have a specified target on unsigned interactions, but it's never been strictly necessary.  The source and target ordering for an unsigned interaction does not matter.
  - Remove unneeded `static` methods in the model.  Just `require()` the `PARTICIPANT_TYPE` etc. directly.  This was a TODO previously, but it wasn't prioritised.
  - Clean up `toString()` fields under `InteractionType`.  Only `verbPhrase` and `postPhrase` are necessary.
  - Correct the `toString()` outputs according to the specification document.  Some cases did not match the document, and some were broken (partially `undefined`).
  - Remove `setParticipantType()` comment under `Interaction`.  This function is not weird.  It's just setting participant type (sign) on top of a more general-case grouping feature in the model.  In an interaction, each participant may be assigned a group in order to flexibly classify the participants.
- Form cleanup
  - Remove manual specification of interaction association list.  This makes the `render()` more readable.
  - Correct dropdown widths to support full view of labels.
  - Make sure dropdown labels are in all lowercase to be formatted properly as a sentence.
  - When the user selects a new sign for an interaction, the interaction association is reset if the (sign, type) combination is not valid.

### DIFF
--- a/src/client/components/form-editor/interaction-form.js
+++ b/src/client/components/form-editor/interaction-form.js
@@ -3,8 +3,9 @@ const _ = require('lodash');
 const { tryPromise } = require('../../../util');
 const dirtyEvents = ['remoteassociate', 'remoteretype'];
 const EntityForm = require('./entity-form.js');
-const Interaction = require('../../../model/element/interaction');
 const h = require('react-hyperscript');
+const { INTERACTION_TYPES, INTERACTION_TYPE } = require('../../../model/element/interaction-type');
+const { PARTICIPANT_TYPES, getPptTypeByVal } = require('../../../model/element/participant-type');
 
 class InteractionForm extends DirtyComponent {
 
@@ -70,22 +71,24 @@ class InteractionForm extends DirtyComponent {
   }
 
   updateActivationInhibition(val){
+    let intn = this.data.interaction;
+    let assoc = intn.association();
+    let newPptType = getPptTypeByVal(val);
+    let resetNeeded = !this.isInteractionTypeAllowedForSign(assoc, newPptType);
     let rEnt = this.getOutputParticipant();
-    let assoc = this.data.interaction.association();
-    let setPptType = () => {
-      assoc.setParticipantAs( rEnt, Interaction.getPptTypeByVal(val));
-    };
+    let setPptType = () => assoc.setParticipantAs(rEnt, newPptType);
+    let resetIntnType = () => resetNeeded ? this.updateInteractionAssociation(INTERACTION_TYPE.INTERACTION) : Promise.resolve();
     let complete = () => this.completeIfReady();
     let dirty = () => this.dirty();
 
-    return tryPromise( setPptType ).then( complete ).then( dirty );
+    return Promise.all([ setPptType(), resetIntnType() ]).then( complete ).then( dirty );
   }
 
   getModificationType(){
     return this.data.interaction.association().value;
   }
 
-  updateModificationType(val){
+  updateInteractionAssociation(val){
     let associate = () => this.data.interaction.associate(val);
     let complete = () => this.completeIfReady();
     let dirty = () => this.dirty();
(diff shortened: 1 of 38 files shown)


maxkfranz:
Note: If you test locally, you may want to disable pubchem under the aggregate grounding service.  At least for me, pubchem has been timing out a lot.

maxkfranz:
Screenshot sequence:

src/client/components/form-editor/interaction-form.js
```diff
@@ -70,87 +71,68 @@ class InteractionForm extends DirtyComponent {
...
-              ASS.TRANSCRIPTION_TRANSLATION.displayValue),
-            actVals: [SIG.POSITIVE.value, SIG.NEGATIVE.value]},
-          {h: h('option', { value: ASS.MODIFICATION.value },
-              ASS.MODIFICATION.displayValue),
-            actVals: [SIG.POSITIVE.value, SIG.NEGATIVE.value]},
-          {h: h('option', { value: ASS.INTERACTION.value },
-              ASS.INTERACTION.displayValue),
-            actVals: [SIG.UNSIGNED.value, SIG.POSITIVE.value, SIG.NEGATIVE.value]}
-        ].filter(o => o.actVals.indexOf(actVal)>-1).map(o => o.h))
+        (
```
IgorRodchenkov: I like this way and hope it works great :)

src/model/element/interaction-type/binding.js
```diff
@@ -39,7 +47,7 @@ class Binding extends InteractionType {
   }
 
   toString(){
-    return super.toString('binds');
+    return super.toString('binds with');
```
IgorRodchenkov: "A binds B" sounds better than "A binds with B", which has also a different meaning in English, i.e. and e.g: "A binds with B" usually means B is a tool or substance to stick A to smth. That's why I removed "with" before.

src/model/element/participant-type.js
```diff
@@ -1,18 +1,11 @@
 const _ = require('lodash');
 
-const pptType = ( value, displayValue, icon, verb ) => Object.freeze({ value, displayValue, icon, verb });
+const pptType = ( value, displayValue, icon, verbPhrase ) => Object.freeze({ value, displayValue, icon, verbPhrase });
 
 const PARTICIPANT_TYPE = Object.freeze({
```
IgorRodchenkov: Nice that we finally got rid of UNSIGNED_TARGET

src/model/element/participant-type.js
```diff
@@ -1,18 +1,11 @@
...
-  // unsigned target only indicates the target of an interaction; not meant for ui display
-  UNSIGNED_TARGET: pptType('unsigned target', '–',
-    '<i class="material-icons">remove</i>', 'affects'),
-  POSITIVE: pptType('positive', '→',
-    '<i class="material-icons">arrow_forward</i>', 'positively affects'),
-  NEGATIVE: pptType('negative', '⊣',
-    '<i class="material-icons icon-rot-90">title</i>', 'negatively affects')
+  UNSIGNED: pptType('unsigned', 'unsigned', 'icon icon-arrow-unsigned', 'interacts with'),
+  POSITIVE: pptType('positive', 'positive', 'icon icon-arrow-positive', 'activates'),
+  NEGATIVE: pptType('negative', 'negative', 'icon icon-arrow-negative', 'inhibits')
```
IgorRodchenkov: I still believe that 'inhibits' and 'activates' are less general verb phrases than 'negatively affects' and 'positively affects', which (latter ones) work better in some contexts, where former ones might sound too specific or even incorrect. Inhibition and activation, IMHO, go better or mainly about catalyses; whereas the "other" interaction (e.g., production/consumption of a chemical) and transcription/translation regulation type ain't that simple... Perhaps, "promotes", "supresses" would be right words.

src/model/element/interaction-type/binding.js
maxkfranz: The doc says "A binds to B".  I don't think we should use "to", since that implies direction.  I think "A binds with B" sounds more undirected than "A binds B".

@jvwong, what do you think?

src/model/element/participant-type.js
maxkfranz: These are the display values that are shown users.  We decided to only use the words "activation" and "inhibition" for positive and negative in the specification document.  Unless we decide as a group to revise the text again, I think we should just stick with what we have now.

jvwong:
The app keeps hanging when trying to search for an entity:

image

image

jvwong:
> Note: If you test locally, you may want to disable pubchem under the aggregate grounding service. At least for me, pubchem has been timing out a lot.

Ah never mind I saw your comment.

maxkfranz:
I forgot to mention that we're don't yet have the yellow dot on PTM: 

screen shot 2018-12-04 at 12 07 33 pm

That's a bit tricky to implement properly, so it will need to be a separate PR.

PR #400, opened 2018-12-07
Improved shortcuts for the tool set

maxkfranz:
- Each of the edge buttons now has a shortcut key.
- Each of the shortcut keys are numbers, in the same order as the buttons are displayed.  This way, the tool set acts as a visual queue of the shortcut keys.  This is probably easier to remember than particular letters.  (And `A` for activation would conflict with `A` for select all.)
- Because the shortcut keys (`1`, `2`, `3`, `4`) are on the left of the keyboard, it's easy to use your left hand to activate a tool and your right hand to use the tool with the mouse.
- When using the keyboard shortcut `1` to create an entity, the entity is added at the current mouse position rather than by the tool set button.

### DIFF
--- a/src/client/components/editor/buttons.js
+++ b/src/client/components/editor/buttons.js
@@ -41,15 +41,15 @@ class EditorButtons extends React.Component {
       );
 
       grs.push([
-        h(Tooltip, _.assign({}, baseTooltipProps, { description: 'Add an entity', shortcut: 'e' }), [
+        h(Tooltip, _.assign({}, baseTooltipProps, { description: 'Add an entity', shortcut: '1' }), [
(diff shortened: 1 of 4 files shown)


jvwong:
Works for me except below behaviour.
 
 -Press '1' to create a node. 
- Press '2', '3', and '4' in sequence. 
- Add another node with the toolbar icon/mouse. 

In this state, I can no longer drag any of the network nodes. 

Chrome Version 70.0.3538.110 (Official Build) (64-bit)

IgorRodchenkov:
If there’s a bug or design flow, why don’t we fix the PR first or open another issue?

IR

PR #404, opened 2018-12-31
Made switcher in main menu more salient and solved overlap problem of form editor title 

metincansiper:
The followings are how the main menu looks like in network and form editors when the mouse is over the switcher:

screen shot 2018-12-31 at 1 46 35 pm

screen shot 2018-12-31 at 1 46 12 pm

### DIFF
--- a/src/styles/main-menu.css
+++ b/src/styles/main-menu.css
@@ -43,8 +43,7 @@
   }
 }
 
-.main-menu-logo,
-.main-menu-switcher {
+.main-menu-logo {
   font-size: 2.25em;
   margin-right: 0.125em;
   transition-property: filter;
--- a/src/styles/main-menu.css
+++ b/src/styles/main-menu.css
@@ -43,7 +43,8 @@
   }
 }
 
-.main-menu-logo, .main-menu-switcher {
+.main-menu-logo,
+.main-menu-switcher {
   font-size: 2.25em;
   margin-right: 0.125em;
   transition-property: filter;
--- a/src/styles/main-menu.css
+++ b/src/styles/main-menu.css
@@ -43,7 +43,7 @@
   }
 }
 
-.main-menu-logo {
+.main-menu-logo, .main-menu-switcher {
   font-size: 2.25em;
   margin-right: 0.125em;
   transition-property: filter;
--- a/src/client/components/editor-type-switcher.js
+++ b/src/client/components/editor-type-switcher.js
@@ -1,0 +1,29 @@
+const { Component } = require('react');
+const Tooltip = require('./popover/tooltip');
+const { Link } = require('react-router-dom');
+const h = require('react-hyperscript');
+
+class EditorTypeSwitcher extends Component {
+  constructor( props ){
+    super( props );
+  }
+
+  render(){
+    let { networkEditor, document } = this.props;
+
+    let id = document.id();
+    let secret = document.secret();
+    let description = networkEditor ? 'Form Editor' : 'Network Editor';
+    let url = `/${networkEditor ? 'form' : 'document'}/${id}/${secret}`;
+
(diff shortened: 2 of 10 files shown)


src/styles/main-menu.css
```diff
@@ -43,7 +43,8 @@
   }
 }
 
-.main-menu-logo {
+.main-menu-logo,
+.main-menu-switcher {
```
metincansiper: Actually I think that using a single class instead of main-menu-logo and main-menu-switcher would be better but I could not decide what would be a convenient name for that class.

metincansiper:
@sacdallago Actually for the secreenshot from the network editor the reason why 0 is written in a new line is not that it did not fit to one line. It was so because that 0 was representing another information (I suppose the year). The following is the related code segment:

```
          h('div.editor-title-info', [
            h('span', document.authorName() ? `${document.authorName()} et al., ` : ``),
            h('span', `${document.year()}`),
            h('span', document.journalName() ? `, ${document.journalName()}` : ``)
          ])
```

How about keeping it as is and still putting the switcher below it?

metincansiper:
I updated the PR so that the switcher is located under the title in network editor and next to the title in the form editor. The results are as the followings:

screen shot 2019-01-03 at 3 59 50 pm
screen shot 2019-01-03 at 3 59 58 pm

I did not do any changes for cutting document title after some length because I think it is still unclear considering my previous comment. After it become clear I can either update this PR or it can be handled as a separate issue if any change is needed.

sacdallago:
Thanks @metincansiper .

I think in terms of consistency, it's either year in both editors, or in none. I don't see it as an absolutely necessary piece of information, and would remove it from either.

jvwong:
A couple of observations:

Network: Centre position overlaps with 'undo':

pr_404_network

Form: Title position gets obscured by other items:

pr_404_form

metincansiper:
@jvwong I think that the first problem was already existing. Looks like it is not related to the changes in this PR, so maybe a separate issue. For the second one I could not reproduce it. The following is what I get when I have a long title in the form editor:

screen shot 2019-01-10 at 10 47 34 pm

jvwong:
OSX 10.14.2

- Chrome V71.0.3578.98 (768 px width):
pr_404_767x1200_network_chromev71 0 3578 98
pr_404_767x1200_form_chromev71 0 3578 98

- Firefox V71.0.3578.98 (768 px width):

pr_404_767x1024_network_firefoxv64 0 2
pr_404_767x1024_form_firefoxv64 0 2


- Safari V12.0.2(768 px width):
pr_404_767x1024_network_safariv12 0 2
pr_404_767x1024_form_safariv12 0 2

metincansiper:
@jvwong thank you. How is it working for the same titles in unstable branch?

metincansiper:
@jvwong I use the exactly same version of google chrome. I tried it with 768 px width as you did. What I get is the following:

screen shot 2019-01-17 at 10 23 31 am

In my case the switcher is still located at the right of title but, the title and the menu overlaps. However, I see that the same problem happens in the unstable too for the same screen size. See the following ss taken from unstable:

screen shot 2019-01-17 at 10 24 18 am

In that case title overlaps submit button as well. Therefore, I think that the problem is out of the context of this PR, so maybe addressed by a separate issue?

jvwong:
Sure, must fix overlap issue first. Can you open?

metincansiper:
I attached the commit that solves overlapping titles problem to this PR because the related change conflicts the changes here and this PR is not merged yet. I will update the description of this PR accordingly. If that fix is urgently needed in unstable before this one is merged I can fix it there as well.

The following is what I get now for the same case discussed in the comments above:
screen shot 2019-01-17 at 3 01 16 pm

Issue #437, opened 2019-03-18
Can't complete interaction wizard if participant is ungrounded


maxkfranz:
![beam2](https://user-images.githubusercontent.com/989043/54563513-bd88d600-499f-11e9-8e5f-af761ce98f94.gif)

PR #438, opened 2019-03-18
Can't complete interaction wizard if participant is ungrounded #437

maxkfranz:
Addresses : Can't complete interaction wizard if participant is ungrounded #437


maxkfranz:
Wrong branch

PR #440, opened 2019-03-18
One user pilot tweaks

maxkfranz:
Addresses issues in the UI that detract from the single-user pilot phase.


maxkfranz:
Is that only on this branch?

jvwong:
In every version, the read-only URL shows this.

maxkfranz:
OK, I'll make a separate PR for that

PR #458, opened 2019-04-27
Added complex support

metincansiper:
- [This function](https://github.com/PathwayCommons/factoid/blob/0a5bd42e00bb5d837af0011b4f959f7579b7f1de/src/client/components/editor/index.js#L297) removes entities from its interactions on ``bus.on('remove')`` but I think that event is never emitted. Therefore, I made the similar changes for complexes (remove an entity from its complex (by updating its parent) when itself is being removed) [here](https://github.com/PathwayCommons/factoid/pull/458/files#diff-1952cbc4bb19081d2699525dbc107bf4R505) instead (for now). I can change that as you wish.

- If you remove an entity, its listeners are unbinded.However, if you make "undo last delete" to restore it back they are not binded back. I realized that this causes bugs in some cases. For example, if you remove and entry, undo last delete and then update its name (or something else) then the changes will not be visible on remote clients because the clients will not be able to listen to 'onremoteupdate'. The same bug affects complexes as well. For example if an entity is removed and restored back, then located in a complex, then complex is removed and restored back, the entity will not be rendered inside the complex for the remote clients because of the same reason. However, since it is a general bug I think it must be solved out of this PR.


maxkfranz:
> This function removes entities from its interactions on bus.on('remove') but I think that event is never emitted. Therefore, I made the similar changes for complexes (remove an entity from its complex (by updating its parent) when itself is being removed) here instead (for now). I can change that as you wish.

It's just not active in the UI anymore.  You can remove an entity from an interaction in the model, but this isn't used in the current UI:  Interactions are created and deleted with two existing participants.

For complexes, you're right.  The behaviour is necessary and would be active in the UI.

> If you remove an entity, its listeners are unbinded.However, if you make "undo last delete" to restore it back they are not binded back. I realized that this causes bugs in some cases. For example, if you remove and entry, undo last delete and then update its name (or something else) then the changes will not be visible on remote clients because the clients will not be able to listen to 'onremoteupdate'. The same bug affects complexes as well. For example if an entity is removed and restored back, then located in a complex, then complex is removed and restored back, the entity will not be rendered inside the complex for the remote clients because of the same reason. However, since it is a general bug I think it must be solved out of this PR.

Yes, let's do a different PR for that.  I'll try to review this PR as soon as I can, but you could take a look at solving that bug you mentioned in the meantime.  It probably just needs to have an extra event handler for restore in the cytoscape sync setup file.

metincansiper:
> Yes, let's do a different PR for that. I'll try to review this PR as soon as I can, but you could take a look at solving that bug you mentioned in the meantime. It probably just needs to have an extra event handler for restore in the cytoscape sync setup file.

Okay I will make a separate PR for that.

maxkfranz:
I don't notice any obvious things to improve on first pass.  How about we merge this into a separate feature branch for now instead of unstable?  Complexes will be an important part of phase 3.

maxkfranz:
@jvwong I'm not sure whether the function names `addParticipant()` and `removeParticipant()` make sense for a complex.  Maybe `subunit` would be a better term?

jvwong:
>  I'm not sure whether the function names addParticipant() and removeParticipant() make sense for a complex. Maybe subunit would be a better term?

`participant` sounds safest to me as it is general. To me, 'subunit' implies some molecular machine with a particular activity. This is different that an entity that simply physically binds (e.g. drug)

metincansiper:
> I don't notice any obvious things to improve on first pass. How about we merge this into a separate feature branch for now instead of unstable? Complexes will be an important part of phase 3.

Okay I created a feature branch "complex-support" based on unstable and updated the base branch of this PR as "complex-support". Therefore, I will just merge this into "complex-support".

PR #461, opened 2019-04-30
Rebind unbinded events (unbinded during deletion of an ele) on undo last delete

metincansiper:
The main problem that is solved by this PR is described in #458

- While solving the problem I created a new EventEmitter class that extends the third party EventEmitter by functions to undoRemoveAllListeners() and undoLastRemoveAllListeners().
- Used the new (extended) EventEmitter class whenever the other one was being used.
- I can add unit tests for the new EventEmitter functions  if this approach looks okay.


maxkfranz:
Most listeners are already handled by `addEleListeners()` and using an undo would add duplicate listeners and be difficult to manage long-term.

What are the particular listeners that aren't handled by `addEleListeners()`?

metincansiper:
>Most listeners are already handled by addEleListeners() and using an undo would add duplicate listeners and be difficult to manage long-term.

>What are the particular listeners that aren't handled by addEleListeners()?

Looks like only the events that are binded in constructors of ``Element`` and subclasses of it are missing. 

My one finding is that if an element is just being added it will not be having any listeners to restore. However, if it is being added to the document as the result of "undo remove" it will be having some listeners to restore and that ones will be including the ones in ``addEleListeners()``. Therefore, I can upgrade the ``EventEmitter`` to provide a function that tells if the restore list is empty. If it is empty it we can call ``addEleListeners()`` as usual but if it is not empty we can prevent ``addEleListeners()`` call. How does it sound?

metincansiper:
However, the way I mentioned would be problematic for [rmEleListeners()](https://github.com/PathwayCommons/factoid/blob/complex-support/src/client/components/editor/cy/doc.js#L437)

metincansiper:
@maxkfranz then would the safest way to handle this be the following?

```
class ElementType extends BaseElementType {
    constructor() {
        super();
       ...
       ...
       this.addSelfListeners();
    }

   addSelfListeners() {
        // TODO: add the listeners specific to this class not the ones that are inherited from the 
       // base class
   }

   addListeners(){
       super.addListeners();
       this.addSelfListeners();
   }
}

TODO: call el.addListeners() before adding the other listeners in doc.js()
```

metincansiper:
BTW I am not sure if ``rmEleListeners()`` is functional in practice. Because listeners are already removed [here](https://github.com/PathwayCommons/factoid/blob/complex-support/src/client/components/editor/index.js#L98). If ``rmEleListeners()`` is not functional and can/should be removed the first way I mentioned may still work.

maxkfranz:
Maybe it would be best then for the undo feature to be revised.  That way, the listeners can stay simple as they were before.

Instead of restoring things on undo, we could recreate the entities and interactions from scratch.  It should be straight forward, since you can serialise with `json()` and pass those jsons to the constructors.

metincansiper:
@maxkfranz I am working on it but ``el.json()`` does not keep anything for grounding of ``el``. Is it safe to update that method so that it will hold some grounding info? I think otherwise the grounding info will be lost for the restored elements.

metincansiper:
I think that it is already included in json of interactions but not json of entities. It is because ``json()`` is implemented in the following way in both:

```
  json(){
    return _.assign( {}, super.json(), _.pick( this.syncher.get(), _.keys(DEFAULTS) ) );
  }
```

No problem for interactions since their defaults already includes a key value pair for ``association`` but entities miss that. May I possibly add ``association: null`` to defaults of ``Entity`` class?

maxkfranz:
> No problem for interactions since their defaults already includes a key value pair for association but entities miss that. May I possibly add association: null to defaults of Entity class?

I think that should be OK

metincansiper:
I am already able to solve the main problem here in my branch but looks like the bug fix for (https://github.com/PathwayCommons/factoid/issues/465) may also be implemented in this PR, since they are highly related and making 2 separate PR for them will possibly cause a lot of conflicts.

maxkfranz:
Whatever is simpler is better, so one PR is good

metincansiper:
Handled by #465, so closing this one.

Issue #457, opened 2019-04-15
Error handling on Document API routes 


jvwong:
Any calls to the endpoint `/api/document/*` that results in an error are not being handled. For instance, if a GET call to a non-existent ID is made the server throws an error that it is not found but the client simply hangs.

Should return an appropriate HTTP status and message upon error.

PR #462, opened 2019-05-01
Document api routes: error handling. 

jvwong:
- catch errors and pass to next
- use default handler from ExpressJS

Refs #457

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -70,7 +70,7 @@ app.use(function(req, res, next) {
 });
 
 // general error handler
-app.use(function(err, req, res, next) {
+app.use(function(err, req, res) {
   let status = err.status || 500;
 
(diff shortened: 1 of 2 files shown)


src/server/index.js
```diff
@@ -61,35 +61,35 @@ if( app.get('env') === 'development' ){
 app.use( '/', require('./routes/index') );
 
 // catch 404 and forward to error handler
-app.use(function(req, res, next) {
-  let err = new Error('URL not found: ' + req.url);
+// app.use(function(req, res, next) {
```
maxkfranz: I don't think this can be deleted...

maxkfranz:
If the default error handler is removed, things like 404 pages won't work

jvwong:
I thought I pushed a version with it back in there. Will do again...

PR #463, opened 2019-05-07
Roll back the name & video

maxkfranz:


jvwong:
There's a bunch of changes to text I want to make (and new intro video) so I'll add to this branch...

PR #467, opened 2019-05-07
Fixed listeners problem on undo last remove operation, also fixed #465

### LINKED ISSUE #465: Undo remove is buggy when an interaction and one or more participant of it is removed together
In http://unstable.factoid.baderlab.org try to make a remove operation that will remove an interaction and at least one of its entities together. Then perform "undo last delete". The undo operation will fail with the fallowing error message in browser:

```
Uncaught TypeError: Cannot read property 'id' of undefined
    at s (bundle.js:1)
    at bundle.js:1
    at Array.reduce (<anonymous>)
    at a (bundle.js:1)
    at J (bundle.js:1)
    at e.Z (bundle.js:1)
    at i.emit (deps.js:1)
    at e.value (bundle.js:1)
    at n (bundle.js:1)
    at bundle.js:1
```

metincansiper:
See #461 and #465

### DIFF
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -59,6 +59,7 @@ class Editor extends DataComponent {
       if( now - l.lastTime > RM_DEBOUNCE_TIME ){
         l.els = [];
         l.ppts = [];
+        l.oldIdToEl = new Map();
       }
 
       l.lastTime = now;
@@ -134,6 +135,7 @@ class Editor extends DataComponent {
       rmList: {
         els: [],
         ppts: [],
+        oldIdToEl: new Map(),
         lastTime: 0
       }
     });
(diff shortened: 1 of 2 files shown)


metincansiper:
In [this line](https://github.com/metincansiper/factoid/blob/undo-rm-listener-fix2/src/client/components/editor/index.js#L329) it is needed to travers ``rmList.ppts`` and filter the ones where ``intn`` matches. And after the changes this is the only code segment where ``rmList.ppts`` is utilized (after the changes no need for the old ``restorePpts()``). 

I think with this usage it may be more efficient to map the participants to interactions instead of keeping a flat array (in case of a big number of elements are removed together) while the current way is simpler. I did not want to make that change directly in the first version but I can make it if that sounds better.

PR #471, opened 2019-05-09
Switch between pc grounding search service and current aggregate search implementation by using an environment variable

metincansiper:


metincansiper:
I am not able to test the reach side since in my all trials reach returned 0 entities (in the deployed unstable as well).

maxkfranz:
Don't worry about the reach part.  We will no longer be using it in future to seed documents with entities and interactions anyway.

src/config.js
```diff
@@ -37,6 +37,8 @@ let defaults = {
   MAX_SEARCH_SIZE: 50,
   BIOPAX_CONVERTER_URL: 'https://biopax.baderlab.org/convert/v2/',
   PC_URL: 'https://apps.pathwaycommons.org/',
+  GROUNDING_SEARCH_URL: 'http://localhost:3001',
```
maxkfranz: Maybe it should be marked as "base" url to be clearer?

metincansiper:
Okay, I updated it as `GROUNDING_SEARCH_BASE_URL `

PR #478, opened 2019-05-16
Added docker compose file

jvwong:
This more serves as documentation about components, environment variables.  
Refs #476.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,26 @@
+version: '3'
+services:
+  webapp:
+    image: pathwaycommons/factoid:unstable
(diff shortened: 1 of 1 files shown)


IgorRodchenkov:
@jvwong, @maxkfranz  I think we need to use Docker volume for the Elasticsearch container; otherwise the database/index will be lost every time we rebuild the container.

jvwong:
@IgorRodchenkov OK, I wanted to ask your opinion about that - but I can add.

maxkfranz:
Yes, Igor's right.  You need a mount point to store permanent data outside of the container

Issue #452, opened 2019-04-12
Fatal error attempting to add existing database 


jvwong:
Booting up the app when the rethinkdb database/tables don't already exist throws an error: `ReqlOpFailedError: Database `factoid` already exists in:
r.dbCreate("factoid")`. 

To reproduce this, clear out the 'factoid' database and start the app. Strangely, the error doesn't always appear - no idea. I  originally noticed this behaviour trying to couple the app and Rethinkdb in docker containers. 

- node v10.15.3
- unstable branch (58dfff0fda9ca8f039ea940945121e9ada613469)

maxkfranz:
Yeah, it's a known issue.  A newer version of the rdb driver may help.

PR #481, opened 2019-05-22
Adding volumes.

jvwong:
Adding volumes for elasticsearch and rethinkdb.


docker-compose.yml
```diff
@@ -27,6 +26,15 @@ services:
...
+    volumes:
+      - esdata:/usr/share/elasticsearch/data
+  # db: # NB: https://github.com/PathwayCommons/factoid/issues/452 
+  #   image: rethinkdb:2.3.6 
+  #   volumes:
+  #     - dbdata:/data
+
+volumes:
+  esdata:
+  # dbdata:
```
maxkfranz: Why is dbdata commented out?

jvwong: That bug #452 seems to cripple the app.

maxkfranz: If you can reproduce it every time with that docker setup, then it would be a good opportunity to try to find a workaround for it.  I can't reproduce the issue reliably on my machine.

jvwong: It happens on the ci server (`...128`) - i set up a mini docker-compose file in `/home/baderlab/tmp`.

jvwong: addressed by #482

Issue #493, opened 2019-06-05
Update docker-compose and documentation


jvwong:
The compose needs a few tweaks to be more user-friendly. 

- Most variables (ports, tags) are hardcoded; can look into env file for more flexibility
  - ideally tags would reference releases of our software
  - database credentials
- Update documentation to include versions, use and notes

PR #494, opened 2019-06-05
Updates to docker-compose and documentation

jvwong:
Added 
  - variables to configure the app image tags and ports
  - expose database admin

Refs #493


jvwong:
- Added backup and restore scripts for data volumes 
  - Output is simple gzip archive
  - Avoids need for massive images (elasticsearch) and buggy data dump (rethinkdb)

Issue #509, opened 2019-06-19
Disable document creation via GET


jvwong:
A new document can be created simply by navigating to  `/api/document/:id` with a dummy `id`. I suggest to disable this as should never be used (in production anyways) and is possible source of orphaned docs in database.

maxkfranz:
That should be deleted completely at this point.

PR #523, opened 2019-07-10
Remove routes that create docs automatically and prevent the editor from automatically creating docs

maxkfranz:
Ref : Disable document creation via GET #509

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4192,8 +4192,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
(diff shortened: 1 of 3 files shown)


maxkfranz:
Note: This means you just get an exception and a blank page for now if you go to something like /document/foo/bar.  Later we could catch the exception and put an error message in the UI, but this small change should suffice to solve the issue.

metincansiper:
Looks fine to me. As far as I understand viewing documents in form editor is also disabled by the commit. Am I right?

maxkfranz:
Yes.  Those routes would automatically create documents, and the form is already not working.  So the simplest solution is to just to remove those routes for now.  

The form code is still there if we need it in future.

Issue #520, opened 2019-06-25
Delete button


maxkfranz:
Add a button to delete the currently selected elements.  It should work the same as the existing feature that uses the delete key.

Issue #519, opened 2019-06-25
Difficult to delete a node : don't deselect a node when closing its popover


maxkfranz:
A user has to click outside of a node's popover to close the popover.  Because the node itself takes up little area, the user is more likely to click on the background rather than the node.  This normally deselects the node.

However, a user may have clicked an ungrounded node to delete it.  In that case, it is difficult to delete the node.  The user has to click the node a second time before the delete key works.

This issue can be mitigated by not mutating the selection state when a popover is closed.  It would also help to have a delete button that can be used after clicking the node once.

maxkfranz:
Ref : Delete button #520

PR #527, opened 2019-07-16
Make it easier to delete

maxkfranz:
- Adds a delete button  #520 
- Revises selection s.t. closing an element's popover does not deselect the element  #519

Demo of deleting after clicking the background using the delete key, followed by deleting by the delete button:

beam3


maxkfranz:
Yeah, the original "x" icon is probably better: 

Screen Shot 2019-07-16 at 4 13 19 PM

New commit

PR #529, opened 2019-07-30
Grounding query fixes

maxkfranz:
This PR includes minor fixes related to building queries that are sent to the grounding service.

### DIFF
--- a/src/model/document/document.js
+++ b/src/model/document/document.js
@@ -253,7 +253,7 @@ class Document {
     let json = {};
 
     for( let [org, count] of this.organismCounts() ){
-      json[ org.id() ] = count;
+      json[ org ] = count;
     }
 
     return json;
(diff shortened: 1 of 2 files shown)


jvwong:
Was there an issue or particular bug for this?

src/model/document/document.js
```diff
@@ -253,7 +253,7 @@ class Document {
     let json = {};
 
     for( let [org, count] of this.organismCounts() ){
-      json[ org.id() ] = count;
```
maxkfranz: This addresses the issue re. creating a new node that can't be grounded.

src/server/routes/api/element-association/grounding-search.js
```diff
@@ -38,6 +39,10 @@ const search = opts => {
     queryOpts.organismOrdering = organismOrdering;
   }
 
+  if( namespace ){
+    queryOpts.namespace = namespace;
+  }
+
   return searchAll( queryOpts )
```
maxkfranz: This change makes it so that we don't send `namespace: undefined` to the query service.

maxkfranz:
No github issues were created for these issues, but they are described in the comments above.

PR #531, opened 2019-07-30
Node flashes on compound node restored from "undo delete button" 

### LINKED ISSUE #515: Node flashes on compound node restored from "undo delete button"
The compound node flashes from a non-compound node to a compound node when it is restored by clicking the "undo delete" button.  It should be restored in one visual step.

Maybe `cy.batch()` would help.

metincansiper:
To fix #515


maxkfranz:
Batching makes sense.  Merging

Issue #530, opened 2019-07-30
Travis builds are broken


maxkfranz:
Travis no longer installs rethinkdb properly.

maxkfranz:
Maybe it's easy to fix with a new preset: https://docs.travis-ci.com/user/database-setup/#rethinkdb

PR #533, opened 2019-08-07
Revise .travis.yml to use addon rethink

maxkfranz:
Previously, we used a manual install of rethink on travis.  That no longer works.

Ref : Travis builds are broken #530

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,7 @@
-dist: trusty
 language: node_js
 node_js:
   - "8"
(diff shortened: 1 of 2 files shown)


maxkfranz:
Though this change allows travis to install rethinkdb properly, it does not connect in some cases.

maxkfranz:
Let's just disable the model tests on Travis for now.  At least Travis will be useful for linting.

Issue #516, opened 2019-06-25
Remove the grounding popover for complexes


maxkfranz:
Complexes won't be grounded so they won't use the same UI as we have now.  We also will be changing the entity info UI for this issue: https://github.com/PathwayCommons/factoid/issues/504

So for now, it's best to just disable showing the popovers when clicking a compound node.

jvwong:
Human nature being what it is, user might, given the freedom, use this feature to depict any number of biological concepts for which it was never intended.  

Consider:
  - cellular compartment
  - groups of genes or gene products (AND/OR)
  - generic entities
  - anything where people are used to seeing a box?

Wirth keeping this in mind if we're not explicitly disambiguating.

maxkfranz:
I think the compartment case is likely the most problematic one.

To play devil's advocate:  If we find users generally tend to put entities into compound nodes for both compartments and complexes, then could we use compound nodes for both concepts?  It may suffice to allow a user to "ground" a compound node to a complex or a compartment.

maxkfranz:
I'll take this one, since it's related to this issue: Ground a named node by default #504

Issue #504, opened 2019-06-18
Ground a named node by default


maxkfranz:
If the grounding service is of sufficient quality in getting the expected grounding as the first result, then the UI for grounding a node should be simplified:

(1) The node popover is opened.

(2) The text box and notification in the popover are the same as today.  The wizard arrows are removed.  The user enters the name in the text box, say "p53".

(3) The top grounding result is shown underneath the text box.  The user has the option to click a button to select a different grounding entry if the top result is wrong, e.g. "not this p53".  When the user clicks this button, a grounding can be chosen from a list.  

(4) The way that the alternate groundings are presented could be smart.  For example, if there are many p53 entries and the main difference between them is the organism, we could simplify the question of "Which p53 do you mean?" to "Which organism do you mean?"  If the main difference between the entries is type, then we could simplify the question of "Which ER do you mean?" to "Do you mean this ER protein or this ER chemical?" (two entries listed side by side).

In most cases, the user is finished at step (2).  This makes it so there is only one step for the user to ground a node, and it's the only step the users really want to do.  A user just wants to name a node and be done, and this feature allows him to do just that.

To start, (4) could just be a simple list.  Later it could be improved to be smarter.

jvwong:
This is related to a grounding test analysis performed on July 8, 2019, the details of which were in an email to Sorger group:

![Fig 1  Grounding summary ](https://user-images.githubusercontent.com/4706307/63186001-5728e200-c029-11e9-9052-24296b665b82.png)

![Fig 2  Failures_ Rank in returned records ](https://user-images.githubusercontent.com/4706307/63186012-5b54ff80-c029-11e9-9b70-0e4748ff3768.png)

![Fig 3  'Absent' failures ](https://user-images.githubusercontent.com/4706307/63186014-5d1ec300-c029-11e9-96b8-27e4e9f0b8b9.png)

jvwong:
See #550.

PR #536, opened 2019-08-16
Automatic grounding of entities (initial version)

maxkfranz:
This revises the UI as per #504 for items (1), (2), and (3).  

The user starts by only typing the name, without anything else to distract him:

Screen Shot 2019-08-16 at 12 48 44 PM

The user types the name of the entity and the best-guess grounding is used by default:

Screen Shot 2019-08-16 at 12 48 51 PM

In all but a small minority of cases, the expected grounding is correct.  If the user does not like the grounding, he can click the button to select a better match:

Screen Shot 2019-08-16 at 12 51 06 PM

If the user types something for which we can't find any possible groundings, we display an error message:

Screen Shot 2019-08-16 at 1 01 36 PM

The UI for complexes is consistent with the automatic grounding UI:

Screen Shot 2019-08-16 at 12 54 46 PM

The UI for interactions is also consistent:

Screen Shot 2019-08-16 at 12 56 58 PM

Related changes:

- Remove support for grouped selection via interaction nodes -- we don't support interaction nodes for n-ary interactions anymore.  Without this change, there are conflicts with automatic grounding.
- Update labels for "entity" to instead use the words "gene or chemical".
- Fix a related selection issue.  When clicking from one node to another node, the old node remains selected.
- Remove the extra spacing for interaction popovers.
- Complexes now have a label, to be consistent with other complete entity nodes.
- Notifications are removed, since they are not relevant with an automatic grounding UI.
- The submission task list excludes complexes and interactions.  They are always submittable.

Ref: 
- Ground a named node by default #504 : (1), (2), and (3) complete
- Remove the grounding popover for complexes #516 : No longer applicable

### DIFF
--- a/src/client/components/editor/buttons.js
+++ b/src/client/components/editor/buttons.js
@@ -41,7 +41,7 @@ class EditorButtons extends React.Component {
       );
 
       grs.push([
-        h(Tooltip, _.assign({}, baseTooltipProps, { description: 'Add an entity', shortcut: '1' }), [
+        h(Tooltip, _.assign({}, baseTooltipProps, { description: 'Add a gene or chemical', shortcut: '1' }), [
           h('button.editor-button.plain-button', { onClick: () => controller.addElement().then( el => bus.emit('opentip', el) )  }, [
             h('i.material-icons', 'fiber_manual_record')
           ])
--- a/src/client/components/editor/cy/compound-dnd.js
+++ b/src/client/components/editor/cy/compound-dnd.js
@@ -1,6 +1,6 @@
-import { tryPromise } from '../../../../util';
+import { tryPromise, delay } from '../../../../util';
 
-export default function({ cy, document, controller }){
+export default function({ cy, document, controller, bus }){
     if( !document.editable() ){ return; }
     cy.compoundDragAndDrop();
     let lastOldParent = null;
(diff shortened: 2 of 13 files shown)


jvwong:
- Case 1: auto-ground to top hit
  - Works for me
  - Here, the user is thinking,  "My text name matched the information they displayed (wherever that's from) so i'm done"
- Case 2: 'Select a better match'
  - Do users need to be prompted to actually click something? Should we indicate the grounding that will be applied? 
  - Users are already in the 'autocomplete' mindset.  When their label is not the top hit, they could be thinking  "Hmm like they don't like my name. I need to choose a better one" and you could get the situation below (second hit is the one I want):

image

jvwong:
A related minor point in case 2 is that the 'rank' of an applied ground is always the same. Meaning if I selected one in the middle of the list, then go back and try to ground again, it 'jumps' back into the same rank position.

jvwong:
Minor bleed in interaction tooltip button:

image

jvwong:
Also some bleed on the node tip:

image

maxkfranz:
OK, let's fix those issues in separate PRs.  I'm merging the feature into the unstable branch so it's publicly testable at unstable.factoid.baderlab.org

Issue #551, opened 2019-09-17
Entity subtypes : `ggp`, `dna`, `rna`


maxkfranz:
## Description

**Q: What is the name of the feature?**

A: Support for DNA, RNA, GGP.

**Q: What does this feature enable the user to do?**

A: The user can select the subtype for an  NCBI grounding.

**Q: What information must the user provide to use the feature?**

A: Nothing is mandatory, as there is a default type.  The user can override the selection with one of `ggp`, `dna`, `rna`, `protein`.

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: The default should be reasonable.  The default is whatever comes back from the grounding service.  This means that the grounding service should return groundings of type `ggp` for NCBI groundings, rather than `protein`.

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: Better express his paper
- Editor: Display more detailed figures
- Computational biologist: More refined data
- Curator: "

### Mockup

The user should be given a selector for the subtype:

```
( Gene or gene product | DNA | RNA | Protein )
```

Only one can be selected at once.

### Details

The horizontal toggle allows the input element to fit within the grounding summary UI.

maxkfranz:
Preview:

<img width="495" alt="Screen Shot 2019-09-18 at 9 52 07 AM" src="https://user-images.githubusercontent.com/989043/65154709-fe6ab180-d9f9-11e9-8d26-203456a847f2.png">

PR #552, opened 2019-09-17
Add support for entity types `ggp`, `dna`, `rna`

maxkfranz:
- Add model support.
- Add display support based on existing protein implementation.
- Add a subtype selector for NCBI groundings.

Ref : Entity subtypes : `ggp`, `dna`, `rna` #551


maxkfranz:
> Default should probably be ggp, but lets merge.

That's a change for the grounding service.  The UI just uses whatever comes from the service; it doesn't have its own defaults.

Issue #547, opened 2019-09-09
Testing complexes


jvwong:
Initial observations for complexes. 

Referencing #488 as example. Background:

> - AMP-activated protein kinase [(AMPK)](https://en.wikipedia.org/wiki/AMP-activated_protein_kinase) is a heterotrimeric protein complex that is formed by α, β, and γ subunits. - 
>   - The following human genes encode AMPK subunits:
α – PRKAA1, PRKAA2
β – PRKAB1, PRKAB2
γ – PRKAG1, PRKAG2, PRKAG3
> - AMPK is activated by CaMKK2

Desired result:
![image](https://user-images.githubusercontent.com/4706307/64543705-b1992380-d2f3-11e9-9f4b-94f893a1f903.png)

---

Bug:

- [x] (1) Disappearing box upon participant removal
  - Recreate
    - Create a three participant complex. 
    - Upon removing a single participant box disappears
    - Upon page refresh, box reappears around remaining pair of participants
  - Desired: The box should remain when >= 2 participants upon removal of additional entities

User experience:

- [ ] (2) Bounding region for adding/creating complex is too generous
  - Recreate
    - Create a complex, allowing participants a moderate amount of space in between 
    - Add an entity, not intended for the complex (e.g. Exo I below)
    - Dragging the additional entity around the screen often adds itself to the complex
  - Desired: Adding a participant should require greater intention on the user's part. 

  - ![image](https://user-images.githubusercontent.com/4706307/64544655-5ff19880-d2f5-11e9-9327-50a9dbbdd415.png)

maxkfranz:
Re. (1): I think this is related to https://github.com/PathwayCommons/factoid/pull/535, and it's probably just a style issue.

Re. (2): That's a general issue with large compound nodes.  Applying a layout to make the nodes use less space helps with this, but it removes some freedom for the user.  Another option is to only add or remove after a delay, but that also has some downsides.

maxkfranz:
(1) is addressed and (2) is addressed to some degree.  We'll have to evaluate whether more is needed for (2) before closing this.

PR #553, opened 2019-09-17
Complex fixes

maxkfranz:
- Fix for missing complex label.
- Fix for overly springy complexes.  We may have to adjust `outThreshold` with more testing.
- Fix for the display list of complex subunits in the popover.  If the entity is ungrounded, it has no display name.
- Fix for prematurely disappearing complex box.


maxkfranz:
Ref #547

Issue #550, opened 2019-09-10
User testing: Entity grounding  


jvwong:
## Goal  
User-test entity grounding user interface.

## Setup
Brief sentence containing entity that must be created. Declare species explicitly. In-person observation.

### Approach
- Tests
  - Simple; Top rank
  - Intermediate rank
  - Low rank

- Evaluation
  - Ground applied
  - Route to grounding

#### Passages 
Entity of interest in **bold** 

- *In humans, receptor-linked tyrosine kinases such as the **epidermal growth factor receptor (EGFR)** are activated by extracellular ligands.*
  - Expect NCBI:1956, protein
- *Drosophila **Myc mRNA transcript** is encoded by the dm gene (diminutive, from the mutant phenotype).* 
  - Expect NCBI:31310, RNA
- *The E. coli **rela** enzyme catalyzes the formation of pppGpp from ATP.*
  - Expect NCBI:7150942, protein

jvwong:
Feedback from 2 users ('A' and 'B'; computational):

- General
  - users found the entity search is fast and smooth
- Type selection ( RNA, Protein, etc. -  #551 )
  - users could easily identify, understand and use the selection UI effectively
- Top hit is incorrect
  - The test case is selection of  Drosophila Myc, typically human comes up
  - Both users attempted to click the 'Organism' text in tooltip to change this
  - Button `This isn't the X that I meant`
    - User A never noticed the tooltip button to display ground search hits
    - User B eventually noticed
      - Thought it might be 'delete' based on text
      - Offered up different text ideas `Other choices`; `More results`

PR #567, opened 2019-10-17
Autoground tweaks

maxkfranz:
Changes:

- The wording and salience of the grounding revision button have been revised.
- A (rare) grounding with an empty synonym list has "-" shown instead of whitespace.
- Allow specifying the grounding by organism dropdown.  This turned out to be a smaller change than expected, so I'm including it here.

Ref : User testing: Entity grounding #550

-- 

Preview:

beam11


jvwong:
I want to look at this more closely  - in the meantime, I seem to be getting a bunch of 'undefined' values for the drop down. Is there something I should know about setup?

Issue #524, opened 2019-07-10
Document (contributor) admin


jvwong:
- Dashboard summary of documents 
  - complete, incomplete, submitted
  - metadata: author name, paper name
  - link to document viewer, etc.
  - metrics : measure how successful the process / editor are
- View document edit history
- Automate process (e.g. sending emails at right times, reminder emails)

maxkfranz:
See https://github.com/PathwayCommons/factoid/tree/management-panel

That branch has a stubbed-out admin page

jvwong:
Stuff we might want to store:

- context: How that document came to be created, for instance, via signup page or journal
- correspondence: Did an email invitation/reminder get sent? 
- ...

maxkfranz:
Ref : https://github.com/PathwayCommons/factoid/blob/f34b221b9c82379e69b4d182a5625ee65569fe17/src/client/components/document-management.js#L143-L176

jvwong:
#573 will break the admin panel temporarily but that's fine because we should settle on a model JSON schema before fixing it.

PR #571, opened 2019-10-24
Management panel

jvwong:
Addition of a document management admin panel:
- List article, document views and document status/timestamps/submitted(check)
- Sending email (invitation) is enabled
- Disabled `/new/seed` email

Refs #524 

image


jvwong:
Using human readable  timestamps:

image

Issue #570, opened 2019-10-24
PubMed modules for extracting metadata from PMID and Title.


jvwong:
## Description

**Q: What is the name of the feature?**

A: PubMed metadata retrieval from PMID 

**Q: What does this feature enable the user to do?**

A: Reduces the requirements for signup 

**Q: What information must the user provide to use the feature?**

A: Allowing authors to identify the paper via PMID 

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: Must be indexed by PubMed; Introduces external dependency

**Q: How does this feature affect each class of user (persona)?**

A: Authors. Enabling submission of a URL/PMID through the homepage greatly reduces the barrier to entry. Reduces amount of work for the user and error. Leverages the rich article metadata provided by NLM and PubMed. Enables associating with additional relevant and useful data (e.g. references, keywords, gene2pubmed etc ).

## Specification

- Inputs
  - Accept valid 
    - UID(s) from PubMed
      - PMID(s)
    - PubMed URL with path referencing 
       - PMID (e.g. `https://www.ncbi.nlm.nih.gov/pubmed/21292160`)
       - Exact, unique title (e.g. `https://www.ncbi.nlm.nih.gov/pubmed/?term=PTEN+Suppresses+Glycolysis+by+Dephosphorylating+and+Inhibiting+Autophosphorylated+PGK1`)

- Outputs
  - JSON containing subsets of full records for each UID 
    - See https://dtd.nlm.nih.gov/ncbi/pubmed/out/pubmed_190101.dtd 

- Edge cases
  - Unsupported inputs
  - Input URLs that reference more than one UID
  - PMCID is a bit of nightmare so maybe just double back to the PMID
  - erroneous id strings: `id=9751xd059` is interpreted as `id=9751`

## Notes

> On December 1, 2018, NCBI will begin enforcing the use of new API keys for E-utility calls.

jvwong:
Steps:
  - EFetch mapper module/test
  - ESearch module/tests
  - wrapper for services

PR #572, opened 2019-10-28
Data mapper for NCBI PubMed XML

jvwong:
Takes XML output from the NCBI EFetch web service utility for a PubMed article and outputs a JSON object that roughly approximates it. Nulls out fields that do not exist in raw data.

Refs #570 , Part I


maxkfranz:
Merging this in for now.  For future consideration & discussion:

- Performance of xml file parsing
- Live fetching from pubmed
- Server side routes:  To be used by home page & admin panel

PR #573, opened 2019-10-29
Demo and secrets

maxkfranz:
Add support for a demo route & enforce secret existence on create()

- Add a /demo route that always resolves to an editor view that has a document ID of 'demo' and secret of 'demo'.  If the document does not exist, it will be created by accessing this URL.
- Enforce secret existence prior to client `syncher.create()` requests.  This makes it impossible for a client to create synched objects, unless the provided secret already exists within the 'secret' table.  An entry is added to the secret table
- The demo document is excluded from POST /document listings.
- You can DELETE /demo to remove all data that uses the demo secret.  This removes the demo document and secrets.  This requires the API key.


maxkfranz:
@gbader @cannin 

With this change, you will be able to do demos at path `/demo` without having to worry about affecting production data.

src/server/routes/api/document/index.js
```diff
@@ -118,9 +127,16 @@ http.get('/', function( req, res, next ){
...
     .then( t => {
-      let { table, conn } = t;
-
-      return table.skip(offset).limit(limit).run(conn);
+      let { table, conn, rethink: r } = t;
+      return table
+        .skip(offset)
+        .limit(limit)
+        .filter(r.row('secret').ne(DEMO_SECRET))
+        .pluck(['id'])
```
jvwong: This will break the admin panel, as it previously expected the array of doc JSON. Was this on purpose?

maxkfranz:
The new commits address the comments, as we discussed in person.

src/server/routes/api/document/index.js
jvwong: OK this is fine. I'll update the admin and these fields in another PR.

src/server/routes/api/document/index.js
```diff
@@ -147,6 +214,37 @@ const checkApiKey = (apiKey) => {
...
+  let { apiKey } = req.body;
+  let secret = DEMO_SECRET;
+
+  let clearDemoRows = table => table.filter({ secret }).delete();
+
+  ( tryPromise(() => checkApiKey(apiKey))
+    .then(loadTables)
+    .then(({ docTable, eleTable }) => (
+      Promise.all([docTable, eleTable].map(clearDemoRows))
+    ))
```
jvwong: Its complaining that `TypeError: Cannot read property 'filter' of undefined` for `factoid/src/server/routes/api/document/index.js:236:38`

Issue #577, opened 2019-11-06
GET /api/document error when documents contain non-empty entries


jvwong:
Doing a GET on /api/document, particularly with documents having non-empty elements, throws an error. I would expect it to return a list of document JSON objects.

Reproduce:
- create a document
- add an entity
- hit the GET `/api/document` endpoint

```
TypeError: obj.json is not a function
   at toJson ...factoid/src/model/document/document.js:401:29)
```

jvwong:
Nevermind, I keep hitting the 'Gotcha' with GROUNDING_SEARCH_BASE_URL

PR #578, opened 2019-11-06
Error in api/document GET returning json list of docs

jvwong:
I believe the document `organisms` function just returns a list of strings.
Refs #577


maxkfranz:
I encountered the same issue on a share feature branch. My fix is the same, so let’s just merge this in for now if the fix is useful for your work on the admin panel.

PR #581, opened 2019-11-11
A pubmed service wrapper 

jvwong:
Expose `getPubmedArticle` that takes a `paperId` (e.g. signup). Tries to figure out if there's a Pubmed UID in the input and return the PubmedArticle metadata. 
  - Case 1: Demo, where input is exactly `demo`
  - Case 2: Numeric input, like a UID  e.g. '29249655'
  - Case 3: URL-like input starting with `https://www.ncbi.nlm.nih.gov/pubmed/`
    - a: Contains numeric item in path  e.g. `https://www.ncbi.nlm.nih.gov/pubmed/29249655`
    - b: Contains `term` URL search parameter with unique Pubmed Search result  
      - e.g. `https://www.ncbi.nlm.nih.gov/pubmed/?term=CtIP-Mediated+Fork+Protection+Synergizes+with+BRCA1+to+Suppress+Genomic+Instability+upon+DNA+Replication+Stress` 
      - e.g. `https://www.ncbi.nlm.nih.gov/pubmed/?term=30344097`

refs #570

### DIFF
--- a/src/server/routes/api/document/pubmed/demoPubmedArticle.js
+++ b/src/server/routes/api/document/pubmed/demoPubmedArticle.js
@@ -1,0 +1,277 @@
+export default {
+  "MedlineCitation": {
+    "Article": {
+      "Journal": {
+        "Title": "Journal of Example",
+        "Volume": "1",
+        "Issue": "1",
+        "PubDate": {
+          "Year": "2019",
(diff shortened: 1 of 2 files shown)


jvwong:
Node 8 doesn't support named capture groups for Regex, but 10 does.

Issue #627, opened 2020-02-03
Add-back 'Submit'


jvwong:
## Description

**Q: What is the name of the feature?**

A: Submit (document) button

**Q: What does this feature enable the user to do?**

A: Enables a user to signify that the contents of their document is complete. 

**Q: What information must the user provide to use the feature?**

A: Action: Click button and confirm. 

**Q: How does this feature affect each class of user (persona)?**

A: This issue emerged from user testing:

> - "'Share' sounds like publish"; "'Submit' is clearer"
> - Submit button? 
> - "What do I do now"? "How do I save"?
> - How do I know if it is saved? I want to save it.

### Mockup

A 'Submit' button existed before it was [replaced with the 'Share'](https://github.com/PathwayCommons/factoid/commit/6711cfa8ac032cbf7f03d0ec19716823bf96dbfa#diff-6d22f404077099f53dfe11bd2b197e13). Should update the onboarding cartoon to reflect changes to the workflow.
 

### Details

Flow control
  - User clicks Submit
  - Confirmation message shown
  - Upon confirmation,
    -  Submit is disabled and text replaced with 'Submitted'
  - Optional
    - Share to @biofactoid Twitter - but what about the 'blurb'?
  - Follow-up email sent 

NB: I don't believe that the act of settings a document 'submitted' status should be coupled to email sending and/or posting to Twitter. Consider an admin page, where we wish to alter the doc status flag or other cases we wish to work 'behind the scenes'.

Related to #620, #592, #532

maxkfranz:
Possible improvements:

- [ ] The post-submission text could be revised in style and content. It would be nice to have a checkmark, and the whitespace at the top and bottom of the panel could be removed. The text itself seems a bit conditional to me ("should" versus "will"). The actions at the bottom could be separated and made more direct.
- [ ] It would be good feedback to the user for the button to turn green after submission, especially if you consider the action to be "publish".
- [ ] We may want to go directly to "published" and skip "submitted" by default. It's better for a user to see his summary right away, and we can manually demote a document from "published" to "submitted" using the admin interface --- as needed.

jvwong:
NB: 

I thought this was a bug but it may be a feature: If the editor is open, an admin-driven update to 'status=submitted' results in a follow-up email request. This is probably OK.

PR #630, opened 2020-02-07
Add back submit

jvwong:
- Add back the old submit button to editor (editable only)
  - I left the all the validation code that tells you if something isn't 'finished' but doesn't block submit 
  - No button on read-only view
- Button tracks the document 'status' field (e.g. approved, submitted, trashed)
- Addded new doc 'status' value: published, in order to have granularity with what might be released publically

Refs #627

Doc status: requested, approved
image

Doc status: submitted, published
image

### DIFF
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -22,7 +22,7 @@ import * as defs from './defs';
 import EditorButtons from './buttons';
 import MainMenu from '../main-menu';
 import UndoRemove from './undo-remove';
-import { ShareView } from '../share';
+import { TaskView } from '../tasks';
 
 const RM_DEBOUNCE_TIME = 500;
 const RM_AVAIL_DURATION = 5000;
@@ -475,7 +475,7 @@ class Editor extends DataComponent {
   }
 
   render(){
-    let { document, bus, showHelp, cy } = this.data;
+    let { document, bus, showHelp } = this.data;
     let controller = this;
     let { history } = this.props;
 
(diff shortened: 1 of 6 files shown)


jvwong:
I'll put up a separate PR for automating emails on submit.

maxkfranz:
(1) The incomplete item count seems to be off by one in the example.

(2) The post-submission text could be revised in style and content.  It would be nice to have a checkmark, and the whitespace at the top and bottom of the panel could be removed.  The text itself seems a bit conditional to me ("should" versus "will").  The actions at the bottom could be separated and made more direct.

(3) It would be good feedback to the user for the button to turn green after submission, especially if you consider the action to be "publish".

(4) We may want to go directly to "published" and skip "submitted" by default.  It's better for a user to see his summary right away, and we can manually demote a document from "published" to "submitted" using the admin interface --- as needed.

(5) It may make sense to use a separate PR for these items.

We can discuss in more detail tomorrow.

jvwong:
> (1) The incomplete item count seems to be off by one in the example.

I think there are two things worth noting here:
  - a) The interaction type defaults to 'Other' but needs to be clicked to be stored; Otherwise it's incomplete which I never noticed before
  - b) The complex isn't named by default ; I don't think this should be a requirement (someone suggested just put the names of the children, which is kinda redundant in simple cases)

maxkfranz:
From discussion: (1) will be addressed in this PR, and the other improvements have been noted in #627 for a future PR.

Issue #622, opened 2020-01-24
Explore view: Basic version


jvwong:
## Description

**Q: What is the name of the feature?**

A: Explorer view

**Q: What does this feature enable the user to do?**

A: Visualize a single pathway and obtain information related to the article. 

**Q: What information must the user provide to use the feature?**

A: Nothing in this version will require input.

### Mockup

// In progress 

### Details

- Information
  - Citation 
    - Title
    - Authors
      - Email(s)
    - Reference info (Journal, volume, year)       
- Links
    - journal website (i.e. via doi.org)
    - PubMed 
      - article 
      - author search

PR #636, opened 2020-02-14
Explore View: Basic

jvwong:
Enhance the editor (non-editable) view with relevant links in the title.

- Links
  - Article title link points to journal by resolving `https://doi.org/<DOI>`
  - Author links point to PubMed author search (e.g. 'Wong JV[Author]')
  - Email icon, if exists, does a link for `mailto:<EMAIL>`
- Component: 'expandable list'
  - configure # leading, trailing items, toggle button etc
  - calculates if and when the toggle can be displayed at all   

Notes: You can see the limitation of the explore view when there are a lot of authors. I wouldn't try to handle this now as we're likely pushing for a 'mobile' first approach anyways.
 
Refs #622 

Below are some examples when I configure the component to show 3 leading and 2 tailing authors in a simulated phone screen (iPhone 8 plus). 

- Left:  Less than the number of leading + tailing authors
- Middle:  A reasonable number of authors
- Left:  Massive number of authors 

image

### DIFF
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -5,8 +5,11 @@ import EventEmitter from 'eventemitter3';
 import io from 'socket.io-client';
 import _ from 'lodash';
 import Mousetrap from 'mousetrap';
+import queryString from 'query-string';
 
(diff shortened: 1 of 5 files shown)


jvwong:
let's just keep the branch and close this.

PR #639, opened 2020-02-18
Submit: Publish & Email follow-up notification

jvwong:
Upon 'Submit' in editor, attempt to 'publish' and send follow-up email notification automatically.
  - Email notification only occurs when 'Submit' button is clicked in the editor. This avoids a case where updates happen elsewhere, such as the  admin
  - Setting a doc to 'published'   
    - Must not be empty
    - Must not have incomplete entities/nodes
    - Must have status='submitted' 

Refs #627 

> We may want to go directly to "published" and skip "submitted" by default. It's better for a user to see his summary right away, and we can manually demote a document from "published" to "submitted" using the admin interface --- as needed.
>  Note: This may require some basic, automated sanity checks (e.g. don't automatically publish if there are blank nodes).

### DIFF
--- a/src/client/components/tasks.js
+++ b/src/client/components/tasks.js
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import DataComponent from './data-component';
 import h from 'react-hyperscript';
+import queryString from 'query-string';
 
 import { BASE_URL } from '../../config';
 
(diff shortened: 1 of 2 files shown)


maxkfranz:
Note: We should put a TODO as a reminder for merging the PATCH routes

PR #640, opened 2020-02-18
Indra interaction

metincansiper:
Creating the PR from my experimental feature branch as we discussed today. 

- The source of python server to query indra edges is available here (https://github.com/metincansiper/IndraInteractionServer). I run the server in command line interface by ``python server.py`` (I use python3).
- About the UI: I have just listed the possible interactions as how the possible groundings are listed for the entities in Factoid. While doing that I re-used the related components designed for entities and named accordingly. If we need to use the same components (based on how will the final UI decision be) it would be useful to rename that components.


maxkfranz:
@metincansiper, would you be able to set up the dockers for these changes?  It would also be good to have instructions in the readme.

metincansiper:
@maxkfranz I started with implementing the dockerfile for the webservice. It builds and runs but for some reason when I try to navigate the related port in the browser it displays the followings:

```
This page isn’t working
localhost didn’t send any data.
ERR_EMPTY_RESPONSE
```

I commited the docker file to the github (https://github.com/metincansiper/IndraInteractionServer). I build and run the docker file as the following:

```
// build
docker build -f Dockerfile -t indra_intn_server .
// run
docker run -it -p 8000:8000 indra_intn_server
// I tried the following as well but no change
docker run -it -p 127.0.0.1:80:8000 indra_intn_server
```

Then I visit http://localhost:8000/find-interactions?source=MDM2&target=EGFR&sign=N to check if everything is fine. Looks like server gets no data it must be related to the warning of "localhost didn't send any data".

It must not be the case but I still suspect that it may be related to the OHSU network policies. 

- Can you check if that works for you?
- If it does not works for you as well: do you have any idea of what would be the problem?

jvwong:
Maybe try: HTTPServer address should be '0.0.0.0' instead of '127.0.0.1'/loopback inside of the container

https://github.com/metincansiper/IndraInteractionServer/blob/77bd910d5ff01abd6bf66717472c594a542dcdf4/indra-interaction-server/server.py#L46

metincansiper:
@jvwong thanks! Looks like that solves the problem that I described. I will solve the remaining problems and make a commit.

metincansiper:
@jvwong @maxkfranz 

I set up docker for these changes. However,

- I used the indra image from the local docker build as ``image: indra_intn_server:latest`` (https://github.com/PathwayCommons/factoid/pull/640/files#diff-072e81fe88d9b23fe00fc189279936a6R57). I guess eventually we will need to have it available in PC dockerhub. After having it there I can update that line.
- Using indra db/api requires some credentials (https://indra.readthedocs.io/en/latest/modules/sources/indra_db_rest/#module-indra.sources.indra_db_rest). We will talk to indra developers about that soon. In docker file of indra interaction server I set that as a build argument.

metincansiper:
BTW I got the following error when I use the grounding search from the docker in Factoid:

```
grounding_container | 2020-03-05T00:18:38.839008900Z (node:18) UnhandledPromiseRejectionWarning: Error: [index_not_found_exception] no such index, with { resource.type="index_or_alias" & resource.id="groundingsearch" & index_uuid="_na_" & index="groundingsearch" }
grounding_container | 2020-03-05T00:18:38.839082100Z     at respond (/home/appuser/app/node_modules/elasticsearch/src/lib/transport.js:308:15)
grounding_container | 2020-03-05T00:18:38.839145100Z     at checkRespForFailure (/home/appuser/app/node_modules/elasticsearch/src/lib/transport.js:267:7)
grounding_container | 2020-03-05T00:18:38.839177500Z     at HttpConnector.<anonymous> (/home/appuser/app/node_modules/elasticsearch/src/lib/connectors/http.js:166:7)
grounding_container | 2020-03-05T00:18:38.839213000Z     at IncomingMessage.wrapper (/home/appuser/app/node_modules/lodash/lodash.js:4935:19)
grounding_container | 2020-03-05T00:18:38.839246400Z     at IncomingMessage.emit (events.js:194:15)
grounding_container | 2020-03-05T00:18:38.839282100Z     at IncomingMessage.EventEmitter.emit (domain.js:441:20)
grounding_container | 2020-03-05T00:18:38.839315400Z     at endReadableNT (_stream_readable.js:1125:12)
grounding_container | 2020-03-05T00:18:38.839348800Z     at process._tickCallback (internal/process/next_tick.js:63:19)
grounding_container | 2020-03-05T00:18:38.840270700Z (node:18) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 9)
```

It works fine when I use it from where it is deployed.

jvwong:
I sent you a DockerHub invite for the PC account. Maybe get a tagged version up there.

metincansiper:
@jvwong Thanks! We have talked about this with Indra developers today. They tell that there is a rest api for indra as well (http://www.indra.bio/rest_api/docs/). I was aware of it but I was not aware of that we can use it because I guess the method that we need exists but not documented. I think switching to that service would be better anyways (if possible) to eliminate an extra python webservice. I am working on experimenting that. 

BTW initially I mentioned that using indra db client (that I am trying to replace with the rest api) requires an api key but if I understand correctly Ben mentioned that it must be working without a key now.

metincansiper:
I created a repository for the indra server under PC docker hub and updated this branch respectively. 

I also made some experiments on utilizing the related indra services through indra web api without needing an extra python server. Getting the relations between two genes in any type is available as ``https://db.indra.bio/latest/metadata/relations/from_agents?agent0=TGFBR&agent1=SMAD2``. In that link the specific indra statements are collected under generic ones as ''relations". For example ``Activation(MDM2(bound: [MIP, True]), TP53())`` would be collected under ``Activation(MDM2(), TP53())``. The relations contains a list of hashes for indra statements under that relation. Then to get the actual statements and pubmed id of them we need to use these hashes by requesting a service like ``https://db.indra.bio/latest/statements/from_hash/-32929621824215888?with_english=true&format=json-js`` where ``-32929621824215888`` is the hash id (this must be done for each statement). Making such queries for each sentence decreases the performance significantly. Using the python webservice (https://github.com/metincansiper/IndraInteractionServer) yields a much better performance under that conditions. I think using the other indra services would be practical only if we would use only the generic relation so would not visit the actual indra statements. In that case besides rendering only the generic sentences we also would not have access to the evidences so no access to any pubmed id etc. Therefore, I did not discard the related code (``https://github.com/metincansiper/factoid/tree/indra-intn-web-api``) in case it would be needed but I think it is not likely to be utilized. You can let me know if you have any further idea about which way to go.

@maxkfranz @jvwong would you let me know when the related UI design is decided.

maxkfranz:
The official Indra webservice takes the better part of 10 seconds for `from_agents` and each `from_hash` query takes an additional second.  How does this compare with your Python server?

metincansiper:
Actually the problem with the initial way to use the service was not about the response time of a single query. The problem was that we were needing to make another ``from_hash`` query for each statement which was being inefficient. 

However, now I realized that the indra module that I was using in my server is actually utilizing a query like (https://db.indra.bio/statements/from_agents?offset=0&agent0=EGFR&agent1=MDM2). In this way there is no need to make seperate queries for each hash so we can utilize the same service together with the english assembler of indra (http://api.indra.bio:8000/assemblers/english). 

Indra developers also mentioned another query type like ``from_hashes``. An example from them is:

```
fetch('https://db.indra.bio/latest/statements/from_hashes?format=json-js', {
  method: 'POST',
  body: JSON.stringify({hashes: ['hash1', 'hash2', 'hash3', ....]}),
  headers: {'Content-Type': 'application/json'}
});
```

They are not sure which one is faster but probably both of them would work for us. Therefore, there must be no need for the python service like I have implemented. I will update the implementation here based on what I have mentioned above (I will use one of the two ways).

metincansiper:
I updated this PR so that it uses the indra webservices for getting statements from agents (https://db.indra.bio/statements/from_agents?offset=0&agent0=EGFR&agent1=MDM2) and assembling english sentences from indra statements (http://api.indra.bio:8000/assemblers/english). This PR no more uses the webservice that I have implemented in the past.

I can start the actual implementation once the UI details are decided.

maxkfranz:
There are two tasks for now:

1. Any remaining issues re. docker deployment of indra.  @jvwong, does the current docker config suffice?
1. Writing a service layer
    1. Let's remove the changes to the interaction info component for now.  The focus is on the explore view.  The first feature that indra can be useful for in the explore view is related papers.
    1. The main function that we need should be getting papers related to a particular factoid document:
        1. This can be an http route, under something like GET `api/document/some-doc-id/related-papers`
        1. Input: factoid document id
        1. Output: List of pubmed ids (might be nice to also include metadata for each id, but not important to start with).  Let's keep it simple and just return the top 20 ids to start.
        1. Implementation: Score papers by using the pubmed metadata (entity name lists) and indra.  A simple metric w.r.t. pubmed metadata would be the size of the intersection between the entity list in pubmed and the entity list in factoid.  @jvwong can provide more info re. the pubmed data. A simple metric w.r.t. indra would be the intersection of the interaction list from indra and the interaction list from factoid.  If indra does not work for a document (e.g. non-human organism), then you can use reach on the abstract text from pubmed.  Each of those metrics could be values ranging on [0, 1], and they could be combined into a single value by a weighted average.

maxkfranz:
@metincansiper An alternative for 2.ii.d. would be to use reach or indra for the entity set.  Anyway, the implementation details are up to you.  The listed implementation ideas are just some possibilities.  Any implementation that scores the related papers effectively is fine.

jvwong:
>     1. Any remaining issues re. docker deployment of indra.  @jvwong, does the current docker config suffice?

This doesn't touch the `docker-compose.yml`. This is calling `indra.bio` domain now. @metincansiper Emek said on the call yesterday that you guys were working on creating self-contained Docker images for these services. Is that accurate?

maxkfranz:
I was under the impression that the docker related changes included both
the python server and the official Indra server.  If it doesn't already
have the official Indra as a docker image, then we need that added.
Ideally, the Indra team would maintain an official image on Docker Hub.

maxkfranz:
To clarify: It's fine to have the default debug mode (watch) point to
indra.bio (or a similar server on one of our domains).  However, we need
the docker image for the official Indra server for production.

jvwong:
> I updated this PR so that it uses the indra webservices for getting statements from agents (https://db.indra.bio/statements/from_agents?offset=0&agent0=EGFR&agent1=MDM2) and assembling english sentences from indra statements (http://api.indra.bio:8000/assemblers/english). This PR no more uses the webservice that I have implemented in the past.

@metincansiper noticing a couple with the indra.bio service. You supply `db.indra.bio/statements/from_agents` with 'agents', but these lack semantics, also you were using the 'label' provided by the author. Is it safe to assume INDRA is doing some  automated ID mapping? Is there a way to specify/filter a) ID namespace or b) organism c) mechanism? Otherwise it would come back heavily biased towards human.

metincansiper:
@maxkfranz @jvwong 
>This doesn't touch the docker-compose.yml. This is calling indra.bio domain now. @metincansiper Emek said on the call yesterday that you guys were working on creating self-contained Docker images for these services. Is that accurate?

>I was under the impression that the docker related changes included both
the python server and the official Indra server.  If it doesn't already
have the official Indra as a docker image, then we need that added.
Ideally, the Indra team would maintain an official image on Docker Hub.
- In the past I made some docker changes in this branch for the python webservice that I have implemented but as I mentioned there is no more any need for that service. Therefore, I reverted that changes. We are able to get the information from indra using the indra webservices directly.
- I remember that there was a discussion about hosting indra services for Factoid. Indra developers mentioned that the size of data would make it hard. I can ask about that. 
>@metincansiper noticing a couple with the indra.bio service. You supply db.indra.bio/statements/from_agents with 'agents', but these lack semantics, also you were using the 'label' provided by the author. Is it safe to assume INDRA is doing some automated ID mapping? Is there a way to specify/filter a) ID namespace or b) organism c) mechanism? Otherwise it would come back heavily biased towards human.
- I did not pay attention that I was using the user provided labels case. If we will make these queries after the nodes are grounded I can use the name of the grounded entity instead.
- The queries are also available as ``https://db.indra.bio/statements/from_agents?
offset=0&agent0=MAPK@FPLX&agent1=11998@HGNC``. @jvwong is it what did you mean that db names by the namespace? I do not know about whether specifying organism and/or mechanism is also possible. I can ask about it. BTW they also have an auto keyword. The following is what one of the indra developers wrote (it was for another service but probably applies this one too):

"Also, you can get some flexibility in the naming of entities by using @AUTO after each entity name. You can also specify any number of groundings, (e.g. `11998@HGNC for TP53 or MEK@FPLX for MEK grounded to Famplex)."

metincansiper:
>Is it safe to assume INDRA is doing some automated ID mapping? 

@jvwong Actually I could not well understand this. Do you mean whether there is a need for getting a grounding fist and then assigning the db and id in the query url as I was indirectly doing here (https://github.com/metincansiper/IndraInteractionServer/blob/master/indra-interaction-server/finder.py#L15). BTW the line is from the python server that will not be used.

jvwong:
>    I did not pay attention that I was using the user provided labels case. If we will make these queries after the nodes are grounded I can use the name of the grounded entity instead.
>
>    The queries are also available as https://db.indra.bio/statements/from_agents? offset=0&agent0=MAPK@FPLX&agent1=11998@HGNC. @jvwong is it what did you mean that db names by the namespace? I do not know about whether specifying organism and/or mechanism is also possible. I can ask about it. BTW they also have an auto keyword. The following is what one of the indra developers wrote (it was for another service but probably applies this one too):

Yes `db:id` information is the type that would help make the results from INDRA more relevant. As you know, we have an NCBI Gene IDs, however these IDs include many organisms. So Factoid would need to map molecular entities from NCBI Gene ID  to a database supported by INDRA (i.e HGNC).

Issue #651, opened 2020-02-28
Admin: Password protection disabled


jvwong:
For the admin, the document API route changes disabled the password protection logic. This should be reinstated ASAP.

PR #652, opened 2020-02-28
Admin: Reinstituting password protection

jvwong:
- If an API_KEY is set on server, then when a client declares an 'apiKey' url parameter, then the server will attempt to validate.
- If no API_KEY env is set on server, but the apiKey is sent by client it returns private fields.
- Otherwise, the route returns read-only versions of docs.

Refs #651

### DIFF
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -268,14 +268,16 @@ http.get('/', function( req, res, next ){
     })
     .then( cursor => cursor.toArray() )
     .then( res => {
-      try {
-        checkApiKey(apiKey);
+      if( !_.isUndefined( apiKey ) ){
+        try {
+          checkApiKey(apiKey);
+          return { res, haveSecretAccess: true };
+        } catch( err ){ throw err; }
 
-        return { res, haveSecretAccess: true };
-      } catch( err ){
+      } else {
         return { res, haveSecretAccess: false };
       }
-    } )
+    })
     .then( ({res, haveSecretAccess}) => { // map ids to full doc json
       return Promise.all(res.map(docDbJson => {
         let { id } = docDbJson;


maxkfranz:
What is the purpose of this PR?  It doesn't seem to have an impact on security.

jvwong:
It was saying that if you want the private doc fields, then you have to provide a valid apiKey parameter. Otherwise it just spits out the read-only stuff.

maxkfranz:
That was already covered by the original try-catch.  This change makes it so that an unspecified `apiKey` results in an error always.  An unspecified `apiKey` is valid when the server is run in debug mode.

What is the root problem you're working on?

jvwong:
My problem was the admin page became 'public' regardless of the apiKey sent.

maxkfranz:
What does public mean precisely?  Can a user modify a document's data without the secret or the API key?  I don't see how this route could affect that.

jvwong:
Correct, its read-only, which rendered my admin a bit goofy as now there are all sorts of items that wont work. Also, trying to access 'secret' docs with a bad apiKey is sorta ambiguous to the client, you gotta hope there's a record, look inside it for the secret field.

maxkfranz:
OK, that makes sense.  So if I understand you correctly, the admin UI looks weird with the previous `/documents' implementation but the security still works correctly.

If the `/documents` logic is coupled to the admin UI, then it may be difficult to use generally.  Other views depend on `/documents`, like the home page carousel and the future explore view.  So it would be good to keep `/documents` as general as possible.  We could potentially have a mechanism to verify the admin UI separately from the documents listing logic, while keeping `/documents` generic.

Would a separate route like `/admin-verify` or `/api-key-verify` suffice to lock down the UI side for the admin UI?  The admin UI could show an empty result or an error message if a call to that route fails.

jvwong:
> ... We could potentially have a mechanism to verify the admin UI separately from the documents listing logic, while keeping /documents generic.

In a sense, I took the rather obscure/lazy route to this in my PR: If you declare an `apiKey` param, then you're indicating that you want admin-level access to docs.

> Would a separate route like /admin-verify or /api-key-verify suffice to lock down the UI side for the admin UI? The admin UI could show an empty result or an error message if a call to that route fails.

Let me try again, I'll ask you to review first.

PR #654, opened 2020-02-28
Document route check for apiKey, used by admin  

jvwong:
Add a document API route named 'api-key-verify' that throws if not a match. Admin does a check of its apiKey param before trying to loading ('secret'-enabled) docs.


maxkfranz:
Looks good

PR #661, opened 2020-03-04
Revert "Admin: Documentation, explicit errors for bad/missing doc data  "

jvwong:
Something is broken in the bundle:js/webpack when NODE_ENV=prod.

Will try again.

Reverts PathwayCommons/factoid#660


jvwong:
Bah! Don't put 'email-regex' package in the bundle...

Issue #635, opened 2020-02-13
Document model: Define 'Read-only' version attributes of data in REST API


jvwong:
This issue concerns what type of document information should be exposed via the public web service API. We would not want items such as email correspondence or any other information of a sensitive nature provided to say a document GET route. 

This can be handled in the document model in 'read-only' mode so that all this logic is maintainable and localized.

PR #663, opened 2020-03-05
Output of 'read-only' doc data  

jvwong:
Document model outputs private JSON fields when secret is present, added in `/api/document` when apiKey is correct.
Refs #635


maxkfranz:
This looks good, but I'm curious as to why `status` is considered private.

jvwong:
I thought of status as more of an internal flag for routing (i.e. send email if published), but it's not sensitive per se.

maxkfranz:
Makes sense.  I suppose we could change it in future if an anonymous view needed access to the field.

Issue #670, opened 2020-03-11
Complexes are erroneously shown as incomplete by submit UI


maxkfranz:
![Screen Shot 2020-03-11 at 1 51 55 PM](https://user-images.githubusercontent.com/989043/76447735-89bc7280-639f-11ea-9ce7-cb936b3402a1.png)

jvwong:
I recall one idea is to auto-populate name, i.e. `rad51:mdm2`

maxkfranz:
Yes, that's a good idea for a general improvement.

The bug would happen either way.  In the example image, note that the complex is named.  For some reason, even a named complex isn't considered completed.  If I were to guess I'd say it's probably looking for grounding association data, and complexes never have that data.

maxkfranz:
For now: Don't restrict complexes -- just ignore in the list

Later: Automatic, default complex name

PR #679, opened 2020-03-17
Exclude complexes from the set of incomplete elements in the task list

maxkfranz:
Ref : Complexes are erroneously shown as incomplete by submit UI #670

### DIFF
--- a/src/client/components/tasks.js
+++ b/src/client/components/tasks.js
@@ -4,6 +4,7 @@ import h from 'react-hyperscript';
 
 import { BASE_URL } from '../../config';
 import Document from '../../model/document';
+import { ENTITY_TYPE } from '../../model/element/entity-type';
 
 const eleEvts = [ 'rename', 'complete', 'uncomplete' ];
 
@@ -93,7 +94,7 @@ class TaskView extends DataComponent {
     let { document, bus } = this.props;
     let done = document.submitted() || document.published();
     let incompleteEles = this.props.document.elements().filter(ele => {
-      return !ele.completed() && !ele.isInteraction();
+      return !ele.completed() && !ele.isInteraction() && !ele.type() === ENTITY_TYPE.COMPLEX;
     });
 
     let ntfns = document.entities().filter(ele => !ele.completed()).map(ele => {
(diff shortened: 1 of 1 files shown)


src/client/components/tasks.js
```diff
@@ -93,7 +94,7 @@ class TaskView extends DataComponent {
     let { document, bus } = this.props;
     let done = document.submitted() || document.published();
     let incompleteEles = this.props.document.elements().filter(ele => {
-      return !ele.completed() && !ele.isInteraction();
+      return !ele.completed() && !ele.isInteraction() && !ele.type() === ENTITY_TYPE.COMPLEX;
```
jvwong:
```js
return !ele.completed() && !ele.isInteraction() && !(ele.type() === ENTITY_TYPE.COMPLEX);
```

src/client/components/tasks.js
```diff
@@ -93,7 +94,7 @@ class TaskView extends DataComponent {
...
     });
 
     let ntfns = document.entities().filter(ele => !ele.completed()).map(ele => {
```
jvwong: This needs to filter out the `ele`s that are type `complex` as well.

maxkfranz: Isn't that just the same as `ENTITY_TYPE.COMPLEX`?

src/client/components/tasks.js
maxkfranz:
```suggestion
      return !ele.completed() && !ele.isInteraction() && ele.type() !== ENTITY_TYPE.COMPLEX;
```

src/client/components/tasks.js
maxkfranz: According to this condition, any complex is always excluded from the incomplete element list.

jvwong: I'm referring to the `ntfns` - shouldn't this filter out complexes like in `incompleteEles`? Right now in the UI I still see the complexes as being listed as incomplete.

jvwong: image

maxkfranz: That issue is only reproducible with the extra node in the top-left of your screenshot.  With the change, this case look OK:

Screen Shot 2020-03-20 at 11 28 05 AM

Issue #676, opened 2020-03-17
Error directly loading the demo with a clean DB


maxkfranz:
1. Delete everything in rethinkdb.
1. Start the server: `npm run watch`
1. Go to [http://localhost:3000/demo](http://localhost:3000/demo)

Note the errors: 

Client: 

```
An error occurred livening the doc Error: No response from database for ID demo
```
 
Server: 

```
info:    POST /api/document 200 122.084 ms - -
info:    Mocking email transport
info:    Email sent to: author@example.com; Caught by Ethereal: https://ethereal.email/message/XnEnyNZt0AzxGDUFXnEny7GpSgyZ.uYGAAAAAZVocBq0DIZRxMPGtL7eT8g
```

@jvwong, I suspect that recent changes re. 'demo' homepage invitations and/or emails is related.

jvwong:
- Editor POST to `/api/document/`: Server calls `res.end` before doc created.

Issue #675, opened 2020-03-13
Provide immediate access to doc upon task completion


jvwong:
## Description

**Q: What is the name of the feature?**

A: Provide immediate access to doc upon task completion

**Q: What does this feature enable the user to do?**

A: Upon signup, provide a editable document link.  

**Q: What information must the user provide to use the feature?**

A: Same as signup.

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: Errors in input (email), possibly missing article information (PubMed).

**Q: How does this feature affect each class of user (persona)?**

A: This eases the path towards accepting data input; removes friction and temporal delay associated with having to login, find, read and launch via email. 

### Mockup

Essentially a near copy of the email:

<img width="594" alt="image" src="https://user-images.githubusercontent.com/4706307/76650239-8b6d6e00-6538-11ea-91b0-56e56f23476c.png">

### Details

- Look into reusing the TaskView component
- Part of the signup assistant series.

maxkfranz:
Let's test this once on unstable before closing

jvwong:
Not sure if 'ridiculously long titles' is a case worth handling, but maybe I could combine it with the [suggested  text](https://github.com/PathwayCommons/factoid/pull/687#issuecomment-601363504): 

<img width="1476" alt="Screenshot at Mar 20 12-50-34" src="https://user-images.githubusercontent.com/4706307/77186543-6e411e00-6aa9-11ea-8e99-f809836c2339.png">

maxkfranz:
Let’s just truncate for now. The text stuff isn’t important

PR #684, opened 2020-03-17
Create doc before ending server response

jvwong:
This was noticed on the demo (Editor) page, but updates the server so that POST request to `api/document` now end server response  after the doc is created. 

Refs #676.

### DIFF
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -542,17 +542,19 @@ http.post('/', function( req, res, next ){
     if( id === DEMO_ID ) await deleteTableRows( API_KEY, secret );
     return await createDoc({ docDb, eleDb, id, secret, provided });
   };
+  const sendJSONResponse = doc => tryPromise( () => doc.json() )
+  .then( json => res.json( json ) );
 
   checkRequestContext( provided )
     .then( () => createSecret({ secret }) )
     .then( loadTables )
     .then( handleDocCreation )
-    .then( doc => { res.end(); return doc; } )
     .then( setRequestStatus )
     .then( fillDoc )
     .then( setApprovedStatus )
     .then( tryVerify )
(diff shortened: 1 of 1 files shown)


src/server/routes/api/document/index.js
```diff
@@ -544,10 +544,10 @@ http.post('/', function( req, res, next ){
   };
 
   checkRequestContext( provided )
-    .then( () => res.end() )
     .then( () => createSecret({ secret }) )
     .then( loadTables )
     .then( handleDocCreation )
+    .then( doc => { res.end(); return doc; } )
```
maxkfranz: Should something like this be at the very end (just before `next`)?  Maybe we should use [`sendStatus()`](http://expressjs.com/en/5x/api.html#res.sendStatus) instead

maxkfranz: Or maybe the response should be the document JSON

jvwong: I'll return JSON as this is part of my upcoming PR for #675 where the doc link is provided in the success for CTA.

jvwong: Updated to return JSON at end, but you might notice it's a bit laggy (2-3s to end response)

maxkfranz: Is it the email part creating most of the lag?

jvwong: Yes, that's my best guess just having observed the server console logging

maxkfranz: Maybe that part could be after the sendStatus().  The main thing is that the document JSON is stable before sending off the http response.

jvwong: Yes I think that's the solution.

maxkfranz: Sounds good

PR #689, opened 2020-03-23
Limit display of article titles in request form/CTA

jvwong:
Addendum to #675.

Examples: 

- Real article with long title

Screenshot at Mar 23 14-33-08

- Bogus long text

Screenshot at Mar 23 14-33-35


maxkfranz:
It would be good if the truncation would be made at the closest word so that words don't get cut off.  But we don't need that now.

Merging

Issue #686, opened 2020-03-18
Tweet to @biofactoid when a document status is set to publish


jvwong:
## Description

**Q: What is the name of the feature?**

A: Tweet it  

**Q: What does this feature enable the user to do?**

A: View and share their submitted pathway via Twitter

**Q: What information must the user provide to use the feature?**

A: Submit their document.

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: The document should achieve a status of 'published'. 

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: This feature enables users to view and share their work on a well-used platform. It gives them an immediate sense that their work can be accessed. It provides a means to increase the visibility of their work. Ideally this is linked and provides an avenue to  a value-added 'explore view'.

### Mockup & Details

> describe how the automatic tweet is shown to the user in thank-you emails.  This could act as a CTA to retweet or independently share via mobile social apps.

#### A note to self about the 'lifecycle' of a document:

![Notebook](https://user-images.githubusercontent.com/4706307/77357430-8a012a00-6d1e-11ea-99b5-c7401940b655.png)

jvwong:
> describe how the automatic tweet is shown to the user in thank-you emails. This could act as a CTA to retweet or independently share via mobile social apps.

@maxkfranz not sure if we're on the same page here with the Tweet

maxkfranz:
Let's discuss tomorrw

jvwong:
- Email: this is a separate concern from this issue, which deals with this codebase
- Admin: Link to Tweet would be nice

maxkfranz:
@jvwong Does the system tweet at all right now?  If not, this should probably be included in the current iteration (0.8) rather than the initial explorer iteration (0.9).

jvwong:
> @jvwong Does the system tweet at all right now? If not, this should probably be included in the current iteration (0.8) rather than the initial explorer iteration (0.9).

No tweeting. Done. I'm just testing my branch..

jvwong:
I updated the 'follow-up' email template on Mailjet to look like this:

<img width="471" alt="image" src="https://user-images.githubusercontent.com/4706307/77576592-d463e180-6eab-11ea-9590-5833ff1ac22b.png">


- Images not displayed in client:

<img width="477" alt="image" src="https://user-images.githubusercontent.com/4706307/77576641-e8a7de80-6eab-11ea-8ebb-c37b0fcb80f2.png">

PR #694, opened 2020-03-25
Auto-Tweet on doc status 'published'

jvwong:
- Reuse doc endpoint logic for posting to Twitter on doc status set to 'published'
- Can only Tweet maximum of once
- Trying text to contain the templated interaction descriptions
- Updated to allow publish when incomplete complexes (#679)

Refs #686 

Example pathway:

image

Resulting tweet text (minus card, since localhost).

image

### DIFF
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -429,14 +429,12 @@ http.get('/(:id).png', function( req, res, next ){
 });
 
 // tweet a document as a card with a caption (text)
-http.post('/:id/tweet', function( req, res, next ){
-  const id = req.params.id;
-  const { text, secret } = _.assign({ text: '', secret: 'read-only-no-secret-specified' }, req.body);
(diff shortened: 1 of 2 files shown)


maxkfranz:
Great.  That looks like the simplest mechanism for sending out the tweet -- the listen for 'publish' mechanism has a lot of edge cases.

Because of the issue you had with incomplete complexes, it probably makes sense to just mark all complexes as `complete: true` in future.  We shouldn't duplicate the incomplete complex logic in multiple places for the long term.

It would be nice to have a better caption, so this could potentially underscore the importance of a user-editable caption text box in the editor for 0.9

jvwong:
> It would be nice to have a better caption, so this could potentially underscore the importance of a user-editable caption text box in the editor for 0.9

Now we can claim we're using 'Natural Language Generation' (NLG) and a 'Model-to-Text' pipeline :)

Issue #681, opened 2020-03-17
Related papers system


maxkfranz:
## Description

**Q: What is the name of the feature?**

A: Related papers

**Q: What does this feature enable the user to do?**

A: The goal is to develop a system to return a ranked list of articles, given one or more interactions from Biofactoid (participants, relationship).

**Q: What information must the user provide to use the feature?**

A: There are at least two cases: 1. Those related to the entire pathway 2. Those related to a particular interaction 

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: The key issue is providing a source of articles that covers the organisms in Factoid, achieving a reasonable degree of specificity (reduce 'noise' / irrelevant material) via data source selection filtering and ranking.   

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: Provides supporting articles for a given pathway or interaction of interest. Brings in related information that may not have been aware of.
- Author: Added value to their pathway data contribution 'for free'; delivers on the promise of integrating their data with a larger body of knowledge
- Curator: See biologist.

### Mockup

see #683 

### Details

- Query: This is the input, an interaction from Biofactoid 
- Source: These are the sources of articles. We propose to draw from PubMed, INDRA (which draws from other databases + REACH) and the Biofactoid database itself
- Output: This should be a ranked list of ‘documents’. Each doc referenced a unique article ID, one or more interactions with ‘entities’, ‘relation’, ‘source’, etc.  

![image](https://user-images.githubusercontent.com/4706307/79898884-f896dd80-83d9-11ea-85b1-1eeb0c0f5720.png)

There is a more in-depth scoping doc with basic requirements [JW].

cannin:
@maxkfranz @jvwong Potentially useful, but I don't think we have discussed this much. 

* What is the source of the related papers? Only existing factoid papers? 
* Could it be offloaded through a call to the PC API or a just opening a new tab with a PC search?

maxkfranz:
It would be ideal from a technical standpoint if the related papers could be generated only from factoid data.  However, that provides low utility for early users:  There isn’t much data yet.  We need to focus on how to provide high utility to early adopters.  We should show related papers from a large set of the latest articles from Pubmed, or a similar large-scale, multi-organism source.

[...]

PR #701, opened 2020-04-06
Home page style tweaks (expand on hover, etc.)

maxkfranz:
This PR includes a number of style fixes and improvements for the home page, chief among them is the revision to the hover effect for the Netflix-like article scroller:

hover

This avoids the figure in each card from overlapping the expanded text.  This change will be important to allow for reusing this scroller component for the related papers UI (#681).  The related papers UI will need this extra space in order to accommodate the data shown for each paper.

There are a number of other minor stylistic improvements that affect the home page that are included in this PR.  I've tried to separate out each item into a corresponding commit to facilitate review.  The changes include:

- Make the title & caption more legible for some small screens
- Revise the hover behaviour for the scroller
  - The scroller card expands on hover to give more room for the title and author list without overlapping the figure.
  - The figure is no longer blurred out on hover, as there is no overlap.
  - The scroller on the home page is shifted up to accommodate the extra space needed for card expansion.
- Improve the reliability of hovering w.r.t. scrolling
  - Hovering is faster to activate on mobile.
  - The hover state is removed when scrolling.
- The visibility of the pager icons is updated on window resize
- Links on the home page go out to new tabs
- Improve max screen size threshold for hiding the institution logos in the footer

### DIFF
--- a/src/client/components/home.js
+++ b/src/client/components/home.js
@@ -186,16 +186,44 @@ class Scroller extends Component {
     this.state = {
       pagerLeftAvailable: false,
       pagerRightAvailable: false,
+      isScrolling: false,
       docs: []
     };
 
-    this.onScrollExplore = _.debounce(() => {
+    this.updatePagerAvailabilityDebounced = _.debounce(() => {
       this.updatePagerAvailability();
     }, 40);
+
+    this.setScrollState = () => {
+      this.setState({ isScrolling: true });
+
+      this.clearScrollStateDebounced();
+    };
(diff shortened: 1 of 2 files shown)


jvwong:
This one works for me. A couple of notes for a future iteration...
1. Picky detail: Scroller white background shrinks faster than the text on mouse off, so the text seems to momentarily float outside
2. Small screen: Move to a vertical tiling rather than horizontal, as its a little crowded as is.

Issue #693, opened 2020-03-25
Complexes: Provide a default name


jvwong:
## Description

**Q: What is the name of the feature?**

A: Default complex name

**Q: What does this feature enable the user to do?**

A: Understand, in words, the nature of the interaction participants

**Q: What information must the user provide to use the feature?**

A: Create a complex

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: n/a

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: 
- Editor: 
- Computational biologist: 
- Curator: 

### Details

Related to #670.

Current behaviour:

<img width="629" alt="Screenshot at Mar 25 07-10-18" src="https://user-images.githubusercontent.com/4706307/77531530-77473c00-6e69-11ea-9fe7-603ef183e4ba.png">

PR #704, opened 2020-04-09
Default name of complex in interactions

jvwong:
For interaction toString, use the (colon-separated) component names for unnamed complexes.
Refs #693 

image

### DIFF
--- a/src/model/element/complex.js
+++ b/src/model/element/complex.js
@@ -63,7 +63,7 @@ class Complex extends Entity {
     } else {
       let name;
       if( this.named() ){
-        name = this.syncher.get('name');
+        name = super.name();
       } else {
         const participantNames = this.participants().map( p => p.name() ).join(':');
         name = `Complex (${participantNames})`;
--- a/src/model/element/complex.js
+++ b/src/model/element/complex.js
@@ -57,6 +57,21 @@ class Complex extends Entity {
     return this.elementSet.create();
   }
 
+  name( newName ){
+    if( newName !== undefined ){
+      return this.rename( newName );
+    } else {
+      let name;
+      if( this.named() ){
+        name = this.syncher.get('name');
+      } else {
+        const participantNames = this.participants().map( p => p.name() ).join(':');
+        name = `Complex (${participantNames})`;
+      }
+      return name;
+    }
+  }
+
   synch( enable ){
     return tryPromise( () => {
       return super.synch( enable );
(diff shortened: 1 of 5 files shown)


maxkfranz:
Could this be addressed in a simpler, more universal way by altering the `Complex` class?  For example, you could override the `complex.name()` to return the default pattern if `complex.named() === false`.

maxkfranz:
Link to Complex class: https://github.com/PathwayCommons/factoid/blob/unstable/src/model/element/complex.js

maxkfranz:
Link to ancestor `Element` class: https://github.com/PathwayCommons/factoid/blob/unstable/src/model/element/element.js#L113-L125

maxkfranz:
Altering the `Complex` class may resolve all of #693

jvwong:
ok

jvwong:
Issue: The cytoscape.js label for the complex only seems to agree with the default name in entity-info tooltip initially, upon creation. Subsequent changes to participants doesn't get tracked in the label?

src/model/element/complex.js
```diff
@@ -57,6 +57,21 @@ class Complex extends Entity {
     return this.elementSet.create();
   }
 
+  name( newName ){
+    if( newName !== undefined ){
+      return this.rename( newName );
+    } else {
+      let name;
+      if( this.named() ){
+        name = this.syncher.get('name');
```
maxkfranz, line 66:
```suggestion
        name = super.name();
```

jvwong:
Untitled

src/model/element/complex.js
maxkfranz: You should delegate the low-level getter to the superclass so that whatever logic within the superclass in maintained -- e.g. there could be general name post-processing rules within `Entity` that apply to all node types.

maxkfranz:
> Issue: The cytoscape.js label for the complex only seems to agree with the default name in entity-info tooltip initially, upon creation. Subsequent changes to participants doesn't get tracked in the label?

There's probably only a listener for `rename` that updates the cyjs `data` for the renamed node directly.  A complex doesn't get a `rename` event when one of its children is renamed.  I suppose the existing listener could also check to update the parent node `data`.

maxkfranz:
I think this is the listener I mentioned: 

https://github.com/PathwayCommons/factoid/blob/unstable/src/client/components/editor/cy/doc.js#L211-L215

jvwong:
OK will try, I just went down a 3 hour hole looking at this...

maxkfranz:
If you run into trouble, you can message me on hangouts if you want

Issue #698, opened 2020-03-26
Describe the Document web service API


jvwong:
## Description

**Q: What is the name of the feature?**

A: Web service API description for 'document'

**Q: What does this feature enable the user to do?**

A: Human-readable information regarding interaction with Express routes

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: Several routes are key protected and never intended for public consumption; Some data returned can depend on this (GET)

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: Not intended user 
- Editor: Not intended user
- Computational biologist:  Provides information about how they might access the data in various formats for use in their research or integration with other data sources. 
- Curator: Useful as above.

### Details

Use Swagger in 'swagger-ui-express' for docs.  

Related to #685

Issue #685, opened 2020-03-17
Update Document API for data formats: TXT, SBGN, BioPAX


jvwong:
The various web service API endpoints for data formats are either broken (BIOPAX, SBGN) or out of sync with model updates (TXT). The goal is to update and test so they are functioning and providing correct data in return.

jvwong:
@metincansiper - 1. were you the author? 2. do you have the ability to update service at [https://biopax.baderlab.org/convert/](https://biopax.baderlab.org/convert/)

metincansiper:
Interestingly ``EntityReference`` objects cannot be added to biopax model (I mean even not created) as I described here (https://github.com/BioPAX/Paxtools/issues/47).

maxkfranz:
If this is a problem only for incomplete entities (e.g. ungrounded nodes), then one alternative would be to exclude those nodes in the biopax export.  There should generally not be any ungrounded nodes in current interface, so it's a fair assumption.

maxkfranz:
Note that a complex does not have a grounding, so something like `ent.complete() === false` would probably be the best way to identify problematic nodes.

metincansiper:
@maxkfranz looks like the problem is not related to whether the entity is complete (I need to check if that is another problem as well but this one is different). Following up the response of Igor in the issue that I opened in paxtools, I see that ``EntityReference`` is an abstract class. I was not aware of that and so was trying to assign the entity to that if its type is neither protein nor chemical. Now I will probably handle that case as an exception. 

However, the real thing to consider is why we had some entities that is neither of protein nor chemical. I realized that there are new types like ggp, dna and rna. As far as I can understand from some client side code we must interpret them as proteins. Is that correct? By assuming so I will input them as proteins to the converter which will probably solve this problem. 

The other problem is about the complexes. Do we want them to be included in the converter? If we do not want it I can just ignore the interactions that has complex in one or both side. Otherwise, I can make the needed implementation to consider complexes in both the converter and the factoid web app sides.

metincansiper:
>Note that a complex does not have a grounding, so something like ent.complete() === false would probably be the best way to identify problematic nodes.

I just realized this comment. So, I assume that we must consider the complexes in the converter.

jvwong:
[...]

PR #706, opened 2020-04-14
Swagger docs for (Document) web services

jvwong:
- Swagger + UI for the web services (primarily `/api/document/`)
- Access at `/api/api-docs/` 
- GET biopax, sbgn in progress #685

Definitely room to improve later on:
  - Reconcile various PATCH routes
  - More consistent handling of paths and params (i.e. secret)

Refs #698

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,45 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@apidevtools/json-schema-ref-parser": {
+      "version": "8.0.0",
+      "resolved": "https://registry.npmjs.org/@apidevtools/json-schema-ref-parser/-/json-schema-ref-parser-8.0.0.tgz",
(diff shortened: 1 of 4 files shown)


maxkfranz:
Looks like there's a problem on prod builds, e.g. http://unstable.factoid.baderlab.org/api/api-docs

jvwong:
It requires the trailing slash: http://unstable.factoid.baderlab.org/api/api-docs/

maxkfranz:
Could we put some simple forwarding routes in place to make it easier if someone types in the URL from the paper?

It would be great if these were supported:

- /api/api-docs : The trailing slash shouldn't make a difference when you type in a URL.  That's the convention, anyway (e.g. `http://biofactoid.org/demo/` is the same as `http://biofactoid.org/demo`).
- /api : Nothing else is going to point here, so we may as well have a shorter address.
- /api/ : The trailing slash should work here too.

jvwong:
what about putting the docs at `/api(/)`?

maxkfranz:
That would be the natural place to put the documentation, if possible.  I thought that the longer `/api/api-docs/` address was used because of some technical conflict with `/api`.

jvwong:
OK I figured it out - I reversed the order of the swagger setup...

PR #711, opened 2020-04-20
Workaround for mobile page size

maxkfranz:
- Apply a fixed pixel size to the editor page on load.
- Scroll to the top of the document upon loading the editor page.
- The fix is reapplied each time the page is resized.

Ref : #672

Issue #683, opened 2020-03-17
Paper metadata display: explorer view


maxkfranz:
## Description

**Q: What is the name of the feature?**

A: Paper metadata display panel

**Q: What does this feature enable the user to do?**

A: This panel allows the user to explore information related to the paper.  It provides global information about the document, while the network view provides local information about the document.

**Q: What information must the user provide to use the feature?**

A: The document must be previously associated, by the author, with a PubMed ID in order for the metadata to be generated and displayed.  The reader does not need to provide any information.

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: 

- The PubMed metadata must have been previously loaded into the system.
- Any analysis needed to generate information shown in the panel must be immediately available upon loading the view.  Expensive data must be precomputed.

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: For producer/contributors, this delivers on the promise that the pathway data will be linked to their article. This provides concrete value as attribution (author list) and avenue that can contribute to awareness of their article. Similarly for consumers, it makes clear that the network data they see is 'from' the article and by the author. The article metadata also helps give readers critical context  that is otherwise very difficult to capture in structured languages.    
- Editor: Seeing the journal and article prominently featured with hyperlinks is an excellent way to both increase awareness among potential readers/subscribers but also drive traffic to their articles. This is also an ideal place to realize journal branding. 
- Curator: These are consumers, see above
- Computational biologist: This data is already included in the doc JSON

### Mockup

Here is a mockup of the explore view panel, which may contain expanded paper metadata.

The view on large screens (desktop, tablets):

![factoid-explore-panel page 1](https://user-images.githubusercontent.com/989043/77087997-3323e980-69da-11ea-9725-9066b2730083.jpg)

The view on small screens (mobile):

![factoid-explore-panel page 2](https://user-images.githubusercontent.com/989043/77088108-5484d580-69da-11ea-8cb3-d4781661f68f.jpg)

### Details

- The panel replaces the metadata (e.g. title) summary that is normally shown in the editor.
- The panel can hold any paper-level information about the document, such as related papers (#681).
- The panel is of relatively consistent width, regardless of the device's screen size.  This allows for well-fitted, consistent content on both desktop and mobile.
- Paper metadata included in the panel
  - Title
  - Full author list
  - Journal info
  - Abstract
  - Related papers
  - ... More TODO

jvwong:
[...]

Issue #672, opened 2020-03-11
The page size is incorrectly set sometimes in iOS Safari


maxkfranz:
Sometimes you get the network misaligned, because the page size is reported incorrectly:

<image height="200" src="https://user-images.githubusercontent.com/989043/76448331-8f668800-63a0-11ea-9bc0-e61679dee189.png"/>

Sometimes the proper page size is reported, and so everything fits to the screen:

<image height="200" src="https://user-images.githubusercontent.com/989043/76448428-b0c77400-63a0-11ea-9e63-e02ba065fb57.png"/>

Anything that is positioned relative to the bottom of the screen will be randomly cut off and inaccessible.  This bug conflicts with important short-term features, like related paper listing in the exploration view.

I have a workaround in mind.

maxkfranz:
Ref : https://github.com/GeneMANIA/genemania/commit/64ccf9ffae994b8b678fa1a3ca92725d7ec0180f

maxkfranz:
Affects UI such as https://github.com/PathwayCommons/factoid/issues/671

PR #715, opened 2020-04-22
Add an info panel for the explore view

maxkfranz:
Design details:

- The panel is a sidebar on devices that have enough space.
- On small-screen devices, the panel takes up a small area at the bottom of the screen by default.
- The panel can be scrolled.
- On small screens, scrolling the panel moves the panel up.  This puts the panel on top of the graph.  This mechanism allows for the info panel to take up the entire screen on phones, giving more space for content.
- The info panel currently contains a simple version of the paper metadata UI.
- This panel is where a related papers UI would go.

Preview for desktop:

Screen Shot 2020-04-22 at 1 48 45 PM

Preview for mobile:

Technical details:

- The `GOOGLE_SCHOLAR_BASE_URL` has been added to facilitate links in the explore view.
- Subcomponents within `Editor` have been put into separate files, including `Help`, `Submit`, and `Title`.
- When the screen of a phone or tablet is rotated, the graph fits to the screen.  This prevents the user from losing the graph in the explore view.

Ref. #683 #681

Depends on #711 #672


src/client/components/editor/info-panel.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  render(){
+    const { document } = this.props;
+
+    if( document.editable() ){ return null; }
+
+    const citation = document.citation();
+    const { authors, pmid, title = 'Untitled article', reference, doi, abstract } = citation;
+
```
jvwong, line 17: Since 'anything goes' with regards to user input, we could have  `pmid`, `doi` be `null` and `authors` empty, so these links wouldn't be informative.

maxkfranz: The content is just filler for demonstration.  We'll have to consider the content of the panel in more detail in future PRs.  This PR is just structural.

PR #718, opened 2020-04-29
Related papers and PubMed: More flexibility in number of records retrieved 

jvwong:
Updated the fetchPubmed service to permit customization (opts) and permit a large number of records to be requested and returned.

Example usage:
```js
fetchPubmed({
	uids: [
		"32344470",
		"32342131",
		"32341393",
		"32341020",
		"32340282",
		"32338278",
		"32337406",
		"32337286",
		"32336747",
		"32336254",
		"32336023",
		"32335977"
	],
	opts: {
		retmax: 100 // return maximum #  records
	}
})
```

Refs https://github.com/PathwayCommons/factoid/issues/681#issuecomment-621276912

PR #721, opened 2020-04-29
Related papers

metincansiper:
Refs #681

I cherry-picked the commit from Jeff's PR (https://github.com/PathwayCommons/factoid/pull/718) into this branch. That commit is merged into unstable as well. Therefore, I would expect that the changes in that commit are not visible in the diff view of this PR. But they are visible in a way that I do not understand.


src/model/element/entity.js
```diff
@@ -177,6 +177,25 @@ class Entity extends Element {
...
+  toIndraTemplate(){
+    let name = this.name();
+
+    // if the name is not set it means that there is no info for the entity
+    // so that entity must not be included
+    if ( !name ) {
+      return null;
+    }
+
+    let xref = _.pick(this.association(), ['id', 'namespace']);
```
jvwong, line 190: - I believe your intention for `xref` is to retrieve a `dbXref` list if the source is NCBI Gene. This currently always returns `xref=null`.
- the INDRA specific stuff (valid namespaces) could be moved to your api/doc/indra.js module

src/model/element/entity.js
```diff
@@ -177,6 +177,25 @@ class Entity extends Element {
...
+    let validNamespaces = ['hgnc', 'chebi'];
```
jvwong, line 191: - Case-sensitive

src/model/element/entity.js
```diff
@@ -177,6 +177,25 @@ class Entity extends Element {
...
+
+    if ( !_.includes( validNamespaces, xref.namespace ) ) {
+      xref = null;
+    }
+
+    return { name, xref };
```
jvwong, line 197: I suspect you want something output like this:

```
[
    {
      "name": "2'-3'-cGAMP",
      "xref": null
    },
    {
      "name": "SLC19A1",
      "xref": {
        "db": "HGNC",
        "id": "HGNC:10937"
      }
    }
  ]
```

src/server/routes/api/document/indra.js
```diff
@@ -19,10 +21,23 @@ const BASE_MODIFICATION_TYPES = ['Sumoylation', 'Desumoylation', 'Hydroxylation'
...
-  const getForPair = pair => {
-    let agent0 = pair[0].toUpperCase();
-    let agent1 = pair[1].toUpperCase();
+const getDocuments = templates => {
+  const getForIntn = intnTemplate => {
+    const getAgent = t => {
+      let agent;
+
+      if ( t.xref ) {
+        agent = t.xref.id + '@' + t.xref.namespace;
```
jvwong, line 30: Beware: HGNC database records have IDs that look like `HGNC:12345` but INDRA wants `12345`

package.json
```diff
@@ -60,6 +60,7 @@
     "leven": "^2.1.0",
     "lodash": "^4.17.15",
     "lru-cache": "^4.1.1",
+    "moment": "^2.24.0",
```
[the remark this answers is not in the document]
metincansiper: I was not aware of the other date library is being used. I use the moment libaray just to convert months in different formats to the corresponding number. For example both of ``moment().month("Nov").format("M")`` and ``moment().month("November").format("M")`` will yield to "11".

I could not find a way of doing the same in ``date-fns``. Probably putting a second libarary responsible of dates would not sound good. In that case I can implement that conversion by myself. In that case I need to know the possible moth formats. So far I have seen only things like "Nov" or "11". Do you know if having the full month name like "November" or any other format is possible (if there is any other format I am not sure even how I was suing ``moment`` library would be covering that though)?

src/model/element/entity.js
metincansiper: Yes but the namespaces I have inspected so far was in full lowercases. I am able to inspect 'chebi' and 'ncbi' so far.

metincansiper: Is HGNC in full uppercases in factoid?

package.json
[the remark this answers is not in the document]
maxkfranz: Would you please post an example date string from the data?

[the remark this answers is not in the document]
jvwong: https://dtd.nlm.nih.gov/ncbi/pubmed/doc/out/180101/el-PubDate.html 

> Element: Month
> Contains the value of the month in a date. It will be recorded as either the numeric value or the first three letters of the month's name.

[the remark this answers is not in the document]
jvwong: > Element: PubDate
> ( ( Year, ( ( Month, Day? ) | Season )? ) | MedlineDate )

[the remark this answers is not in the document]
maxkfranz: If you have individual variables for things like `month` in `MM` format, you can just use `parse()` with the appropriate format.

https://date-fns.org/v2.12.0/docs/parse

The format strings are similar to what you may have already seen with the Unix `date` CLI: http://man7.org/linux/man-pages/man1/date.1.html

src/model/element/entity.js
jvwong: You're right, the association/grounding-service namespace is lower case ('chebi', 'ncbi'). My confusion is why 'hgnc' is within `validNamespaces`, since the grounding-service doesn't support that datasource?

package.json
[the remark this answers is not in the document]
metincansiper: Okay it says that either numeric or the first three letters. Then I can make it work as the following: ``parse('Nov', 'MMM', new Date()).getMonth() + 1`` which yields to 11 as expected.

src/model/element/entity.js
metincansiper: Looks like I misunderstood something and was thinking that ``hgnc`` is supported in a way. 

Then what should be in the valid namespaces now? I remember in the past we talked about hgnc and chebi.

src/model/element/entity.js
metincansiper: I guess the only difference is you have ``db`` instead of ``namespace``. Am I right? Actually it will only be internally used but I can update it as ``db``both here and where it is used if that is the case.

src/server/routes/api/document/indra.js
metincansiper: Would using what is remaining after ":" save any possible cases?

src/model/element/entity.js
metincansiper: If that is the case is there any need to check for valid namespaces? May I just get the ``dbXref`` even though it is always ``null``currently?

jvwong: I meant that your existing code currently always returns `null` for NCBI Gene records.

src/model/element/entity.js
jvwong: Can just put array in xref and let your indra.js module pick off what it needs.

src/server/routes/api/document/indra.js
jvwong: Can you clarify what you mean by 'save'?

metincansiper: I was meaning: would trimming the text after ":" enough for any case?

jvwong: Yes that's fine for HGNC at least.

src/model/element/entity.js
metincansiper: @jvwong  Okay initially I misunderstood what you said. Actually I did it on purpose because I was thinking that ``hgnc`` is also one of the possible namespaces. Also, since we only care the xrefs for hgnc and chebi I was thinking that for the entries whose namespace is ncbi the xref always should be null here so that the indra query would be formed accordingly.

However, now I suspect that for hgnci entries the namespace is still ncbi but they have some dbxrefs field. And we will consider them if the ``dbXrefs`` is set and the available db is ``hgnc``. Is that correct? If so will we consider the first record from the dbXrefs while forming the query? Is it the same for chebi as well?

jvwong: Yes, we are in agreement. My suggestion below was just to put **everything** (namespace, id; dbXrefs that exist) in an array and let the indra.js module pick out what it can wants. In this way, your function becomes more generic, like  `toSearchTemplates` 

e.g.

```js
[
    {
      "name": "SLC19A1",
      "xrefs": [
        {
          "db": "ncbi",
           "id": "6573"
        },
        ...,
        {
          "db": "HGNC",
           "id": "HGNC:10937"
        } 
      ]
    }
]
```

src/model/element/entity.js
metincansiper: I>You're right, the association/grounding-service namespace is lower case ('chebi', 'ncbi'). My confusion is why 'hgnc' is within validNamespaces, since the grounding-service doesn't support that datasource?

I guess my main confusion is resolved in another part of conversation. As I mentioned I was thinkg that hgnc would be one of the namespaces but now I understand that it is not.

> Can just put array in xref and let your indra.js module pick off what it needs.

Yes I can do so if we rename the ``toIndraTemplate()`` functions as ``toTemplate()`` and make them more generic as you suggested.

In that case I consider including id, namespace, name, dbXrefs, parentId in the return value of ``entry.toTemplate()`` to make it more generic. Let me know if you think some other features should be included as well.

metincansiper:
if I am not missing anything everything that we discussed here must be addressed after my last commit.

src/model/element/entity.js
jvwong: Maybe `toSearchTemplate` or something a little more specific. That's it I think.

src/model/element/entity.js
jvwong: I used `db` and `id` yes.

src/server/routes/api/document/indra.js
```diff
@@ -19,10 +21,41 @@ const BASE_MODIFICATION_TYPES = ['Sumoylation', 'Desumoylation', 'Hydroxylation'
...
+      if ( arr.length == 1 ) {
+        return arr[0];
+      }
+
+      return arr[1];
+    };
+
+    const getXref = t => {
+      let validDbs = ['hgnc', 'chebi'];
+      let xref = _.find( t.dbXrefs, x => _.includes( validDbs, x.db.toLowerCase() ) );
```
jvwong, line 53: In the case of chemical, `chebi` information will be in the entity template `id` and `namespace` fields. No `dbXrefs` will exist for ChEBI.

src/server/routes/api/document/indra.js
```diff
@@ -36,18 +69,45 @@ const getDocuments = pairs => {
...
-    let arr = pmids.map( pmid => {
-      let elements = o[ pmid ];
-      // prevent duplication of the same interactions in a document
-      elements = _.uniqWith( elements, _.isEqual );
-      elements.forEach( e => delete e.pmid );
-      return { elements, pmid };
-    } );
-
-    return arr;
+    return tryPromise( () => fetchPubmed({ uids: pmids }) )
```
jvwong, line 72: This should check to see if `pmids` is empty.
PubMed EUTILS and the wrapper `fetchPubmed` service will return HTTP 400 status when you call it without including a valid pmid. So when INDRA returns 0 statements, your `getDocuments` ends up throwing the HTTP 400.

maxkfranz:
I think that the next step would be to improve the clarity of these services and simplify things.  How about something like this:

- GET `/api/document/:documentId/related-papers`
  - If the `interactionId` parameter exists (e.g. `related-papers?interactionId=uuid`), then the related papers only need to consider one interaction.
  - If the `interactionId` parameters does not exist, then the related papers considers all interactions within the document.

src/server/routes/api/document/indra.js
```diff
@@ -19,10 +21,41 @@ const BASE_MODIFICATION_TYPES = ['Sumoylation', 'Desumoylation', 'Hydroxylation'
...
+      let agent;
+      let xref = getXref( t );
+      if ( xref ) {
+        agent = sanitizeId(xref.id) + '@' + xref.db;
+      }
+      else {
+        agent = t.name + '@TEXT';
+      }
+
+      return agent.toUpperCase();
```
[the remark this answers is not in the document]
metincansiper: Looks like id does. This query returns some results ``https://db.indra.bio/latest/statements/from_agents?offset=0&agent0=TP53@TEXT&agent1=MDM2@TEXT`` but this one returns empty``https://db.indra.bio/latest/statements/from_agents?offset=0&agent0=tp53@TEXT&agent1=MDM2@TEXT`` where only difference is that ``tp53`` is witl lowercases. I realized that interestingly it returns empty when the ``text`` is with lowercases as well.

metincansiper:
>I think that the next step would be to improve the clarity of these services and simplify things. How about something like this:

GET /api/document/:documentId/related-papers
If the interactionId parameter exists (e.g. related-papers?interactionId=uuid), then the related papers only need to consider one interaction.
If the interactionId parameters does not exist, then the related papers considers all interactions within the document.

@maxkfranz Sounds good but is there a specific reason to put ``:documentId`` before ``related-papers``. Looks like the other way around would be more consistent with the existing other end points.

maxkfranz:
> @maxkfranz Sounds good but is there a specific reason to put :documentId before related-papers. Looks like the other way around would be more consistent with the other existing end points.

I think the ID-first order is more logical, but maybe there's a conflict with that approach and the routes that use `:secret` afterwards.  Let's use whatever's consistent with what we have already.  It's not worth changing everything around right now.  The main thing is that I think it would be simpler to have a single route and use a name like `related-papers` so it's clear what it does.

metincansiper:
@maxkfranz @jvwong I made the code updates based on your new reviews.

src/server/routes/api/document/indra.js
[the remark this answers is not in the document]
jvwong: Can we remove the toUpperCase?

jvwong:
> @maxkfranz @jvwong I made the code updates based on your new reviews.

If we just remove that toUppercase stuff its fine with me!

metincansiper:
>If we just remove that toUppercase stuff its fine with me!

But as i mentioned if I do so then it would always ``https://db.indra.bio/latest/statements/from_agents?offset=0&agent0=tp53@TEXT&agent1=MDM2@TEXT`` will return no result while the version of same thing with full upper cases returns some result. I am not sure but it may always be needed to have them full uppercases to have results. Even the text was supposed to be with full uppercases to get some results back in my trials so far. 

Do you still want to remove it? Maybe I can ask to the indra developers why it is happening like that.

jvwong:
You seem to be testing the system using HGNC gene symbols as values for `agent`, which may or may not have anything to do with how those genes are described in actual text by actual researchers.

Examples:
 - Most people would mention `p53` rather than the HGNC symbol `tp53` or `TP53`   
 - Most people would mention `cyclin-D1` rather than the HGNC symbol `CCND1`

In either case, if we don't have a dbXref that is supported, and the @TEXT suffix will map to exact text, then I don't think we should be interpreting the authors input at all.

metincansiper:
> You seem to be testing the system using HGNC gene symbols as values for agent, which may or may not have anything to do with how those genes are described in actual text by actual researchers.

Okay, I asked the question. They mentioned that the ``TEXT`` search is case sensitive. Based on what I understand from your comment we must respect what user entered and make a search like ``tp53@TEXT`` or ``TP53@TEXT`` without making any conversion at the gene name (TEXT must always be with uppercases though).

However, I can mention the other ways they suggested anyways (look like we wont use any of them we will use text). 

- Use ``gene@HGNC-SYMBOL`` if the official gene symbol is known.
- Use GILDA service to get a grouning and use it and make the query as ``gent0=ID@NS``.

jvwong:
> However, I can mention the other ways they suggested anyways (look like we wont use any of them we will use text).
> 
>     * Use `gene@HGNC-SYMBOL` if the official gene symbol is known.
> 
>     * Use GILDA service to get a grouning and use it and make the query as `gent0=ID@NS`.

Good to know!

metincansiper:
@jvwong I removed coverting agent texts to uppercases.

maxkfranz:
OK, let’s just merge this in for now.  The ranking of results with John’s service can be a new PR

Issue #717, opened 2020-04-27
Offer contributed data as bulk file for download 


jvwong:
## Description

**Q: What is the name of the feature?**

A: Bulk data files 

**Q: What does this feature enable the user to do?**

A: Download files containing pathway data in various formats 

**Q: What information must the user provide to use the feature?**

A: n/a

**Q: How does this feature affect each class of user (persona)?**

A: This is targeted mainly at computational biologists for data analysis applications and other biological databases that may wish to integrate this data.

### Details

- Make available for download
  - BioPAX
  - SIF
  - SBGNML
  - JSON
- Download site?

metincansiper:
@jvwong 
- Do we want them to be both downloadable seperately and as bulk?
- Should the bulk file be a zip file that constains all of the exported files for each format?

jvwong:
> @jvwong
> 
>     * Do we want them to be both downloadable seperately and as bulk?
> 
>     * Should the bulk file be a zip file that constains all of the exported files for each format?

I would say export one single ('bulk') file for each format. See [PC](https://www.pathwaycommons.org/archives/PC2/v12/).

Possible exception is SBGNML: is this even worth exporting? This is where the webservice was useful.

metincansiper:
>I would say export one single ('bulk') file for each format. 

Do you mean one single bulk file that contains each of the formats. I mean something like:

```

- modelid.zip
   - modelid.owl
   - modelid.sif
   - modelid.json

```

The naming would be different.

jvwong:
I meant one single file for each format, zipped:

```
- biofactoid.owl.zip
- biofactoid.sif.zip
- ...
```

metincansiper:
But will there be only one file compressed in each zip?

jvwong:
Yes. Does that make sense?

metincansiper:
Yes actually I guess the only concern is probably decreasing the file size. Initially I thought that we will have a zip file that will include all since we mentioned a "bulk" file. Do you want me to add UI items to download that zip files?

jvwong:
Just create the data dumps for now and we can talk about where to download later...

metincansiper:
Sorry, I am still confused then. What I have initally thought was that we would create these files per document upon user request. 

Now I suspect that you want to create each of these for all document together I mean ``biofactoid.owl.zip`` will include the owl files for each documents. That is why you want data dumps to be created before hand.

Which one is correct?

maxkfranz:
[...]

PR #725, opened 2020-05-09
Offer contributed data as bulk file for download 

metincansiper:
Refs #717

### DIFF
--- a/src/util/document.js
+++ b/src/util/document.js
@@ -1,0 +1,29 @@
+import fetch from 'node-fetch';
+
+import { tryPromise } from './promise';
+
+function convertDocumentToBiopax(docId, baseUrl = '') {
+  let SERVER_URL = baseUrl + `/api/document/biopax/${docId}`;
+  let makeRequest = () => fetch(SERVER_URL);
+  return tryPromise( makeRequest ).then( result => result.text() );
+}
+
+function convertDocumentToTxt(docId, baseUrl = '') {
+  let SERVER_URL = baseUrl + `/api/document/text/${docId}`;
+  let makeRequest = () => fetch(SERVER_URL);
+  return tryPromise( makeRequest ).then( result => result.text() );
+}
+
+function convertDocumentToJson(docId, baseUrl = '') {
+  let SERVER_URL = baseUrl + `/api/document/${docId}`;
+  let makeRequest = () => fetch(SERVER_URL);
+  return tryPromise( makeRequest ).then( result => result.text() );
+}
+
+function convertDocumentToSbgn(docId, baseUrl = '') {
+  let SERVER_URL = baseUrl + `/api/document/sbgn/${docId}`;
+  let makeRequest = () => fetch(SERVER_URL);
+  return tryPromise( makeRequest ).then( result => result.text() );
+}
+
+export { convertDocumentToBiopax, convertDocumentToTxt, convertDocumentToJson, convertDocumentToSbgn };
--- a/src/util/document.js
+++ b/src/util/document.js
@@ -1,6 +1,7 @@
 import fetch from 'node-fetch';
 
 import { tryPromise } from './promise';
+import { checkHTTPStatus } from './fetch';
 
 function convertDocumentToBiopax(docId, baseUrl = '') {
   let SERVER_URL = baseUrl + `/api/document/biopax/${docId}`;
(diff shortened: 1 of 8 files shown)


src/util/document.js
```diff
@@ -0,0 +1,29 @@
+import fetch from 'node-fetch';
```
jvwong, line 1: I'm not sure this file is necessary. Also node-fetch may be possible issue if used in browser?

src/util/document.js
```diff
@@ -0,0 +1,29 @@
...
+
+import { tryPromise } from './promise';
+
+function convertDocumentToBiopax(docId, baseUrl = '') {
+  let SERVER_URL = baseUrl + `/api/document/biopax/${docId}`;
+  let makeRequest = () => fetch(SERVER_URL);
```
jvwong, line 7: `node-fetch` won't differentiate a response based on HTTP status. So you could get garbage in the response. I made a `util/fetch.js` helper for this.

src/util/document.js
jvwong: > I'm not sure this file is necessary.

Ok I see you have a client-side module.

src/util/document.js
```diff
@@ -0,0 +1,29 @@
...
+  return tryPromise( makeRequest ).then( result => result.text() );
+}
+
+function convertDocumentToTxt(docId, baseUrl = '') {
```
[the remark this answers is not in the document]
metincansiper: Okay I can exclude txt from the bulk filw. However, I guess you do not mean to remove ``exportDocumentToTxt()`` in ``client-util`` as well? n that case I think that this function would stay just to be called from client side and would not be included in the bulk file.

metincansiper:
@jvwong I made the requested changes. For now I excluded the text conversion just from the bulk file.

src/util/document.js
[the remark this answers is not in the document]
jvwong: > Okay I can exclude txt from the bulk filw. 

Fine with me.

src/util/document.js
maxkfranz: I'm pretty sure `fetch()` should just be available globally on the server.  Doing separate `import`s each time fetch is used on the server could make for more refactoring work later if we ever have to swap out the polyfill.

maxkfranz:
Looks good to me.  

Did we decide on a final name for the route?  Is there a reason to use `/api/document/zip/dl` versus `/api/document/zip`?

metincansiper:
>Did we decide on a final name for the route? Is there a reason to use /api/document/zip/dl versus /api/document/zip?

When I tried to make it ``/api/document/zip`` it did not work because I guess that it conflicted with another service like ``/api/document/${id}`. where id is ``zip``.

jvwong:
Re the route, you can just move it up. 
Also can you add the Swagger for this route?

src/util/document.js
metincansiper: It gives an error by telling that fetch is not defined in server side if I do not import fetch. I tried that it still works in client side as well when I import the node-fetch.

metincansiper:
>Re the route, you can just move it up.

I could not understand this. Do you mean adjusting it as ``api/zip`` or ``api/export/zip``? If so actually ``document`` was very convenient place in terms of implementation. 

>Also can you add the Swagger for this route?

Sure I  can do that.

metincansiper:
>I could not understand this. Do you mean adjusting it as ``api/zip`` or ``api/export/zip``? If so actually ``document`` was very convenient place in terms of implementation. 

Okay I guess you mean just moving the function up in the file before the conflicting service. Looks like that works I can do that.

metincansiper:
>Okay I guess you mean just moving the function up in the file before the conflicting service. Looks like that works I can do that.

@jvwong but would it be 100% safe? I cannot imagine any doument with an id like "zip" but would not it still be a confusion?

metincansiper:
I added the swagger for the route I can rename it along with relocating the service in the file after the final decision is made on that.

maxkfranz:
The ids are uuids. The only exception is the special ‘demo’ case

jvwong:
@metincansiper  I'm going to push an update to the Swagger doc description...

metincansiper:
>@metincansiper I'm going to push an update to the Swagger doc description...

Okay, sounds good.

>The ids are uuids. The only exception is the special ‘demo’ case

Okay then I will relocate and rename the service as Jeff suggested.

metincansiper:
@maxkfranz @jvwong  Okay, I relocated the api to ``api/document/zip``.

jvwong:
Works for me.

PR #730, opened 2020-05-22
CTA redesign

maxkfranz:
Here is the start of a redesign of the CTA.  The main change thus far is the automatic opening of the editor.  The second screen of the CTA is still necessary, to allow the user to get back to the editor if it's accidentally closed.

Use this PR for discussion of the redesign, and feel free to add commits to the branch fi you want to add your own changes.

Here's a preview:

Untitled23


jvwong:
User feedback I got on the form was 'let me know why you need this info' which I explicitly declare here. I moved the email message away from the subsequent 'done' panel here, for that same reason.

image

image

Issue #729, opened 2020-05-21
When article metadata is missing


jvwong:
## Background

*Issue*: No article metadata is associated with a Document

### Consequences

  - Fail to 'attribute' the authors and journal
  - Data integration suffers when pathway data and exports omit article references
  - Explore view paper metadata will be emptyish or worse, have,dead links (e.g. Google Scholar, PubMed)
    - Related papers system will be unable to sort articles based on participant queries (INDRA)
  - Non starter for 'assistants' based upon automated reading etc...

### Reasons for missing metadata

- A. Not in PubMed
  - Prepublication: The metadata exists, but is yet to be deposited by journal. Acceptance-to-publication lag ~4-6 weeks
  - Exclusion: PubMed does not index (e.g. preprints)
  - Nope: Spam

- B. In PubMed
  - Technical glitch: EUTILS service was down or errored when called by Factoid server
  - Clash: There was > 1 result when PubMed EUTILS service was queried (article title)
  - Related: PubMed metadata grows over time as curation occurs (e.g. MeSH headings, reference lists)

### Obtaining article metadata

- Manual: Authors enter/edit
- Automated: Software retrieves/replaces existing data

### Details

- Manual retrieval: This covers the (A) 'Prepublication' case
  - Idea is to format the available metadata information consistently (e.g. explore view, homepage gallery), and make it clear which fields are 'stubbed out'; then allow author to add or edit -- at very least journal name, **maybe** more
  - Where: Call to action; Editor; Submit

- Automated retrieval: This covers the (A) 'Prepublication' and (B) 'Technical glitch', 'Related'
  - Idea is to use existing service to fetch/refresh the metadata from PubMed based on the information provided by authors
  - Possible approaches
    - Cron job: Run the update periodically, optionally filtering some Documents (i.e. articles published a certain time ago)
    - other: Lazy load information when the Document is accessed (e.g. explorer, data export)

### Mockup

//TODO

maxkfranz:
Re. the manual case: The editor is probably a better fit than the CTA for the user entering these details.  An author has only one opportunity to address that in the CTA, and it would be blocking -- introducing friction.  I think it's better to just let them optionally edit a few details within the editor, if absolutely necessary.  

It's better all around to just use the automated approach.  If we're concerned about cases where the paper isn't yet in PubMed, it would probably be more effective and easier for users to just automatically pull in the information from Biorxiv if the information is not yet available in Pubmed.

[...]

PR #735, opened 2020-05-28
Cron system: Update Article metadata

jvwong:
Introduces a CRON system for updating the article metadata from Pubmed.

- General
  - CRON system is scheduled via config variable `CRON_SCHEDULE` to trigger updates provided to it  
- Document updates implemented 
  - article metadata (PubMed)
    - Documents to update are filtered based upon
      - Not the DEMO
      - Status not 'trashed'       
      - Missing values for doi or pmid 
      - Optional config `DOCUMENT_CRON_CREATED_AGE_DAYS` to filter out those created more than this many days ago (seems like acceptance to publication timeline is ~1-2 months)

Refs #729

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,12 @@ General:
 - `BASE_URL` : used for email linkouts (e.g. `https://factoid.baderlab.org`)
 - `API_KEY` : used to restrict new document creation (e.g. `8365E63B-9A20-4661-AED8-EDB1296B657F`)
 
+CRON:
+
+- `CRON_SCHEDULE` : second (optional), minute, hour, day of month, month, day of week
+- `DOCUMENT_CRON_UPDATE_PERIOD_DAYS` : The minimum time between successive Document cron update calls
+- `DOCUMENT_CRON_CREATED_AGE_DAYS` : Only Documents created fewer than this many days will be selecte for update. When undefined (default), ignores filtering on creation date.
+
 Database:
 
 - `DB_NAME` : name of the db (default `factoid`)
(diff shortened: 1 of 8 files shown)


maxkfranz:
One quick piece of feedbacks is that we may want to update docs even if they have an existing pmid. Updating old data may be useful in addition to filling in kissing data

jvwong:
> One quick piece of feedbacks is that we may want to update docs even if they have an existing pmid. Updating old data may be useful in addition to filling in kissing data

Definite over-optimisation at this point (nothing to update currently).

jvwong:
Maybe a better solution is a config flag: `DOCUMENT_CRON_REFRESH_ARTICLE` or something to indicate that it should include docs with data.

maxkfranz:
I agree that refreshes of prior hits may be too early to consider for this
PR, but refreshing existing hits will become more important when it's
considered it in tandem with related papers cron jobs.

Many of these cron-job-style refreshes may be a bit of an over-optimisation
for the general homepage user flow.  As a prospective user, why would I
create a shareable summary of my paper if I can't yet actually share the
link to my paper itself?  The timing and incentives are off.

It's still good to think about refreshes, and it may be useful in some
cases (e.g. enthusiastic authors who take action on the Cell acceptance
letter blurb).  Having config values and flags to toggle these cron
features would be good.  If we decide, for example, that we don't need it
enabled right away, then we can switch it off temporarily with the config
flag.

jvwong:
While we're here, I added a flag `DOCUMENT_CRON_REFRESH_ENABLED` that toggles the refresh of Document metadata.

Right now, everything that is not demo or trashed gets updated.

jvwong:
Travis is failing because of rethinkdb install on VM, not the code.

maxkfranz:
﻿I think we disabled the travis unit tests because travis was flakey with rethinkdb.  It may make sense to just remove all the rehinkdb stuff from the travis config completely if we keep getting false errors

Issue #742, opened 2020-06-03
No related papers results


maxkfranz:
I tried using the related papers feature by submitting a simple paper in a local instance.  The instance points to the public instance of the semantic-search service.

Here is the paper: [MeCP2 Facilitates Breast Cancer Growth via Promoting Ubiquitination-Mediated P53 Degradation by Inhibiting RPL5/RPL11 Transcription](https://pubmed.ncbi.nlm.nih.gov/32483207).  The Pubmed metadata is populated, so this issue seems to be distinct from #738.

The document is just 'MDM2 inhibits P53 via binding'.  I know that the paper would have more interactions, but this single interaction should be sufficient for testing.

<img width="1306" alt="Screen Shot 2020-06-03 at 4 14 48 PM" src="https://user-images.githubusercontent.com/989043/83684762-9115a580-a5b5-11ea-8cc7-17df56b6865d.png">

There are errors in the console:

```
info:    Searching the related papers table for document  c95d7143-f950-4662-a4ad-d05677734783
info:    PATCH /api/document/status/c95d7143-f950-4662-a4ad-d05677734783/743fdc49-a8c6-4f81-a90d-6ae44bb7df49 500 251.431 ms - 142
[object Object]
```

And `doc.relatedPapers()` is `undefined`.

jvwong:
I could reproduce the missing relatedPapers last night but this morning it works fine?  Can you try again?

There is some somewhat related refactoring that needs to be done, (another PRs):
  - There's a missing await keyword in `tryTweetingDoc`. Tweeting without the secret/key is the source of the HTTP 500, but isn't caught in the try-catch
  - The submit UI success panel doesn't wait for server response, resulting in Tweet link missing

jvwong:
Just a sec, the is error specific to your example with INDRA: human mdm2 and p53...

jvwong:
My best guess is that the semantic-search instance isn't scaling well with the number of papers (aka request body `documents`) it is asked to process.

I was trying three cases:

- A. Few papers (n=2)
  - Input: DAPK1, RIG-I
  - Result: OK, returns in seconds
- B. Many papers (n=93)
  - Input: MYC, E2F1
  - Result: OK, returns in minutes
- C. Waaaay too many papers (n=1029)
  - Input: p53, mdm2
  - Result: semantic-search returns error, not JSON, indra.js `getSemanticScores` throws

jvwong:
[...]

PR #750, opened 2020-06-11
No related papers results #742

metincansiper:
Refs #742 

I filtered the N most current papers before running the semantic search and anything else as discussed in the issue.

I made the filtering when SORT_BY_DATE flag is unset which means that the sort must be based on the semantic search. I can add either of the followings as well:

- Filter the top N results when SORT_BY_DATE flag is set as well.
- Run the semantic search only for the most current N papers but still include the others as well in the resulting list. Probably sort the others by the date.

I used 30 as the value of N that would be subject to change.

### DIFF
--- a/src/server/routes/api/document/indra.js
+++ b/src/server/routes/api/document/indra.js
@@ -144,18 +144,25 @@ const getDocuments = ( templates, queryArticle ) => {
         let dateJs = article.Journal.JournalIssue.PubDate;
         let { Day: day, Year: year, Month: month, MedlineDate: medlineDate } = dateJs;
 
-        if ( medlineDate ) {
-          return new Date( medlineDate );
+        try {
+          if ( medlineDate ) {
+            return new Date( medlineDate ).getTime();
+          }
+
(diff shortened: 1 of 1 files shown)


src/server/routes/api/document/indra.js
```diff
@@ -133,17 +137,51 @@ const getDocuments = ( templates, queryArticle ) => {
...
-      .then( semanticScores => ( { pubmeds, semanticScores } ) );
+      .then( semanticScores => ( { articles, semanticScores } ) );
+    };
+
+    const getPubTime = article => {
+        let dateJs = article.Journal.JournalIssue.PubDate;
+        let { Day: day, Year: year, Month: month, MedlineDate: medlineDate } = dateJs;
+
+        if ( medlineDate ) {
+          return new Date( medlineDate );
```
jvwong, line 148: There's a strong probability that this will throw.

metincansiper: Okay, I have returned the minimum possible integer in such cases.

Issue #743, opened 2020-06-04
On submit: Wait for server response (Tweet) before showing done panel


jvwong:
The submit panel should wait for the server to respond before showing the 'done' panel. In particular, assess the status of the Tweet information.

This was previously implemented, but may have been lost in refactoring.

Refs https://github.com/PathwayCommons/factoid/issues/742#issuecomment-638926526

jvwong:
I'm still not nailing doc updates happening on the client vs server, in particular doc status (i.e. 'submitted')....

maxkfranz:
If this issue is difficult to pinpoint, we could do a review of the existing code on a screen-share call

jvwong:
@maxkfranz OK, this is my mount everest. I think we need to talk about this.

jvwong:
Tweet: As far as I can tell, the round trip from client submit --> server --> db is fine. 

However, the  client (tasks.js) seems to be holding onto a stale version of 'document' so it doesn't see the 'tweet'.

jvwong:
nevermind this was a build issue.

PR #751, opened 2020-06-11
Editor submit & success panel 

jvwong:
Simplified the logic for Editor submit to publish.  

Third times a charm!

Refs #743

### DIFF
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -135,6 +135,13 @@ class Editor extends DataComponent {
       }
     });
 
+    doc.on('update', change => {
+      if( _.has( change, 'status' ) ){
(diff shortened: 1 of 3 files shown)


src/client/components/editor/index.js
```diff
@@ -135,6 +135,13 @@ class Editor extends DataComponent {
       }
     });
 
+    doc.on('update', change => {
+      if( _.has( change, 'status' ) ){
+        const isDone = this.data.document.submitted() || this.data.document.published();
+        this.done( isDone );
+      }
+    });
```
maxkfranz, line 143: It may be a bit simpler to listen to the `status` event, but this works too.  Not sure whether the `remotestatus` event was ever hooked up.

Issue #738, opened 2020-06-01
Related Papers: Search aborts when (PubMed) metadata is missing


jvwong:
This concerns retrieving related papers for a given non-empty document. There are two tasks:
  - 1. Retrieve articles (INDRA), based upon the interaction participants present
  - 2. Sort articles (semantic-search), using metadata from the document's associated article

Major issue: If there is no article metadata associated with the document (no abstract), the `searchRelatedPapers` service returns immediately (empty list). The suggestion is to retrieve  articles (i) independent of the ability to sort (ii).

Minor: When no abstract, `searchRelatedPapers`  returns array rather than Promise. This leads to error `searchRelatedPapers( ... ).then is not a function` in no abstract cases and fails to engage the catch in `updateRelatedPapers`

gbader:
Maybe in the future we could generate some text for the similarity search to work on - something like converting the factoid data to some sentences using templates.

maxkfranz:
A text exporter like that already exists, so it may be worth a shot trying this soon. I’m not sure how effectively the templated text would work in place of a real abstract, but it’s probably better than nothing.

PR #752, opened 2020-06-12
Related Papers: Search aborts when (PubMed) metadata is missing 

metincansiper:
Refs #738

### DIFF
--- a/src/server/routes/api/document/indra.js
+++ b/src/server/routes/api/document/indra.js
@@ -2,7 +2,7 @@ import fetch from 'node-fetch';
 import _ from 'lodash';
 import { parse as dateParse } from 'date-fns';
 
-import { INDRA_DB_BASE_URL, INDRA_ENGLISH_ASSEMBLER_URL, SEMANTIC_SEARCH_BASE_URL } from '../../../../config';
+import { INDRA_DB_BASE_URL, INDRA_ENGLISH_ASSEMBLER_URL, SEMANTIC_SEARCH_BASE_URL, NO_ABSTRACT_HANDLING } from '../../../../config';
 import logger from '../../../logger';
 import { tryPromise } from '../../../../util';
 import { INTERACTION_TYPE } from '../../../../model/element/interaction-type/enum';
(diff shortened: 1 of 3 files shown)


src/server/routes/api/document/indra.js
```diff
@@ -91,27 +93,49 @@ const getDocuments = ( templates, queryArticle ) => {
...
-
-      let url = SEMANTIC_SEARCH_BASE_URL;
+      let queryArticle = queryDoc.article();
       let queryText = getAbstract( queryArticle );
+
+      if ( queryText == null ) {
+        queryText = handleNoQueryAbastract( queryDoc );
+      }
+
       let queryUid = getPmid( queryArticle );
```
jvwong, line 136: If there's no PubMed data, then `getPmid( queryArticle )` will throw since accessing `id` == undefined

src/server/routes/api/document/indra.js
```diff
@@ -91,27 +93,49 @@ const getDocuments = ( templates, queryArticle ) => {
...
+      if ( _.isString( abstract ) ) {
+        return abstract;
+      }
+      if ( _.isArray( abstract ) ) {
+        return abstract[ 0 ];
+      }
+      return null;
+    };
+
+    const getPmid = article => article.PubmedData.ArticleIdList.find( o => o.IdType == 'pubmed' ).id;
```
jvwong, line 121: `PubmedData` may not exist: https://dtd.nlm.nih.gov/ncbi/pubmed/doc/out/180101/el-PubmedArticle.html

jvwong: In our case, the fetchPubmed will have `PubmedData: null`. I use lodash _.get for these cases.

metincansiper: Okay, I updated the method in a safer way.

src/server/routes/api/document/indra.js
```diff
@@ -133,7 +144,10 @@ const getDocuments = ( templates, queryDoc ) => {
         queryText = handleNoQueryAbastract( queryDoc );
       }
 
-      let queryUid = getPmid( queryArticle );
+      // semantic search api requires uid for query but there is no
+      // use case of it at least for us. Therefore, it must be
+      // fine to generate a random uuid when the pmid is not available.
```
jvwong, line 149: You're right. At the time, I thought semantic-search service could be responsible for fetching article Abstract/text and the API would be a list of query and document `uids` only which would be more lightweight.

maxkfranz: That would be ideal.  That would be good to have after improving the results of the related papers system -- so that it doesn't return zero results or one result much of the time.

maxkfranz:
@metincansiper, it looks like there's some merge conflicts that need to be addressed before this can be closed.

metincansiper:
@maxkfranz I resolved the merge conflicts.

Issue #753, opened 2020-06-15
Homepage: Tweaks


jvwong:
The following homepage-related items originate from discussion Jun.2.2020.

- Header
  - Contact: Biofactoid is a free academic project by: Bader Lab at the University of Toronto, Sander Lab at Harvard, and the Pathway and Omics Lab at the University of Oregon.

- (middle section) How it works
  - It’s quick and easy to share your published results by drawing the key interactions in your pathway in Biofactoid. Your results are automatically made available for others to explore and cite.
- (bottom section) We publish and share an interactive graphical abstract of your pathway so scientists can find and use the information.
  - Popups
    - 'publish and share': Explain that the information goes into a database and is posted on social media.
    - 'interactive graphical abstract': hyperlink to example?
  - Figure:
    - Fade out the margins, retain white background
    - logo atop / outside
    - Sections: Explore, Discover, Pathways
      - Share
        - Article, visualize, social media
      - Explore
        - Related articles, Related pathways, databases
      - Discover
        - New genes and interactions
        - (Via) Search, integrate, analyze

- CTA: Create your graphical abstract - or - Enter your article information
  - A private editing link will be sent to your email. We will never share your email address.

- Footer
  - Institutions: 
    - too tiny text
    - add color, icons good enough, mouseover names
    - grant numbers

PR #754, opened 2020-06-17
Iss753 homepage tweaks text

jvwong:
- Section 'Your pathway to share and explore'
  - Popup info for 'publish and share' 
  - Updated network image
- Contact: Funding sources appended
- Footer: Logos link out to institutions

Refs #753


jvwong:
- Contact info

image

- Your pathway to share and explore

image

maxkfranz:
Looks good.

The only suggestion I have is around the underlined links.  They can be hard to see in this section:

Screen Shot 2020-06-19 at 12 44 57 PM

Maybe it would be better if the underline were just white for all of those captions.

jvwong:
If you notice, I actually flipped the background image so that it was a bit darker on left.

maxkfranz:
Yeah, I noticed.  The image looks better that way, but I think the blue underline is hard to see in any case

Issue #736, opened 2020-05-29
Grounding: Cannot 'Select better match for X'


jvwong:
In a manual grounding of a node, the option to 'Select a better match for "X"' does not provide a the additional list of entity records. 

Reproduce:
  - Add entity
  - Type "p53"
  - Click Select a better match for "p53"

The tooltip simply closes.

jvwong:
This actually applies to the edge tooltip as well

maxkfranz:
It seems to be universal.  I’ve taken a look and I may have a fix

PR #755, opened 2020-06-17
Related papers UI

maxkfranz:
- Adds related papers UI, labelled as "recommended articles", for the document, each interaction, and each entity.
- Revises the carousel component to be generic.  The carousel is an alternative way to show the related papers that allows for showing more papers with horizontal scrolling.  This design can be swapped out in future, as needed.
- The information panel in the explore view always sits at the bottom of the page, regardless of screen size.
- Large screens allow the related papers to be shown above the fold, by using a two-column layout.
- Misc. fixes & improvements
  - The information panel can be scrolled by scrolling over the network visualisation.
  - Improved entity information display in the explore view:  It now reuses the information panel rather than showing Tippy components.
  - The use of Tippy components has been revised to address #736: Grounding: Cannot 'Select better match for X'
  - @metincansiper The `relatedPapers` method has been moved to the `Element` superclass to support both `Entity`s and `Interaction`s.  See https://github.com/PathwayCommons/factoid/issues/742#issuecomment-644228132
  - New nodes are always added around the same position to allow for more consistency and to allow for a better experience on small screen sizes.  (You need the text box in the popover to be shown near the top of the screen.)
  - Many gestures have been disabled in the visualisation of the explore view.
  - There is a button in the explore view to access the corresponding tweet.
  - And other minor usability improvements...

The default view on desktop:

def

The view on desktop after scrolling the page down to view more of the info panel:

info

The default view on mobile:

mobile


maxkfranz:
@jvwong This PR has the data for the related papers stubbed out because of the issues around #742.  If you think it would be useful for testing, you could could swap out the values to read the real fields:

https://github.com/PathwayCommons/factoid/blob/58b97a476bc40afb2a670edac55f3d1603ebee05/src/client/components/related-papers.js#L12-L20

I haven't had any luck with the service, so I've left in the placeholder values for now.

jvwong:
Twitter icon appears off:

image

jvwong:
> @jvwong This PR has the data for the related papers stubbed out because of the issues around #742. If you think it would be useful for testing, you could could swap out the values to read the real fields:
> 
> https://github.com/PathwayCommons/factoid/blob/58b97a476bc40afb2a670edac55f3d1603ebee05/src/client/components/related-papers.js#L12-L20
> 
> I haven't had any luck with the service, so I've left in the placeholder values for now.

I have this working live, but it occurs to me that doc.relatedPapers should be storing the output of  /util/pubmed 'getPubmedCitation' which is nice and neat and forgiving rather than the raw PubMed data. I can look into that for this branch.

maxkfranz:
> I have this working live, but it occurs to me that doc.relatedPapers should be storing the output of /util/pubmed 'getPubmedCitation' which is nice and neat and forgiving rather than the raw PubMed data. I can look into that for this branch.

Yeah, I agree.  The raw data seemed a bit fragile to use directly in the UI.

jvwong:
Worst case scenario: 

image

Good test case:

image

maxkfranz:
OK, that's good enough for testing.  See #757 re. no-papers-yet feedback

Issue #741, opened 2020-06-02
Pubmed query flakiness


maxkfranz:
Pubmed queries can return no results, even when there's a perfect match in their database.  This makes it impossible to associate a factoid with an article.  For example, consider the following article:

[Wild-type p53 Oligomerizes More Efficiently Than p53 Hot-Spot Mutants and Overcomes Mutant p53 Gain-Of-Function via a "Dominant-Positive" Mechanism](https://pubmed.ncbi.nlm.nih.gov/30174797)

If you search using the API and the exact title, you get no results.  If you search using the Pubmed search box, you go directly to the paper.  I suspect the NCBI query format is to blame.  It uses a SQL-like query, but that doesn't fit our use case.  It helps in some cases to set the query to use `field=title`, but it doesn't seem to solve the issue completely.

We could just scrape the the new Pubmed search page instead of using the E-utilities search API, e.g.:

https://pubmed.ncbi.nlm.nih.gov/?term=Wild-type+p53+Oligomerizes+More+Efficiently+Than+p53+Hot-Spot+Mutants+and+Overcomes+Mutant+p53+Gain-Of-Function+via+a+%22Dominant-Positive%22+Mechanism

The results page includes a `<meta>` tag that includes the article ID for tools and search engines.  This might not be quite as clean of a solution as using the API directly, but it would be simple and reliable.

Based on some testing with random article titles in Pubmed,  it seems that a not insignificant number of titles fail with the API approach.  I think that it's worth a shot to try scraping since it's cheap to implement.

This approach may also resolve the missing metadata issues (#729, #738) to a large degree.  The results page also includes meta tags for the author list, the journal name, abstract, etc.  Further, articles are published to the Pubmed UI faster than they are available in the API.

gbader:
What about using Google?  It will probably do a better job finding the pubmed ID or DOI or other metadata from a standard database based on the title.

maxkfranz:
The scraping method would only be necessary once per paper, at the time that the user clicks the CTA button on the homepage.  Any heavy tasks, like refresh cron jobs, would be able to use the Pubmed API normally.

The meta tags in the new Pubmed UI strongly suggest that expect these kinds of queries are expected -- that's the purpose of those tags.  Would the people at Pubmed really be against such infrequent requests?  We're not talking about scraping all of Pubmed with thousands of successive requests here:  It's one little request once in a while.

maxkfranz:
[...]

PR #756, opened 2020-06-18
Pubmed articles by paperId

jvwong:
- Search pubmed accepts optional configuration parameter, enabling search by `field: "title"`
- Pubmed 'finding' module now accepts:
  - 1. exact doi
  - 2. exact pmid
  - 3. title

Discriminating input: DOI regex provided by Crossref that hits 99.3%. To my knowledge, CrossRef is the primary registrar for research articles. PMIDs just look like plain number.

Double checking search results: Previously, if one search result came back it was assumed to be a match, and there were edge cases. Now, the single search result is checked against the input. The title check is a little more fragile, so I only check if input 'word' characters is contained in the title of the search hit.  

I was looking for more title test cases...

Refs #741

### DIFF
--- a/src/server/routes/api/document/pubmed/index.js
+++ b/src/server/routes/api/document/pubmed/index.js
@@ -70,7 +70,12 @@ const getPubmedArticle = async paperId => {
       const PubmedArticle = _.head( PubmedArticleSet );
       const { title, pmid, doi } = getPubmedCitation( PubmedArticle );
       // Pubmed EFETCH returns lower-cased title, adds trailing period
-      const cleanTitle = raw => _.toLower( _.trim( raw , ' .') );
+      const santitize = raw => {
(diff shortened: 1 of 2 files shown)


maxkfranz:
Here's are some test cases that doesn't work on the CTA:

- Cardiovascular Disease and COVID-19
- Coronavirus Disease 2019 (COVID-19): A Clinical Update

I tried some random COVID-19 ones from Pubmed.  A lot of them worked.  Some, like the above, did not.

Maybe we should be using Crossref to query by title.  It seems to work OK, e.g. https://api.crossref.org/works?query.bibliographic=Cardiovascular%20Disease%20and%20COVID-19

maxkfranz:
We could do something like this:

Crossref search by title => DOI => Pubmed query by DOI => Pubmed data

PR #758, opened 2020-06-19
Entity related papers

metincansiper:
Refs #742

Handles this issue comment (https://github.com/PathwayCommons/factoid/issues/742#issuecomment-644228132)

### DIFF
--- a/src/model/document/document.js
+++ b/src/model/document/document.js
@@ -465,8 +465,10 @@ class Document {
 
   toSearchTemplates(){
     let interactions = this.interactions();
-    let templates = interactions.map( i => i.toSearchTemplate() ).filter( t => t != null );
+    let entities = this.entities();
+    const getTemplates = els => els.map( e => e.toSearchTemplate() ).filter( t => t != null );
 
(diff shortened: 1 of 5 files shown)


jvwong:
I'm getting an error when it comes to storing the results in doc. Does `doc.get` allow  entity ID as parameter?

https://github.com/PathwayCommons/factoid/blob/c19ade5a75b309fe9ada4d1d87b804c5810099f0/src/server/routes/api/document/index.js#L135

jvwong:
This catch should print out `e.message` 

https://github.com/PathwayCommons/factoid/blob/c19ade5a75b309fe9ada4d1d87b804c5810099f0/src/server/routes/api/document/index.js#L1382

metincansiper:
@jvwong the problems that you mentioned must be solved. Would you review again?

metincansiper:
>Does doc.get allow entity ID as parameter?

Yes I think so. The problem was just because of using the wrong id to search.

metincansiper:
>OK, let’s just merge this in and try it out on unstable. It would be good to address sorting in another PR, and in the meanwhile we can see about the GPU speedup

Okay, I merged it.

Issue #759, opened 2020-06-23
Remove 'context' related to user workflows  


jvwong:
There is a 'context' flag that enabled us to switch between homepage 'signup' and 'journal' users who would receive tailored workflows, which amounted to custom emails. This added a bunch of complexity and unnecessary dependencies to the Mailjet templates and the app. For instance, the admin page now erroneously tags all  docs as having been routed through the journal (bug).

The proposal is to expunge all context, and assume all users will proceed via the homepage since it is clear that that journal users will be funnelled there.

jvwong:
Sync and cleanup the email templates too.

PR #761, opened 2020-06-24
Remove 'context'

jvwong:
This PR removes all references to 'context' which was a flag that differentiated the route through which a document was created (journal, homepage). All paths lead through homepage now.

Emails: Given that all users will 'know' about the system already, the  server-side code was stripped of all information other than that required to link to the editor/document. Accordingly, the email invite template have been simplified to only provide the link to the editable doc. Also updated the email template text to conform to the homepage language.

Refs #759


jvwong:
Updated signup email:

image

Issue #757, opened 2020-06-19
Feedback when related papers calculation is in progress


maxkfranz:
Even if #742 is resolved, we may have no related papers to show while the calculation is in progress.  This mostly affects the author, though it can affect readers if the author shares his factoid immediately after submission.

- [x] Give feedback to the user that the related papers are coming soon, e.g. "We're working on getting related articles.  We'll have something in a few minutes."  A spinner may be useful, but it may fit better if the related papers can be returned quickly (seconds instead of minutes).
- [x] Automatically show the related papers when the data has been added on the backend.  The live-sync model should allow for this.

Ref. #755 #742

PR #762, opened 2020-06-26
Related papers loading & notice

maxkfranz:
- Adds a notice to the related papers section when the calculation is in progress.
- Automatically adds the related papers to the UI when the server sets them.
- Renames `relatedPaper` event to `relatedpaper`, to be consistent with all other events.
- Limits the number of shown related papers to a maximum of six.
- Misc. fix to article title underline in the editor.


maxkfranz:
See #757

PR #764, opened 2020-06-30
Use Crossref for more reliable Pubmed searches

maxkfranz:
- Process:
  - Search CR, get the DOI for the paper.
  - Search PMC for DOI.  If we've got a hit, return
  - Otherwise, fall back on a PM search via the DOI.
- Crossref searches get transformed into the Pubmed search result format, so no other modules need to be changed.
- The existing `searchPubmed` module only has newly added functions.  Nothing has been removed.
- The hybrid approach seems to get a lot more hits.
- This PR is still a rough draft.  There are edge cases to consider, and the [request headers should be revised](https://github.com/CrossRef/rest-api-doc#etiquette).
- Ref. https://github.com/CrossRef/rest-api-doc

Ref. #763 #741

@jvwong, could you test this out and evaluate whether it's a feasible alternative?  There are probably also some robustness improvements you could make, based on your previous experience with Pubmed API searches.

### DIFF
--- a/src/server/routes/api/document/pubmed/searchPubmed.js
+++ b/src/server/routes/api/document/pubmed/searchPubmed.js
@@ -80,7 +80,7 @@ const crSearchPubmed = async ( term, IdType, opts ) => {
   // try PubMed ESEARCH
   const queryOpts = IdType === ID_TYPE.TITLE ? { field: ID_TYPE.TITLE } : {};
   const searchResponse = await eSearchPubmed( term, queryOpts );
-  if( searchResponse.count === -1 ){
+  if( searchResponse.count === 1 ){
     return searchResponse;
 
   } else {
--- a/src/server/routes/api/document/pubmed/searchPubmed.js
+++ b/src/server/routes/api/document/pubmed/searchPubmed.js
@@ -2,8 +2,11 @@ import _ from 'lodash';
 import queryString from 'query-string';
 import fetch from 'node-fetch';
 
-import { NCBI_EUTILS_BASE_URL, NCBI_EUTILS_API_KEY } from '../../../../../config';
+import { NCBI_EUTILS_BASE_URL, NCBI_EUTILS_API_KEY, CROSSREF_API_BASE_URL, PMC_ID_API_URL } from '../../../../../config';
 import { checkHTTPStatus } from '../../../../../util';
+
+// const CROSSREF_API_BASE_URL = `https://api.crossref.org`;
+const CROSSREF_WORKS_URL = `${CROSSREF_API_BASE_URL}/works`;
 
 const EUTILS_SEARCH_URL = NCBI_EUTILS_BASE_URL + 'esearch.fcgi';
 const DEFAULT_ESEARCH_PARAMS = {
@@ -52,6 +55,86 @@ const eSearchPubmed = ( term, opts ) => {
 };
 
 /**
+ * Search for an article using Crossref and return a Pubmed-style result.
+ * 
+ * This approach uses Crossref to find the article's DOI.  The DOI is sent to PMC to find a
+ * PMID.  If the PMC query fails for any reason, a query to Pubmed Eutils is made using the 
+ * DOI as the search term.  For some articles, PMC fails.  For others, Eutils fails.  Using
+ * both gives good coverage.
+ * 
+ * @param {String} term The search term to find the article (e.g. title)
+ * @param {Object} opts Pubmed API options
+ * @returns {Object} A Pubmed search result object
+ */
+const crSearchPubmed = async ( term, opts ) => {
+  const doi = await crSearchForDoi(term);
+
+  const pmid = await getPmidFromDoi(doi);
+
+  if( pmid ){
+    return {
+      searchHits: [ pmid ],
+      count: 1,
+      query_key: '-1',
+      webenv: 'imitated_pubmed_format_from_crossref_query'
+    };
+  } else {
+    return eSearchPubmed(doi, opts);
+  }
+};
+
+/**
+ * Search for an article's DOI using Crossref
+ * @async
+ * @param {String} term The search term (e.g. article title)
+ * @returns {String} The DOI, null if not found
+ */
+const crSearchForDoi = async ( term ) => {
+  const params = {
+    'query.bibliographic': term
+  };
+
+  const crUrl = CROSSREF_WORKS_URL + '?' + queryString.stringify( params );
+
+  const crRes = await fetch(crUrl, {
+    method: 'GET'
+  });
+
+  const crJson = await crRes.json();
+
+  const doi = _.get(crJson, ['message', 'items', 0, 'DOI']);
+
+  return doi;
+};
+
+/**
+ * Use the PMC API to get the PMID for an article from the DOI.
+ * @async
+ * @param {String} doi The article's DOI
+ * @returns {String} The article's PMID, null if not found (i.e. doesn't exist or pubmed flakes out)
+ */
+const getPmidFromDoi = async ( doi ) => {
+  const params = {
+    format: 'json',
+    ids: doi,
+    tool: 'Biofactoid',
+    email: 'info@biofactoid.org'
+  };
+
+  const url = PMC_ID_API_URL + '?' + queryString.stringify( params );
+
+  const res = await fetch(url, {
+    method: 'GET'
+  });
+
+  const json = await res.json();
+
+  let pmid = _.get(json, ['records', 0, 'pmid']);
+
+  return pmid;
+};
+
+/**
  * searchPubmed
  * Query the PubMed database for matching UIDs.
  *
@@ -63,6 +146,6 @@ const eSearchPubmed = ( term, opts ) => {
  * @returns { String } result.query_key See {@link https://www.ncbi.nlm.nih.gov/books/NBK25499/#chapter4.ESearch|EUTILS docs }
  * @returns { String } result.webenv See {@link https://www.ncbi.nlm.nih.gov/books/NBK25499/#chapter4.ESearch|EUTILS docs }
  */
-const searchPubmed = ( q, opts ) => eSearchPubmed( q, opts );
+const searchPubmed = ( q, opts ) => crSearchPubmed( q, opts );
 
 export { searchPubmed, pubmedDataConverter };
--- a/src/server/routes/api/document/pubmed/searchPubmed.js
+++ b/src/server/routes/api/document/pubmed/searchPubmed.js
@@ -2,8 +2,11 @@ import _ from 'lodash';
 import queryString from 'query-string';
 import fetch from 'node-fetch';
 
-import { NCBI_EUTILS_BASE_URL, NCBI_EUTILS_API_KEY, CROSSREF_API_BASE_URL, PMC_ID_API_URL } from '../../../../../config';
+import { NCBI_EUTILS_BASE_URL, NCBI_EUTILS_API_KEY,
+  EMAIL_FROM_ADDR, BASE_URL,
+  CROSSREF_API_BASE_URL, PMC_ID_API_URL } from '../../../../../config';
 import { checkHTTPStatus } from '../../../../../util';
+import { ID_TYPE } from './index';
 
 // const CROSSREF_API_BASE_URL = `https://api.crossref.org`;
 const CROSSREF_WORKS_URL = `${CROSSREF_API_BASE_URL}/works`;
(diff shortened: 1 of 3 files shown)


src/server/routes/api/document/pubmed/searchPubmed.js
```diff
@@ -51,6 +54,86 @@ const eSearchPubmed = ( term, opts ) => {
...
+  const res = await fetch(url, {
+    method: 'GET'
+  });
+
+  const json = await res.json();
+
+  let pmid = _.get(json, ['records', 0, 'pmid']);
+
+  return pmid;
+};
```
jvwong, line 135: This service doesn't appear to work with anything from Nature or Science. We're better off using the original `eSearchPubmed` given a doi.

e.g. 10.1126/science.abc1932 , 10.1038/s41586-020-2335-4

src/server/routes/api/document/pubmed/searchPubmed.js
```diff
@@ -51,6 +54,86 @@ const eSearchPubmed = ( term, opts ) => {
...
+
+/**
+ * Search for an article's DOI using Crossref
+ * @async
+ * @param {String} term The search term (e.g. article title)
+ * @returns {String} The DOI, null if not found
+ */
+const crSearchForDoi = async ( term ) => {
+  const params = {
+    'query.bibliographic': term
```
jvwong, line 94: - filter on type: `filter=type:journal-article`
- also `rows=1`

src/server/routes/api/document/pubmed/searchPubmed.js
```diff
@@ -51,6 +54,86 @@ const eSearchPubmed = ( term, opts ) => {
...
+ * This approach uses Crossref to find the article's DOI.  The DOI is sent to PMC to find a
+ * PMID.  If the PMC query fails for any reason, a query to Pubmed Eutils is made using the 
+ * DOI as the search term.  For some articles, PMC fails.  For others, Eutils fails.  Using
+ * both gives good coverage.
+ * 
+ * @param {String} term The search term to find the article (e.g. title)
+ * @param {Object} opts Pubmed API options
+ * @returns {Object} A Pubmed search result object
+ */
+const crSearchPubmed = async ( term, opts ) => {
```
jvwong, line 69: This should look at `opts` to see if the term looks like a 'title', 'doi' or 'pmid'. 

If its a pmid or doi, go direct to `eSearchPubmed` because the `crSearchForDoi` is hardcoded to fields`query.bibliographic`.

jvwong:
Should I make the changes?

src/server/routes/api/document/pubmed/searchPubmed.js
jvwong: NB: `filter=type:journal-article` will exclude `"subtype": "preprint"` i.e. bioRXiv.

src/server/routes/api/document/pubmed/searchPubmed.js
maxkfranz: It seems that neither service gives good coverage on its own

src/server/routes/api/document/pubmed/searchPubmed.js
maxkfranz: Pubmed also has preprint articles, though perhaps not as quickly as Biorxiv.  Would this filter affect those articles?

maxkfranz:
> Should I make the changes?

I think you have a better idea of the title versus DOI inference, so that makes sense.  I'm not sure whether `filter=type:journal-article` would have unintended side effects.

src/server/routes/api/document/pubmed/searchPubmed.js
jvwong: Oh boy that's news to me. I need to do some homework : https://www.ncbi.nlm.nih.gov/pmc/about/nihpreprints/

maxkfranz: I think COVID-19 motivated them to push that project forward.  There's a little marker on Pubmed search results for preprint articles.

src/server/routes/api/document/pubmed/searchPubmed.js
maxkfranz: PMC queries are first, because they have reliable hits:  It queries based on the DOI field, whereas Eutils doen't always pick up the DOI properly.  I agree that PMC doesn't cast as wide a net as we'd like, but I think it's a more reliable first pass.

jvwong:
> > Should I make the changes?
> 
> I think you have a better idea of the title versus DOI inference, so that makes sense. I'm not sure whether `filter=type:journal-article` would have unintended side effects.

Probably shouldn't do this given the preprint pilot.

Interesting interview, with where they seem to be going with this: https://asapbio.org/preprints-on-pmc-and-pubmed

jvwong:
> Aside: The Pubmed search service is failing more often than it seemed to in the previous issue (#741).

Some papers that don't work:

    GATA6 Exerts Potent Lung Cancer Suppressive Function by Inducing Cell Senescence
    miR-331-3p Inhibits Inflammatory Response After Intracerebral Hemorrhage by Directly Targeting NLRP6
    The Role of MDM2-p53 Axis Dysfunction in the Hepatocellular Carcinoma Transformation
    The Expression of p53, CD44, Ki-67, and HER-2/neu Markers in Gastric Cancer and Its Association With Histopathological Indicators: A Retrospective Study
    β-Sitosterol Alleviates Inflammatory Response via Inhibiting the Activation of ERK/p38 and NF- κ B Pathways in LPS-Exposed BV2 Cells

>Papers that worked:

    The Expression of p53, CD44, Ki-67, and HER-2/neu Markers in Gastric Cancer and Its Association With Histopathological Indicators: A Retrospective Study
    Phthalide Derivative CD21 Ameliorates Ischemic Brain Injury in a Mouse Model of Global Cerebral Ischemia: Involvement of Inhibition of NLRP3

@maxkfranz Hold on, I tried the above on unstable and I'm 7/7 (caveat, `β-Sitosterol Alleviates ..` has extra spaces between NF- k B). Maybe you were trying a stale version?

jvwong:
image

maxkfranz:
> @maxkfranz Hold on, I tried the above on unstable and I'm 7/7 (caveat, β-Sitosterol Alleviates .. has extra spaces between NF- k B). Maybe you were trying a stale version?

I was using unstable.  Has anything changed since I tried?

maxkfranz:
Maybe Pubmed updated their DB in the interim?

jvwong:
I'm not sure if this is related, but things will get stamped with a DOI (i.e. published) anywhere up to 3 weeks before they get indexed in pubmed. So maybe DOIs are getting indexed before text but I'm just guessing here.

jvwong:
Do you have an example that fails through PubMed ESEARCH?

maxkfranz:
I know some of the examples I posted were like that, but maybe that’s changed. I’ll take a look to see if I can find another one

maxkfranz:
I've tried more articles that are recent, and it seems Pubmed must have been updated.  The results are a lot better, even for recent preprint articles.  If they're working behind the scenes to make things more accurate and more recent, then we should just stick with Eutils search by title for now.

src/server/routes/api/document/pubmed/searchPubmed.js
jvwong: > This service doesn't appear to work with anything from Nature or Science. We're better off using the original `eSearchPubmed` given a doi.
> 
> e.g. 10.1126/science.abc1932 , 10.1038/s41586-020-2335-4

Weird, now these do work with PMC?....

maxkfranz: I suspect they're scrambling to keep up with all the preprints, especially the COVID-19 ones.  Their DB seems to be changing frequently, sometimes improving things and sometimes adding some instability.

The Pubmed queries seem like they're a bit variable, but they'll also probably get better with time.  So an alternative to using Crossref, at least for now, could be to just increase the metadata cronjob frequency.

If we do go with the hybrid Crossref route, then it may be useful to have a final fallback case that searches Eutils by title:

- First, try PMC by DOI
- If PMC didn't work, then try Eutils by DOI
- If Eutils DOI didn't work, try Eutils by title

jvwong: I have some updates cooking here that more or less do what you said. Will push soon

jvwong:
>* This approach uses a multi-stage approach:
> *   1. Query the PubMed EUTIL service
> *   2. If no unique hit, query Crossref API works resource component
> *   3. Send the CrossRef-derived DOI to PMC to map to a PMID
> *   4. If no PMID, try the Pubmed EUTILS service with DOI

src/server/routes/api/document/pubmed/searchPubmed.js
```diff
@@ -38,31 +56,109 @@ const checkEsearchResult = json => {
...
+ * @param {String} IdType One of ID_TYPE
+ * @param {Object} opts Pubmed API options
+ * @returns {Object} A Pubmed search result object
+ */
+const crSearchPubmed = async ( term, IdType, opts ) => {
+
+  // try PubMed ESEARCH
+  const queryOpts = IdType === ID_TYPE.TITLE ? { field: ID_TYPE.TITLE } : {};
+  const searchResponse = await eSearchPubmed( term, queryOpts );
+  if( searchResponse.count === 1 ){
```
maxkfranz, line 83: Would it be worthwhile to also check if the first result has an exact match of the title?  There may be edge cases where we get more than one result, but the top result is still the right match.

maxkfranz:
> * This approach uses a multi-stage approach:
> * 1. Query the PubMed EUTIL service
> * 1. If no unique hit, query Crossref API works resource component
> * 1. Send the CrossRef-derived DOI to PMC to map to a PMID
> * 1. If no PMID, try the Pubmed EUTILS service with DOI

This looks like a more conservative variation of the method, so that looks like the way to go.  The only part I'm unsure about is the "if no unique hit" condition.  Would it be worthwhile to loosen that condition a bit?

src/server/routes/api/document/pubmed/searchPubmed.js
```diff
@@ -17,6 +23,18 @@ const DEFAULT_ESEARCH_PARAMS = {
   api_key: NCBI_EUTILS_API_KEY
 };
 
+const safeFetch = url => {
+  const userAgent = `${process.env.npm_package_name}/${process.env.npm_package_version}`;
```
maxkfranz, line 27: Maybe the contact email could be included here.  Otherwise, I guess CR would have to google us if something goes wrong

jvwong: From their [docs](https://github.com/CrossRef/rest-api-doc): 
> Specify a User-Agent header that properly identifies your script or tool and that provides a means of contacting you via email using "mailto:". For example: GroovyBib/1.1 (https://example.org/GroovyBib/; mailto:GroovyBib@example.org) BasedOnFunkyLib/1.4.

maxkfranz:
Just a note for future reference:  I've done some looking around, and CR does have abstracts.  However, it's up to the journal to decide whether they want to include abstracts when they deposit articles to CR.  So even in the long term, Crossref won't suffice on its own.

src/server/routes/api/document/pubmed/searchPubmed.js
jvwong: That's doable: search through top 10 hits?

maxkfranz: Yeah, that's more robust than just looking at the top one

jvwong: Do you have a test case for this?

src/server/routes/api/document/pubmed/searchPubmed.js
jvwong: > > This service doesn't appear to work with anything from Nature or Science. We're better off using the original `eSearchPubmed` given a doi.
> > e.g. 10.1126/science.abc1932 , 10.1038/s41586-020-2335-4
> 
> Weird, now these do work with PMC?....

Today these don't work?

src/server/routes/api/document/pubmed/searchPubmed.js
maxkfranz: No, I don't have an example.  If it's a simple approach and we can verify that it doesn't break the main case (one result), then I think that's fine for now.  

It's hard to find persistent test cases when Pubmed is changing the data so frequently.

jvwong: Testing this is a big challenge. I actually ended up doing a major revamp of this in my branch which we can keep on the backburner. Will post ASAP.

jvwong:
FYI: Here's an example where Pubmed ESEARCGH will return a non-unique result (which our system correctly matches):

- Query: 'TRPV1 SUMOylation regulates nociceptive signaling in models of inflammatory pain'
  - ESEARCH result
    - 'Author Correction: TRPV1 SUMOylation regulates nociceptive signaling in models of inflammatory pain'
    - 'TRPV1 SUMOylation regulates nociceptive signaling in models of inflammatory pain'

jvwong:
Another ambiguous title: Structure-Mediated RNA Decay by UPF1 and G3BP1

maxkfranz:
For corrections, you could prioritise the larger (more recent) ID.

Issue #775, opened 2020-07-13
Related paper implementation tweaks


maxkfranz:
Revised:

- (1) **For each element in the factoid:**
  - (a) Make a query to Indra to get statements (associated with papers) -- as today.
  - (b) **Filter down the Indra result to some reasonable N (e.g. 30) -- different from today because it's per-element and it's not chronological.**
  - (c) Get Pubmed results for each pubmed ID -- same as today, but on a per-element list rather than a paper-level list.
  - (d) Send the results to Semantic Search -- same as today.
  - (e) Store the results in `relatedPapers` for the element -- same as today.
- (2) For the entire document: (Note: alternative by taking a slice of each element result, say if the approach below is slow or problematic)
  - (a) **Take all of the results from (1)(c) for all the elements, and put them into a single document-level list.**
  - (b) Send the results to Semantic Search -- same as today.
  - (c) Store the results in `relatedPapers` for the element -- same as today.

Refs.

- Entity-level related papers sorting #763
- No related papers results #742
- Optional 'text': Fetch text by uid https://github.com/PathwayCommons/semantic-search/issues/10

maxkfranz:
@jvwong to investigate Pubmed requests in Semantic Search re. the elimination of step (1)(c)

maxkfranz:
@jvwong @metincansiper I put together a (really) quick and dirty prototype.  It needs more work, but it's a starting point to kick things off.  Review, try it out, and then we can discuss next steps.

@metincansiper, it would be useful if you could review this since it's your code that I changed.  I may have missed something.  There's a lot going on in the Indra code.

@jvwong, it would be useful if you could test this out.  I just ran a watch and resubmitted the same document over and over again by using the admin panel to change the document's status back to 'accepted'.  I think the document-level results need a rework.  The straightforward approach doesn't work -- I suspect -- because of how the Indra code is structured.  It looks like it's just a repeat of the first entity's results.  Document screenshot attached:

<img width="1065" alt="Screen Shot 2020-07-13 at 16 56 23" src="https://user-images.githubusercontent.com/989043/87352891-d2409400-c529-11ea-8a78-05bccb798644.png">

maxkfranz:
@metincansiper @jvwong -- Refer to #777 to discuss the code.  Currently, the document-level results look OK.  

We have a chronological filter by using the PMID as a proxy for the date.  A newer article has a higher-value PMID integer, sufficient for ordinality.

PR #777, opened 2020-07-14
Related paper tweaks 775

maxkfranz:
This PR is for discussion of the code for #775.  It is not intended to be merged now.


src/server/routes/api/document/indra.js
```diff
@@ -167,49 +179,61 @@ const getDocuments = ( templates, queryDoc ) => {
...
+    //   }
+
+    //   return articles;
+    // };
+
+    console.log(`HAVE transform() PMID LIST OF SIZE ${pmids.length}`);
+    console.log(pmids);
+
+    // filter again for case of multiple tempaltes in one query (i.e. document)
+    const filteredPmids = (
```
jvwong, line 226: The place to sort and slice to limit the PMID count to `SEMANTIC_SEARCH_LIMIT` is  just after the merging and just prior to transform: 

https://github.com/PathwayCommons/factoid/blob/7c5cd9c794eaba50e0de3d2d4ff4a71068799641/src/server/routes/api/document/indra.js#L263-L264

In this case, `ints` is a list of `{text, type, pmid}` objects in which only the (text,pmid) tuple is unique.

jvwong:
For the sake of reference, some sample values you'd see (RIG-I interacting with DAPK1) 

https://github.com/PathwayCommons/factoid/blob/7c5cd9c794eaba50e0de3d2d4ff4a71068799641/src/server/routes/api/document/indra.js#L98
```js
[
  { text: "DAPK1 phosphorylates DDX58.", type: "phosphorylation", pmid: "28132841", elId: "5e11faea-ae62-4a61-9583-9be434012227"},
  { text: "DAPK1 inhibits DDX58.", type: "interaction", pmid: "28132841", elId: "5e11faea-ae62-4a61-9583-9be434012227"
  },
   ...,
  {text: "DAPK1 decreases the amount of DDX58.", type: "transcription-translation", pmid: "24531619", elId: "5e11faea-ae62-4a61-9583-9be434012227"}
]
```

jvwong:
After grouping the above: 

https://github.com/PathwayCommons/factoid/blob/7c5cd9c794eaba50e0de3d2d4ff4a71068799641/src/server/routes/api/document/indra.js#L111

```js
{
  "24531619": [{text: "DAPK1 decreases the amount of DDX58.", type: "transcription-translation", pmid: "24531619",
      elId: "5e11faea-ae62-4a61-9583-9be434012227"}
  ],
  "28132841": [
    { text: "DAPK1 phosphorylates DDX58.", type: "phosphorylation", pmid: "28132841", elId: "5e11faea-ae62-4a61-9583-9be434012227"},...
  ]
}

**Comment by jvwong** [comment]
So here's a full description of the example from above:

- Article: [Phosphorylation-Dependent Feedback Inhibition of RIG-I by DAPK1 Identified by Kinome-wide siRNA Screening. 
Mol Cell. 2017 Feb 2;65(3):403-415.e8.](https://pubmed.ncbi.nlm.nih.gov/28132841/)
- Interactions: [DAPK1](https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:2674) phosphorylates [RIG-I](https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:19102)
- Expected related papers for
  - Interaction 'DAPK1 phosphorylates RIG-I': 2 ["24531619", "28132841"] from 2 returned from INDRA
  - Entity 'DAPK1': 30 filtered from 1109 returned from INDRA
  - Entity 'RIG-I': 30 from 3274 returned from INDRA

File: src/server/routes/api/document/indra.js (hunk 1 above)
Comment (maxkfranz) [reply src/server/routes/api/document/indra.js:226]: In practice, a single-filtering approach may not be ideal.  There are two cases, element-level and document-level queries.  

For elements, your approach would result in more or less what we have now.  The only difference would be that the N value would be effectively reduced for queries that have PMID overlap.  In practice, the overlap is small.

For the document, filtering at this stage is necessary but not sufficient.  If you don't filter the subquery results, then the overall document result will be dominated by one of the elements.  If you use chronological ordering, one element tends to have more recent articles than all the others -- and so it dominates.  If you use slicing, the first element always dominates -- this is the behaviour we saw [with the earlier single filter approach](https://github.com/PathwayCommons/factoid/issues/775#issuecomment-657790292).

If anything, I think we should increase the N value for the document-level filter (i.e. the pre-merge filter).  We could even go as far as to remove that filter altogether in order to mitigate this domination effect.

**Comment by maxkfranz** [comment]
> One more thing, the getDocuments merge step does not exclude duplicate values being collected under a given key (object has pmid key and array of {text, type, pmid}), but we never really use that information so its not a big deal.

That is how @metincansiper's code is structured, and I don't want to change it at this point.  Maybe @metincansiper could revisit this structure, but probably not soon given that he said he's busy.  We need to mitigate risk, and so whatever changes we make should be minimally invasive.

**Comment by jvwong** [comment]
Beware: https://github.com/PathwayCommons/semantic-search/issues/16

**Comment by maxkfranz** [comment]
The main case I'm concerned about right now is where an interaction gets a small number of results back (i.e. less than six).  I don't think that @metincansiper's code addressed this case yet, and I don't think that the existing code changes improve that case either.

@jvwong, would you be able to test for one of these cases?  I think you had some bookmarked papers where the interactions were novel.

**Comment by maxkfranz** [comment]
> Beware: [PathwayCommons/semantic-search#16](https://github.com/PathwayCommons/semantic-search/issues/16)

If Metin's [old post-processing](https://github.com/PathwayCommons/factoid/blob/7c5cd9c794eaba50e0de3d2d4ff4a71068799641/src/server/routes/api/document/indra.js#L240-L256) doesn't handle this case, then perhaps it could be filtered out here, just as the query is being formed?

https://github.com/PathwayCommons/factoid/blob/7c5cd9c794eaba50e0de3d2d4ff4a71068799641/src/server/routes/api/document/indra.js#L161-L169

**Comment by jvwong** [comment]
> The main case I'm concerned about right now is where an interaction gets a small number of results back (i.e. less than six). I don't think that @metincansiper's code addressed this case yet, and I don't think that the existing code changes improve that case either.
> 
> @jvwong, would you be able to test for one of these cases? I think you had some bookmarked papers where the interactions were novel.

The example above has 2 for the interaction between DAPK1 and RIG-I. Is that what you mean?

**Comment by maxkfranz** [comment]
> The example above has 2 for the interaction between DAPK1 and RIG-I. Is that what you mean?

Yes.  For cases like those, I suppose we could append document-level results to the interaction.  Even a random selection of the document-level results would probably be pretty good.  All of the final results meet the Semantic Search score cutoff.

**Comment by metincansiper** [comment]
>One more thing, the getDocuments merge step does not exclude duplicate values being collected under a given key (object has pmid key and array of {text, type, pmid}), but we never really use that information so its not a big deal.

Here (https://github.com/PathwayCommons/factoid/blob/7c5cd9c794eaba50e0de3d2d4ff4a71068799641/src/server/routes/api/document/indra.js#L245) we prevent the duplication of same element/interaction in a document/article in our results. Meaning that the elements listed for a document has the uniq combination of text, type, pmid (their pmid's are already supposed to be same since they are under the same document).

Is what you mean something different?

**Comment by metincansiper** [comment]
@maxkfranz @jvwong Sorry for the inconvenience and thanks for working on it. The code updates look good to me. @maxkfranz looks like you choose the alternative way that is mentioned in #775 for the document level (``Note: alternative by taking a slice of each element result, say if the approach below is slow or problematic``).

**Comment by maxkfranz** [comment]
> Is what you mean something different?

I wouldn't worry about this issue right now.  It doesn't affect the results significantly, so it's not necessary to address right now.  It would be nice to address in some later release.

--

> @maxkfranz looks like you choose the alternative way that is mentioned in #775 for the document level (Note: alternative by taking a slice of each element result, say if the approach below is slow or problematic).

I just went with the simplest approach that fit in with your existing code.  It's probably slower, but it's simple and low-risk.

--

@metincansiper, @jvwong:

Would you take a look through the code and test it out?  Jeff, you probably have a better handle than I do on what kinds of papers will trip things up.  Metin, you have a better sense of the code -- so it would be great if you could iron out any new issues that come up from testing.

**Comment by maxkfranz** [comment]
Note: A typical factoid seems to take around 20 to 30 seconds to complete related papers on an old macbook.  If that's the typical speed, then it's fast enough for now.

**Comment by metincansiper** [comment]
>Metin, you have a better sense of the code -- so it would be great if you could iron out any new issues that come up from testing.

I could not find our any major issues. Therefore, I added commits that just cleans up the console logs and the commented out code segments and makes some values configurable as specified by TODO comments.

I only have found out the following 2 minor things considering the code:

- Here (https://github.com/PathwayCommons/factoid/pull/777/files#diff-255cd77344ec16f0add2bf14e14f4f87R103) is the code segment that makes the filtering per factoid element. In that line the related interactions are filtered. If our aim is to filter the pmid's as we do for the overal filtering step then it may not hold. I mean the number of pmids may be less then then limit after grouping the filtered interations by pmid (https://github.com/PathwayCommons/factoid/pull/777/files#diff-255cd77344ec16f0add2bf14e14f4f87R108) if there are some of them sharing the same pmid. I assume that it may not be a big deal (especially if the pmids are various enough) but I wanted to mention.
- In both of the cases (https://github.com/PathwayCommons/factoid/pull/777/files#diff-255cd77344ec16f0add2bf14e14f4f87R105 and https://github.com/PathwayCommons/factoid/pull/777/files#diff-255cd77344ec16f0add2bf14e14f4f87R182) the number of filtered elements is one more of the semantic search limit. Is there a reason for that +1?

File: src/server/routes/api/document/indra.js (hunk 1 above)
Comment (jvwong) [reply src/server/routes/api/document/indra.js:180]: The sort/slice issue I raised above still stands: If we want to filter here, we have to remove duplicate `intns` elements ( wrt pmid) prior to the slice -

https://github.com/PathwayCommons/factoid/blob/14cab00260bc0271961eaf933d334b31bb5b425e/src/server/routes/api/document/indra.js#L103-L105

For instance, with `DAPK1 interaction with RIG-I` these are the per-element results:
  - DAPK1
    - INDRA: 384 PMIDs
    - Related papers
      - Actual: 9
      - Expected: 30 (`SEMANTIC_SEARCH_LIMIT`) 
  - RIG-I
    - INDRA: 1480 unique PMIDs
    - Related papers
      - Actual: 12
      - Expected: 30 (`SEMANTIC_SEARCH_LIMIT`) 
  - DAPK1 Phosphorylates RIG-I
    - INDRA: 2 unique PMIDs
    - Related papers
      - Actual: 2
      - Expected: 2

**Comment by jvwong** [comment]
> > One more thing, the getDocuments merge step does not exclude duplicate values being collected under a given key (object has pmid key and array of {text, type, pmid}), but we never really use that information so its not a big deal.
> 
> Here (
> 
> https://github.com/PathwayCommons/factoid/blob/7c5cd9c794eaba50e0de3d2d4ff4a71068799641/src/server/routes/api/document/indra.js#L245
> ) we prevent the duplication of same element/interaction in a document/article in our results. Meaning that the elements listed for a document has the uniq combination of text, type, pmid (their pmid's are already supposed to be same since they are under the same document).
> 
> Is what you mean something different?

Yes this addresses the issue.

**Comment by maxkfranz** [comment]
Re. n+1, I didn’t bother to look up the semantics of the slice parameters (index versus length etc.). n+1 works in any case, and the exactness of the n value isn’t very important. 

We could use something like _.take(arr, n) to make it more accurate and readable

**Comment by maxkfranz** [comment]
You’re right, but does this affect the main goal significantly at this time?  The goal is to have six good quality results. Unless there is a near worst case with a net result less than six, the user isn’t negatively affected. 

If Metin is able to address this filtering in a more sophisticated way, that would be great. If not, an alternative would be to increase the N value for that specific filter: The net effect would be similar in most cases, though technically not quite as robust. 

Generally, we can afford to increase the N values a bit — say from 30 to 50. That would be good to try even if we also take the route of robust filtering. Ultimately, the N value has the most significant effect on the quality of the top six.

**Comment by jvwong** [comment]
> You’re right, but does this affect the main goal significantly at this time? The goal is to have six good quality results. Unless there is a near worst case with a net result less than six, the user isn’t negatively affected. If Metin is able to address this filtering in a more sophisticated way, that would be great. If not, an alternative would be to increase the N value for that specific filter: The net effect would be similar in most cases, though technically not quite as robust. Generally, we can afford to increase the N values a bit — say from 30 to 50. That would be good to try even if we also take the route of robust filtering. Ultimately, the N value has the most significant effect on the quality of the top six.
> […](#)
> On Jul 17, 2020, at 11:15, Jeffrey ***@***.***> wrote: ﻿ @jvwong commented on this pull request. In src/server/routes/api/document/indra.js: > + // const cmp = ( a, b ) => { + // const getDate = e => getPubTime( getMedlineArticle( e ) ); + // return getDate( a ) - getDate( b ); + // }; + + // articles = Heap.nlargest( articles, SEMANTIC_SEARCH_LIMIT, cmp ); + // } + + // return articles; + // }; + + console.log(`HAVE transform() PMID LIST OF SIZE ${pmids.length}`); + console.log(pmids); + + // filter again for case of multiple tempaltes in one query (i.e. document) + const filteredPmids = ( The sort/slice issue I raised above still stands: If we want to filter here, we have to remove duplicate intns elements ( wrt pmid) prior to the slice - https://github.com/PathwayCommons/factoid/blob/14cab00260bc0271961eaf933d334b31bb5b425e/src/server/routes/api/document/indra.js#L103-L105 For instance, with DAPK1 interaction with RIG-I these are the per-element results: DAPK1 INDRA: 384 PMIDs Related papers Actual: 9 Expected: 30 (SEMANTIC_SEARCH_LIMIT) RIG-I INDRA: 1480 unique PMIDs Related papers Actual: 12 Expected: 30 (SEMANTIC_SEARCH_LIMIT) DAPK1 Phosphorylates RIG-I INDRA: 2 unique PMIDs Related papers Actual: 2 Expected: 2 — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

I just pushed a fix for it. It is trivial

**Comment by maxkfranz** [comment]
> I just pushed a fix for it. It is trivial

Great.  I think I confused it with the section that needs all the map lookups; this is just the flat array

File: src/config.js [hunk 2]
```diff
@@ -54,7 +54,10 @@ export const DB_CERT = env('DB_CERT', undefined);  // path to a certificate (cer
 export const PC_URL = env('PC_URL', 'https://apps.pathwaycommons.org/');
 export const REACH_URL = env('REACH_URL', 'http://reach.baderlab.org/api/uploadFile');
 export const BIOPAX_CONVERTER_URL = env('BIOPAX_CONVERTER_URL', 'https://biopax.baderlab.org/convert/v2/');
-export const GROUNDING_SEARCH_BASE_URL = env('GROUNDING_SEARCH_BASE_URL', 'https://grounding.baderlab.org');
+//
+// TODO PUT THIS BACK TO THE PROPER VALUE -- master only for convenient testing
+export const GROUNDING_SEARCH_BASE_URL = env('GROUNDING_SEARCH_BASE_URL', 'https://master.grounding.baderlab.org');
```
maxkfranz, line 59: This should be changed back before merging

metincansiper: @maxkfranz I did so and comitted . However, the build has been failed so I have not been merged this PR yet. The build error logs are as the following:

```
Failed to install 10. Remote repository may not be reachable.
Using locally available version 10, if applicable.
$ nvm use 10
N/A: version "10 -> N/A" is not yet installed.
You need to run "nvm install 10" to install it before using it.
Unable to use 10
The command "false" failed and exited with 1 during .
Your build has been stopped.
```

maxkfranz:
I think we need a minimum of node v10. I use v12. I think the build should work if you upgrade.  If nvm doesn’t work, you can use the regular installer from the node website

The docker uses v10

metincansiper:
I was meaning the docker build. I just restarted it and it worked well this time. Therefore, I just merged this PR.

PR #783, opened 2020-07-22
AppSignal

maxkfranz:
This just adds basic support for AppSignal, which should track and notify re. stability and performance.  This should allow for a greater level of proactivity w.r.t. stability.  If this AppSignal integration proves successful for this project, then it could be integrated into other projects (esp. Cytoscape cloud projects).

Note: This is just the basic integration copy-pasted from their wizard.  There are other features that we could leverage, but I just wanted to springboard things.  The key should be documented in the private repo.

@jvwong, I've added you to the AppSignal organisation.  Could you add Jeff Liu?  I don't have his email address, and he should have access to the dashboard and notifications since he handles IT.

Thanks to the guys at AppSignal for setting us up!


maxkfranz:
Without the key, it should just be disabled.

Issue #787, opened 2020-07-28
Post submission popover has overflowing URL


maxkfranz:
<img width="515" alt="Screen Shot 2020-07-28 at 14 18 23" src="https://user-images.githubusercontent.com/989043/88705254-43ad4480-d0dd-11ea-8a53-fc70fd2fb8b1.png">

jvwong:
Two more things I noticed:

1. The link is gone in Safari (Version 13.1.1 (15609.2.9.1.2))
<img width="330" alt="image" src="https://user-images.githubusercontent.com/4706307/88818913-48313600-d18d-11ea-860f-93214965657d.png">

2. The horizontal bar is shifted

jvwong:
What browser are you using for the initial issue image? I can't reproduce that.

maxkfranz:
Chrome

PR #789, opened 2020-07-29
Submit success: Wrapping url and text

jvwong:
This one should do the trick with respect to wrapping long text in the submit success tooltip. Also fixed the horizontal divider.

I removed a bug where the explore link was only shown contingent on the share capability being present. Odd.

Refs #787


jvwong:
Firefox (78.0.2 (64-bit):
image

Chrome (Version 84.0.4147.89 (Official Build) (64-bit))
image

Safari (Version 13.1.1 (15609.2.9.1.2)):

image

Issue #801, opened 2020-08-13
Related papers: Seed articles


jvwong:
#### Background

A few issues surrounding the related papers, both technically and from a quality standpoint:

- Articles retrieved based upon querying with for a entity (gene, chemical), are frankly less than relevant #782 #784
- New papers, generally with novel interactions, will be padded with these entity-based papers
- Desire to provide/display related papers ASAP
- INDRA can fail, with no strategy for recovery
- What a 'document'-level related paper is, is unclear; currently a reshuffling of network-based papers

#### Requirements

Ideally, we'd like to have a set of 'seed' related articles that are relevant to the paper, from a relatively stable service, that can also be used to add-upon with subsequent queries for related papers (i.e. INDRA), and are available right away. 

#### Proposed

When an author inputs their article title, we populate the document with article information. The suggestion is to immediately retrieve pmids linked to the paper via [ELINK](https://dataguide.nlm.nih.gov/eutilities/utilities.html#elink):

- article types
  - related
  - papers citing (incoming)
  - references (outgoing
  - reviews
  - also viewed

Notes: Can configure to return date range (published less than 100 days ago) or any subset (citing this article). An article from Mol Cell published 6 days ago [returned 130 pmids](https://eutils.ncbi.nlm.nih.gov/entrez/eutils/elink.fcgi?id=32755595&retmode=json&db=pubmed&dbfrom=pubmed).  

PS: The service also provides the ability to link from 'UID' (gene/protein) to articles, that is, can be used to seed the entity-level papers.

maxkfranz:
> Desire to provide/display related papers ASAP

It would be great to get Elink related papers shown without having to wait for Indra results.  We may have several phases:

(1) Elink results are just appended to the regular workflow.  This covers Indra failures and requires minimal structural changes.

(2) Elink results are queried and sorted independently from Indra results.  This allows for showing related papers quicker after submission.  This requires more structural changes, and there are UX concerns (i.e. the list refreshes twice).

For now, it may be more prudent to just focus on phase (1).

jvwong:
Task plan:

- [ ] Implement ELINK service wrapper(s)
  - link from
    - db=gene uid(s)
    - db=pubmed uid(s)
- [ ] Update related papers retrieval
  - TODO - need to consider appending pmids with INDRA data

Issue #785, opened 2020-07-27
Related Papers: Improving robustness by providing fallback and recovering upon initial failure  


jvwong:
Background: The related papers feature relies upon external (third-party) services to obtain information about PubMed articles used to populate the Factoid document and subsequently the explorer view (UI).

Issue:  The article information may not be available at the time of request, which in this case, follows immediately after document submission through the editor. The relevant document data will not be updated and there is no further attempt to retrieve this information. Another side-effect is that explorer view displays a persistent 'waiting'-type message.  

- Requirements
  - Provide a fallback when the system is unable to retrieve articles externally
    - At least remove the 'infinite spinner' 
    - Could consider informative note or even back-up data for users
  - Attempt to retrieve articles for documents where a related-paper update has not been achieved
    - My experience is that INDRA retrieval failures can persist for prolonged periods (minutes)
    - We do have a cron service that we could make use of 

Refs #780, feedback from @cannin

maxkfranz:
Here's a document that fails repeatedly: https://unstable.factoid.baderlab.org/document/a6051ebc-216b-4513-b19a-d29f266cee8e

maxkfranz:
Non-human text queries seem to be more fragile, but sometimes they work if you retry soon after the failure (i.e. within seconds).  Something like this may be a simple mitigation for this case: https://www.npmjs.com/package/fetch-retry

For longer error durations, the same approach could be used with a slightly different strategy (e.g. try three times quickly, then try again every few minutes).

We may also reconsider using a different, more reliable version of Indra.

maxkfranz:
@metincansiper, would you be able to evaluate the fetch-retry approach with your Indra queries?

jvwong:
> Here's a document that fails repeatedly: https://unstable.factoid.baderlab.org/document/a6051ebc-216b-4513-b19a-d29f266cee8e

I can reproduce locally. Getting a lot of these in response from INDRA: 

```json
{"message":"Endpoint request timed out"}
```

Also, as I recall, Promise.all is an 'all or none' with regards to success/error.

jvwong:
For the above pathway, once I hit submit with anything below 4 interactions I got successful results back. Above 3 interactions, it gives me that error and the whole thing dies.

maxkfranz:
(1) It can happen even with a single interaction. Larger pathways would then increase the probability of overall failure, because there are more chances for individual failure.

(2) There is also the possibility of failure due to load. 

[...]

PR #805, opened 2020-08-18
Retrieve doc-level related papers via ELINK

jvwong:
- Wrapper module for EUTILS ELINK service
  - Limit pmids retrieved to 50/subset, so on the order of 100's returned in worst case
- Expose the semantic-search from indra.js module
- Set document related papers via ELINK
  - Retrieval occurs just following user signup, article retrieval (takes ~5s)
  - ranked via semantic-search
  - Updated via cron
  - Retrieve and backfill network-based related papers, as before

Refs #801 , #785


maxkfranz:
Looks good.

Is the intention that the Elink results are the only results for the document-level related papers?  Or could it be that Elink results were just better (according to Semantic Search) for the papers I tried?

jvwong:
> Is the intention that the Elink results are the only results for the document-level related papers? Or could it be that Elink results were just better (according to Semantic Search) for the papers I tried?

Doc-level papers are exclusively from ELINK. My experience is that INDRA's strengths lie at the interaction-level.

maxkfranz:
Yeah, makes sense.  Just wanted to be sure of the details.  I think it’s good to merge

PR #812, opened 2020-08-24
Release v0.9.1

jvwong:
- Patches
  - Puppeteer updated their Docker installation instructions
  - Bug in entity BioPAX export template


jvwong:
I will note that this puppeteer software is temperamental, wouldn't work in Docker (some cy error), then after reboot does works fine.

maxkfranz:
> I will note that this puppeteer software is temperamental, wouldn't work in Docker (some cy error), then after reboot does works fine.

On Mac?

jvwong:
I see this happening from time to  time on the unstable (Ubuntu) and today it happened on Docker (Ubuntu) with the prod instance. It's fine now.

jvwong:
its only at bootup

maxkfranz:
The cluster isn't very fast, and it can't generate anything until the chrome instance has started up (which is heavy).

jvwong:
Oh ok maybe I have to wait a bit.

jvwong:
FYI if you want to 'see' an SBGN of a document: https://apps.pathwaycommons.org/factoids/

PR #817, opened 2020-08-24
Use fetch-retry for the indra queries.

metincansiper:
Refs #785

### DIFF
--- a/src/server/routes/api/document/indra.js
+++ b/src/server/routes/api/document/indra.js
@@ -17,6 +17,9 @@ import { getPubmedCitation } from '../../../../util/pubmed';
 const INDRA_STATEMENTS_URL = INDRA_DB_BASE_URL + 'statements/from_agents';
 const SORT_BY_DATE = false; // TODO remove
 
+const FETCH_RETRIES = 3;
+const FETCH_RETRY_DELAY = 3000;
+
 const SUB_MODIFICATION_TYPES = ['Phosphorylation', 'Dephosphorylation', 'Dephosphorylation',
   'Deubiquitination', 'Methylation', 'Demethylation'];
 const BASE_MODIFICATION_TYPES = ['Sumoylation', 'Desumoylation', 'Hydroxylation', 'Dehydroxylation',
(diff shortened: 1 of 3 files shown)


src/server/routes/api/document/indra.js
```diff
@@ -31,6 +32,12 @@ const ENTITY_STR = 'entity';
 
 let sortByDate = SORT_BY_DATE;
 
+const fetchRetry = FetchRetry(fetch);
+const fetchRetryUrl = url => {
```
maxkfranz, line 36: Missing options: `fetchRetryUrl = (url, options) ...`

maxkfranz: Affects any non-get request

metincansiper:
Thanks for noticing it! I updated the PR so that ``fetch`` options are considered as Max mentioned.

jvwong:
What I noticed before is that (node-)fetch only errors out under very specific circumstances: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/ meaning it won't catch any HTTP errors. I was testing by using `INDRA_DB_BASE_URL=https://httpstatuses.com/500`

To catch other errors, you gotta use the `retryOn` opt for `fetch-retry`, but it wasn't obvious how to configure that. To make a long story short you can update the code like: 

```js
...
const FETCH_RETRIES = 3;
const FETCH_RETRY_DELAY = 3000;
const fetchRetry = FetchRetry(fetch);
const fetchRetryUrl = ( url, opts ) => {
  let retryOpts = {
    retryDelay: () => FETCH_RETRY_DELAY,
    retryOn: function( attempt, error, response ) {
      // retry on any network error, or 4xx or 5xx status codes
      const { statusText, status, ok } = response;
      if ( attempt < FETCH_RETRIES && !ok ) {
        logger.error(`Error for ${url}`);
        logger.error(`${status}: ${statusText}`);
        logger.info(`Retrying, attempt ${attempt + 1}`);
        return true;
      } else {
        return false;
      }
    }
  };

  return fetchRetry(url, _.extend( opts, retryOpts ));
};

...
```

metincansiper:
@jvwong okay I pushed the code change that you mentioned.

PR #819, opened 2020-08-28
Export biopax zip

metincansiper:
Added an api end point to download a zip file that contains all the biopax files in Factoid. I expect to use this end point for adding factoid to cpath2.


metincansiper:
>I tend to use the Swagger nowadays, so could add the /zip/biopax

Okay, I added the swagger docs and merging the PR as everyone has approved it.

PR #832, opened 2020-09-21
Homepage fixes for 0.10.1

maxkfranz:
- Text updates
- Carousel has 20 doc limit & corrected service request
- Updated raf video -- consistent with 'researchers' section
- Visual improvements to carousel
- Better full-screen sizing for above-fold section
- Removes commented-out code


maxkfranz:
@jvwong, I'm not sure if this is your latest banner.  You may want to drop in a newer version in this branch.

jvwong (approved):
I dropped in the banner and addded 'SARS-CoV-2' to popup.

maxkfranz:
I think we may have two interactions popups now.  Does that cover both?

jvwong:
> I think we may have two interactions popups now. Does that cover both?

OK missed the second one. I think there's should be a 'Should I use Biofactoid?' sort of popover in the future.

Issue #839, opened 2020-09-23
Carousel: Only 'published' docs 


jvwong:
A document is 'published' when it is submitted, has at least one complete element.

Carousel should show only published docs.

e.g. Someone submitted a bunch of empty docs.

jvwong:
Also, unedited documents have a lastEdited date displayed as 'Over 50 years ago'.

PR #840, opened 2020-09-23
Carousel: Only published

jvwong:
Also dropped timestamp if unedited.

Refs #839

### DIFF
--- a/src/client/components/carousel.js
+++ b/src/client/components/carousel.js
@@ -171,7 +171,7 @@ export class Carousel extends Component {
             abstractDiv,
             h('div.carousel-doc-footer', [
               h('div.carousel-doc-text', doc.text),
-              h('div.carousel-doc-datestamp', doc.lastEditedDate ? formatDistanceToNow( new Date( doc.lastEditedDate ), { addSuffix: true } ) : null )
+              h('div.carousel-doc-datestamp', formatDistanceToNow( new Date( doc.lastEditedDate || 0 ), { addSuffix: true } ))
             ]),
           ]),
           figureDiv,
--- a/src/client/components/carousel.js
+++ b/src/client/components/carousel.js
@@ -23,12 +23,12 @@ export class Carousel extends Component {
       refresh: props.refresh || (() => { // get all docs from service by default
         const url = `/api/document`;
         const params = queryString.stringify({
-          status: ['published', 'submitted'].join(','),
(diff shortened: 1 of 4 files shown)


src/client/components/carousel.js
```diff
@@ -23,12 +23,12 @@ export class Carousel extends Component {
       refresh: props.refresh || (() => { // get all docs from service by default
         const url = `/api/document`;
         const params = queryString.stringify({
-          status: ['published', 'submitted'].join(','),
```
maxkfranz, line 26: What is the current distinction between 'published' and 'submitted'?  It seems that the allowed statuses here should be the same as the allowed statuses for the download zip.

If there is no distinction between 'submitted' and 'published' anymore, then 'submitted' should be removed everywhere.

src/client/components/carousel.js
```diff
@@ -171,7 +171,7 @@ export class Carousel extends Component {
             abstractDiv,
             h('div.carousel-doc-footer', [
               h('div.carousel-doc-text', doc.text),
-              h('div.carousel-doc-datestamp', formatDistanceToNow( new Date( doc.lastEditedDate || 0 ), { addSuffix: true } ))
```
maxkfranz, line 174: Shouldn't that data exist for all allowable statuses?  If it doesn't exist, then the data ought to be amended.

src/client/components/carousel.js
jvwong: The 'published' status is a shortcut to say that  the document meets some minimum standard: Contains at least one element and nothing is incomplete (e.g. un-grounded node).

src/client/components/carousel.js
jvwong: Maybe initialize the document 'lastEditedDate' to `createdDate`?

maxkfranz: That would be a reasonable fallback

src/client/components/carousel.js
maxkfranz: OK, then it sounds like the zip should only include 'published', whereas it includes 'submitted' right now (see https://github.com/PathwayCommons/factoid/pull/831).

Could that consistency be included in this PR?  This status issue has come up several times now, and it would be great to resolve it completely.

jvwong: OK will do

Issue #843, opened 2020-09-25
Document status: Refine; Use Enumerations in document model 


jvwong:
## Background

These flags have evolved over time as we began to understand our user workflow and which ones were necessary to quickly categorize/filter documents. For instance,  we needed a way to easily identify those to export to BioPAX and display on the homepage carousel.

The document model provides an enum `status`: 
  - `requested`: An author has submitted a request to edit the document
  - `approved`:  Biofactoid-team approves the document for editing
  - `submitted`: The document to be complete
  - `published`: The document has >=1 complete element 
  - `trashed`: This document should not be used further 

## Updates:

The `status` value should be more meaningful, attempt to describe some specific aspect of the document itself

- 1. `requested` and `approved` are not relevant any longer and redundant. 
  - Suggest to collapse into `initiated`, more or less to say its none of the other  
- 2. Change `published`
  - Change this to `public`, i.e. to export and show on homepage, Tweet etc 
- 3. Reference these values via the static function in document model

PR #844, opened 2020-09-28
Update document status fields.

jvwong:
- Available status fields, which hopefully are self-evident
  - `INITIATED`
  - `SUBMITTED`
  - `PUBLIC`
  - `TRASHED`

Refs #843

### DIFF
--- a/src/client/components/carousel.js
+++ b/src/client/components/carousel.js
@@ -5,6 +5,8 @@ import _ from 'lodash';
 import { formatDistanceToNow } from 'date-fns';
 import queryString from 'query-string';
 
+import Document from '../../model/document';
+
 export const CAROUSEL_CONTENT = {
   FIGURE: 'figure',
   ABSTRACT: 'abstract'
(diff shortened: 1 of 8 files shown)


maxkfranz (commented):
Looks good.  Two comments:

(1) Are you planning on converting all the existing documents manually with the new status?  Or should this be handled with an automatic conversion in the code?

(2) The naming of `isPublic()` is a bit inconsistent with the other status methods.  I suppose `public()` and `publicize()` would line up with your other methods, but this isn't very important.

jvwong:
> Looks good. Two comments:
> 
> (1) Are you planning on converting all the existing documents manually with the new status? Or should this be handled with an automatic conversion in the code?

At this point I can migrate this easily with such a simple change. My approach has been to do it 'offline' (local) then load it into an updated instance

> 
> (2) The naming of `isPublic()` is a bit inconsistent with the other status methods. I suppose `public()` and `publicize()` would line up with your other methods, but this isn't very important.

Yeah I just went with the most unambiguous. Didn't feel like `public` sounded like a getter

maxkfranz (approved):
Fair enough.  Let's just merge it in

jvwong:
I'm gonna clear out the db at some point ...

Issue #842, opened 2020-09-24
Improve visibility via search (Google)


jvwong:
## Description

**Q: What is the name of the feature?**

A: Search engine optimization

**Q: What does this feature enable the user to do?**

A: Find and discover documents in search engines (text, image, snippets)

**Q: What information must the user provide to use the feature?**

A: Keywords, article title

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: Need to explicitly provide URLs for each document

## Specification

Will the following work?

- `/`
  - Provide a robots.txt
    - [Intro to robots.txt](https://support.google.com/webmasters/answer/6062608?hl=en)
    - [Google reference](https://developers.google.com/search/reference/robots_txt)
    - Provide a site map with list of (read-only) published document URLs
      - [Google site map reference](https://support.google.com/webmasters/answer/183668?hl=en)
  - Down the road, can simply have an archive page, much like any journal 
- `/document/{id}`
  - We already do this for Twitter cards so same idea
  - various meta tags
- Google is already indexing the documents, probably from Twitter: https://www.google.com/search?q=site%3Abiofactoid.org

### Details

//TODO

- Routing
  - add `/robots.txt`

- [`<meta>` tags](https://support.google.com/webmasters/answer/35624?visit_id=636880082951453933-4272044119&rd=1)
    - [Special tags that Google understands](https://support.google.com/webmasters/answer/79812?hl=en)
    - <head><title>
    - <meta> description
    - <meta> author
    - <meta> robots
    
- Structured data
  - [Article](https://developers.google.com/search/docs/data-types/article#non-amp) rich snippets

- Support search bar via `site:biofactoid.org ras interacts with raf`

maxkfranz:
Example rich snippet for https://biofactoid.org/document/e1e4b079-258d-4258-b5a2-3a665046d880

```
<!-- JSON-LD markup generated by Google Structured Data Markup Helper. -->
<script type="application/ld+json">
{
  "@context" : "http://schema.org",
  "@type" : "Article",
  "name" : "p27 controls Ragulator and mTOR activity in amino acid-deprived cells to regulate the autophagy-lysosomal pathway and coordinate cell cycle and cell growth.",
  "author" : {
    "@type" : "Person",
    "name" : "Ada NowosadPauline JeannotCaroline CallotJustine CreffRenaud Thierry PercheyCarine JoffrePatrice CodognoStephane ManentiArnaud Besson"
  },
[...]

PR #846, opened 2020-09-29
Generate sitemap

jvwong:
- Module: `sitemap`  exposing `generateSitemap( docs )`:
  - assembles a JSON version of the sitemap data
  - uses `xml-js` to generate the xml string
  - writes to static folder
  - Tests for sensible xml
- web service API: `/document/sitemap?apiKey=xyz`
  - filter for 'published' documents (see #844)
  - uses `generateSitemap`
  - apiKey protected
  -  Swagger docs

Sample out

```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
  <url>
    <loc>https://biofactoid.org/document/7af64ad2-64a5-466e-b42d-acf259c3883d</loc>
    <lastmod>2020-09-28T20:19:27.063Z</lastmod>
    <changefreq>weekly</changefreq>
    <priority>0.5</priority>
  </url>
  <url>
    <loc>https://biofactoid.org/document/13a08fd5-0299-47c3-be94-0efc0f878a90</loc>
    <lastmod>2020-09-28T20:18:18.196Z</lastmod>
    <changefreq>weekly</changefreq>
    <priority>0.5</priority>
  </url>
</urlset>
```

Refs #842 

- Notes:
  - [sitemaps.org](https://www.sitemaps.org/protocol.html)
  - Should sync status filter with #844
  - Maybe create on boot?

### DIFF
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -44,7 +44,7 @@ import { BASE_URL,
 import { ENTITY_TYPE } from '../../../../model/element/entity-type';
 import { db2pubmed } from './pubmed/linkPubmed';
 import { fetchPubmed } from './pubmed/fetchPubmed';
-import { generateSitemap } from '../../../sitemap';
+import { docs2Sitemap } from '../../../sitemap';
 const DOCUMENT_STATUS_FIELDS = Document.statusFields();
 
 const http = Express.Router();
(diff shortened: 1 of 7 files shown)


maxkfranz:
Would it be simpler for the sitemap to be generated dynamically?  It's always up-to-date that way.

jvwong:
> Would it be simpler for the sitemap to be generated dynamically? It's always up-to-date that way.

I think I misread your 'attach it to a route' comment. 

Maybe you mean when a document becomes `PUBLIC`, regenerate the sitemap?

maxkfranz:
I meant that it's a dynamic route rather than a static file.  So something like `/sitemap.xml` is just a dynamic route like `/api/documents` that returns all the current documents.

jvwong:
> I meant that it's a dynamic route rather than a static file. So something like `/sitemap.xml` is just a dynamic route like `/api/documents` that returns all the current documents.

Oh ok that's a better idea.

maxkfranz:
The xml format may be OK, but you can also just use a simple text file with each url separated by newlines, e.g.

```
https://biofactoid.org/document/foo
https://biofactoid.org/document/bar
https://biofactoid.org/document/baz
```

Maybe the xml format has some other advantages, but the text format is really simple.

maxkfranz:
The robots.txt should be included in this PR to activate the site map.  See https://github.com/PathwayCommons/factoid/issues/842#issuecomment-701561771

I think that this should be enough to start:

```
User-agent: *
Allow: /

Sitemap: https://biofactoid.org/api/sitemap.xml
```

maxkfranz:
Note that the sitemap should also include the homepage: `https://biofactoid.org/`

src/server/routes/api/document/index.js
```diff
@@ -319,6 +320,44 @@ let getSbgnFromTemplates = templates => {
...
+    })
+    .then( cursor => cursor.toArray() )
+    .then( res => {
+      return Promise.all( res.map( docDbJson => {
+        let { id } = docDbJson;
+        let docOpts = _.assign( {}, tables, { id } );
+        return loadDoc( docOpts ).then( getDocJson );
+      }));
+    } )
+    .then( docs2Sitemap )
```
maxkfranz, line 357: Maybe append the home page here?

jvwong: It's already appended in the sitemap.js code. Isn't that OK?

Issue #848, opened 2020-09-30
Identifying author's article: Additional constraints for PubMed search results


jvwong:
When an author enters their article title, the system fetches search hits from PubMed EUTILS and returns the first hit for which the input is contained in the search hit's title. 

An edge case is published corrections, which contain the full title. Example: `ROCK-mediated selective activation of PERK signalling causes fibroblast reprogramming and tumour progression through a CRELD2-dependent mechanism` which returns the top hit `Publisher Correction: ROCK-mediated selective activation of PERK signalling causes fibroblast reprogramming and tumour progression through a CRELD2-dependent mechanism`

Idea: Filter raw EUTILS ESEARCH results

The raw article data contains a [PublicationType](https://dtd.nlm.nih.gov/ncbi/pubmed/doc/out/180101/el-PublicationType.html) attribute which we can use to include/exclude results. 

There is a huge list of possible PublicationTypes listed here: https://www.nlm.nih.gov/mesh/pubtypes.html so I'm more inclined to just exclude certain ones that seem problematic, i.e. ` <PublicationType UI="D016425">Published Erratum</PublicationType>`

Refs #845

jvwong:
Not sure why this is still happening, this case in particular:

Input: [DNA-guided genome editing using the Natronobacterium gregoryi Argonaute. Nat Biotechnol. 2016 Jul;34(7):768-73.
doi: 10.1038/nbt.3547. Epub 2016 May 2.  ](https://pubmed.ncbi.nlm.nih.gov/27136078/)

Actual: [Retraction: DNA-guided genome editing using the Natronobacterium gregoryi Argonaute. Nat Biotechnol. 2017 Aug 8;35(8):797. doi: 10.1038/nbt0817-797a. ](https://pubmed.ncbi.nlm.nih.gov/28787423/)

See: https://biofactoid.org/document/74f186e6-b6c9-4938-a56a-b4dc37908782 (which I manually corrected)

jvwong:
Solution: Add `<PublicationType UI="D016440">Retraction of Publication</PublicationType>` to exclude list

PR #849, opened 2020-10-01
Capture the Pubmed article type 

jvwong:
- Store article PublicationType information from EUTILS EFETCH
- Use this info to exclude search hits of type `Published Erratum` for an author's paper

Trade-off is authors won't be able to add an erratum 

Refs #848


maxkfranz (approved):
Looks good

Issue #851, opened 2020-10-02
Document post-mortem: Missing metadata and misc problems 


jvwong:
PubMed: [Coordinated Conformational Processing of the Tumor Suppressor Protein p53 by the Hsp70 and Hsp90 Chaperone Machineries](https://pubmed.ncbi.nlm.nih.gov/31027879/)
Document URL: https://biofactoid.org/document/e3ad1fe0-7cc9-4075-b9a7-32626ce0c2ff

Problems

- 1. Biofactoid could not identify their article based on author input
  -  Author added variant of the PubMed indexed title `Coordinated conformational processing of p53 by the Hsp70 and Hsp90 chaperones machineries`  
  - It's worth noting that EUTILS ESEARCH returns nothing;  PubMed website returns the correct article
  
- 2. No related papers 
  - Document: No PubMedID 
  - Network: INDRA responded with `504: Gateway Timeout`  

- 3. Display in carousel is different than explorer 
  - Carousel:

<img src="https://user-images.githubusercontent.com/4706307/94934103-48092b80-0499-11eb-8b44-f38c82c0f0d5.png" width="200px" />
 
- explorer
<img src="https://user-images.githubusercontent.com/4706307/94934239-7981f700-0499-11eb-8bf7-9e0ebc7ce038.png"  width="200px" />

maxkfranz:
Re. (iii): Did the user submit early, or might this have been related to the previous status values?

Re. (ii): Is this a one-off Indra issue for this document (https://biofactoid.org/document/a6b4351d-b474-49d9-bc3d-f044513d5a0c)?  Could Metin add more retries to the Indra queries to resolve this?  Or is this a reproducible error for the document specifically?

Another issue (iv): Is there the wrong grounding for this document https://biofactoid.org/document/aa18ee18-a78f-4d46-a244-9a6dc97bb238 ?  'Cofilin' doesn't exist for mouse in NCBI.  Only 'cofilin-1' exists for mouse as the closest synonym.

jvwong:
> Re. (iii): Did the user submit early, or might this have been related to the previous status values?

Likely submitted followed by editing. I told one user that didn't like their submission that they could still edit the document via the same link.

> Re. (ii): Is this a one-off Indra issue for this document (https://biofactoid.org/document/a6b4351d-b474-49d9-bc3d-f044513d5a0c)? Could Metin add more retries to the Indra queries to resolve this? Or is this a reproducible error for the document specifically?

[...]

Issue #845, opened 2020-09-29
Real users: Article metadata retrieval failures


jvwong:
Just logging some failures I noticed with real user input

- 1. Error: `Unable to retrieve a PubmedArticle` / ` Error filling doc article`
  - Date: 2020-09-29T14:39:24.221050250Z
  - Author input: `Human insulin growth factor 2 mRNA binding protein 2 increases miR-33a/b inhibition of liver ABCA1 expression and alters low-density apolipoprotein levels in mice`
  - Possible reason: Title in PubMed is `...Increases MicroRNA 33a/b...` 

- 2. Error: `getRelPprsForDoc: Bad Request (400)`
  - Date:  Tue Sep 29 2020 13:47:15 GMT+00:00
  - Author input: `Atrx Deletion in Neurons Leads to Sexually Dimorphic Dysregulation of miR-137 and Spatial Learning and Memory Deficits` [https://pubmed.ncbi.nlm.nih.gov/32610139/]
  - Possible reason: Something in semanti-search (`"value is not a valid dict"`)

- 3. Error: `HTTPStatusError: Bad Request (400)` / ` Error filling doc article`
  - Date: 2020-09-29T17:11:02.430147444Z
  - Author input: 'Critical roles of phosphoinositides and NF2 in Hippo pathway regulation'
  - Possible reason: I refreshed the data and it worked???

- 4. Problem: System kept giving me an 'Author correction: ..' entry
  - Date: 2020-09-29
  - Author input: `ROCK-mediated selective activation of PERK signalling causes fibroblast reprogramming and tumour progression through a CRELD2-dependent mechanism` [https://pubmed.ncbi.nlm.nih.gov/32451439/]
  - Problem: Exact input to title match isn't prioritised, only 'contains'  

- 5. Error: Unable to retrieve a PubmedArticle
  - Date: 2020-10-02T11:02:50.823177975Z
  - Author put in `Coordinated conformational processing of p53 by the Hsp70 and Hsp90 chaperones machineries` for `Coordinated Conformational Processing of the Tumor Suppressor Protein p53 by the Hsp70 and Hsp90 Chaperone Machineries` [https://pubmed.ncbi.nlm.nih.gov/31027879/]
  - Possible reason: They're playing fast and lose as if it's a search engine?

- 6. Unable to retrieve a PubmedArticle
  - Date: 1602119209.78
[...]

PR #865, opened 2020-10-08
More aggressive Indra retries

maxkfranz:
- Increase the number of retries.
- Add a random delta so that different queries don't tend to come in at the same time.

Refs. : Document post-mortem: Missing metadata and misc problems #851


jvwong:
FYI there was a case detailed here: https://github.com/PathwayCommons/factoid/issues/845#issuecomment-705275241 where the exisinting retry did in fact work

jvwong (approved):

Issue #861, opened 2020-10-06
Grounding: Entity organism preference / mixed organisms for genes 


jvwong:
There have been two cases whereby a user has  added genes from different organisms. Turns out this is not what the user intended.

Enhancement is a (non-blocking) message to the author that they have added bioentities from >=1 organism.

maxkfranz:
Popover for grounding:  This organism is different than the other proteins you have (human).  __Did you mean X (human)__?

jvwong:
NCBI gene method:

```
ID: 21803 | transforming growth factor, beta 1 [Mus musculus (house mouse)]
```

jvwong:
> - Warn user when factoid contains multiple organisms
>   - Change “Warnings” to “tips” or alike -- more friendly

Issue #860, opened 2020-10-06
Block users from submitting empty documents


jvwong:
There are several instances where users submitted a blank document, that is, devoid of elements. 

The enhancement is to:
  - notify the user that they need to add an element
  - disable the submit button accordingly

jvwong:
- Only confirmation button is disabled
- Warning, possible link to help

PR #866, opened 2020-10-08
More salient organism in entity grounding popover

maxkfranz:
- Make the organism name salient (at the top) in the grounding popover.
- Add `ent.organism()` shortcut function to get the `Organism` an `Entity` is grounded to (may be null for chemical).
- Add `doc.commonOrganism()` to get the most common `Organism` in a `Document` -- may be null if no NCBI groundings exist.
- Add `doc.irregularOrganismEntities()` to get an array of entities that are grounded to an organism other than the common one.

Refs. :

- Grounding: Entity organism preference / mixed organisms for genes #861
- Block users from submitting empty documents #860


maxkfranz:
@jvwong, you should be able to use those functions for the submission task list.

jvwong (approved):

PR #868, opened 2020-10-09
Iss860 empty submit iss861 irregular org

jvwong:
Add warnings in the submit tooltip:

1. The user has entities (genes/products) that come from >1 organism

Screenshot_2020-10-09 Detection of small vessels for microsurgery by a Doppler flowmeter

2. The user has no elements in their doc

Screenshot_2020-10-09 Detection of small vessels for microsurgery by a Doppler flowmeter (1)

Here, the 'show me help' toggles the help, but I am having trouble understanding how to close the tooltip on clicking the hypertext.

image

Refs #860 and #861

### DIFF
--- a/src/client/components/editor/submit.js
+++ b/src/client/components/editor/submit.js
@@ -1,17 +1,27 @@
+import EventEmitter from 'eventemitter3';
+
 import { TaskView } from '../tasks';
 import Popover from '../popover/popover';
 import { makeClassList } from '../../../util';
 import h from 'react-hyperscript';
 
 export const Submit = props => {
+
+  const emitter = new EventEmitter();
   const { document, bus, controller } = props;
 
   if( !document.editable() ){
     return null;
   }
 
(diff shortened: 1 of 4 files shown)


maxkfranz:
Looks good, but I'd just remove the button rather than disable it.  Invariably, someone's going to try to click the greyed out button anyway.

maxkfranz:
Maybe 'show me how' could be a salient button instead of link.

jvwong:
Do we care about the tooltip closing when the help is open? If not then this one is good to go.

maxkfranz:
Does the submit tooltip stay open after you close the help?  If so, the submit tooltip should be closed on click.

jvwong:
> Does the submit tooltip stay open after you close the help? If so, the submit tooltip should be closed on click.

No everything closes.

maxkfranz:
OK then it’s good to go

PR #871, opened 2020-10-13
Carousel misc tweaks

maxkfranz:
Misc. fixes that would be good to include in the next release


maxkfranz:
Without smooth scroll:

no-smooth

With smooth scroll:

smooth

PR #877, opened 2020-10-14
V0.12.2

jvwong:
- Minor
  - Adding the google-chrome deps for Docker


maxkfranz (approved):
Looks good.  I just double-checked unstable, and it looks like it's running OK as well.

jvwong:
This is running on prod now.

PR #881, opened 2020-10-14
Provide an admin facility to add articles using biopax file from url

metincansiper:
- Works with [biopax_to_factoid branch of factoid-converters](https://github.com/PathwayCommons/factoid-converters/tree/biopax_to_factoid).


src/server/routes/api/document/index.js
```diff
@@ -1295,6 +1307,52 @@ http.post('/', function( req, res, next ){
...
+        };
+
+        return searchGrounding( opts ).then( res => {
+          if( res && res.length > 0 ) {
+            return entity.associate( res[0] );
+          }
+        } );
+      } );
+
+      let intnPromises = intns.map( intn => intn.associate( 'interaction' ) );
```
maxkfranz, line 1345: Are all the interactions of the general 'interaction' type, or do we use specific types like 'binding'?

maxkfranz:
> The entity names are not rendered on entities but when I click them I see that their name and also see that they are actually associated. It must be a UI issue in rendering entities.
When I click on an interaction it still does not show the last UI step. I think that it also probably is a UI issue similar to the case of entities.

You could do a diff on `doc.json()` to see what's different between a BioPAX-generated document and a manually created one:

1. Create a document via BioPAX.
2. Create a document with the same entities and interactions as (1) manually.
3. Diff the JSON: `doc1.json()` versus `doc2.json()`

maxkfranz:
@metincansiper & @jvwong : To test this, it might be simplest to merge this into unstable so that this can be tried out on unstable.factoid.baderlab.org.  The unstable instance should use an environment variable to use the latest `factoid-converters`.

@metincansiper, do you have a publicly-accessible instance of the latest `factoid-converters`?

src/server/routes/api/document/index.js
metincansiper: @maxkfranz you are right it may possibly be other vlaues and actually it is already specified in ``association`` and ``tyoe`` fields (I always put the same value to both) in interaction json. Therefore, I totally removed the related code now.

maxkfranz:
> The entity names are not rendered on entities but when I click them I see that their name and also see that they are actually associated. It must be a UI issue in rendering entities.

An element may have to be complete (`ele.complete()`).  This normally marks that the user has performed some action to complete the entity (e.g. clicking the interaction type manually versus leaving it the default value -- in both cases there is still an associated type).

metincansiper:
> An element may have to be complete (`ele.complete()`). This normally marks that the user has performed some action to complete the entity (e.g. clicking the interaction type manually versus leaving it the default value -- in both cases there is still an associated type).

Thank you, the problem has been fixed after calling `ele.comptlete()` on elements.

metincansiper:
>@metincansiper, do you have a publicly-accessible instance of the latest factoid-converters?

No. Is the factoid-converters deployed automatically to "http://biopax.baderlab.org/convert" from either master or unstable of factoid-converters?

metincansiper:
One other thing is what must be the email address used here (https://github.com/PathwayCommons/factoid/pull/881/commits/eb1564d801195107c728905d90b8d1475bae837a#diff-24330a3dafb9473fe683be4b0a77ec948baf915b8a675a9fd6981009332aa77aR91). I did not want to assign it to the pc help email address to avoid the email that would be sent. I wanted to use an address most probably non-existing like ``pc@biofactoid.com``. 

The other question is should we avoid sending emails in a way when multiple articles are created through a biopax file url.

jvwong:
> > @metincansiper, do you have a publicly-accessible instance of the latest factoid-converters?
> 
> No. Is the factoid-converters deployed automatically to "http://biopax.baderlab.org/convert" from either master or unstable of factoid-converters?

The unstable/master instances (running on XXX.XXX.81.174 ) can access a local converters instance exposed on port 9999. Currently, it is accessing the prod converter version, but the `factoid-unstable.sh` can be updated to point to local. Just tell me what branch/commit to run.

src/server/routes/api/document/index.js
maxkfranz: Great

jvwong:
> I think it's this branch here `biopax_to_factoid`: https://github.com/PathwayCommons/factoid-converters/tree/biopax_to_factoid
> 
> It's got a dockerfile, so that's probably the easiest way to deploy it. Manually setting up all the java stuff on a vm would be a pain.

I've got [pathwaycommons/factoid-converters:biopax_to_factoid](https://hub.docker.com/layers/pathwaycommons/factoid-converters/biopax_to_factoid/images/sha256-4c732216cd456490e9f4c55c1e1327b99370ccdbced1d48aca5c69a5c2b0d7bb?context=repo) running. Turns out the unstable.factoid.baderlab.org instance was always pointing at a local converters via `BIOPAX_CONVERTER_URL="http://localhost:9999/convert/v2/"`

metincansiper:
>The document creation route could have a new parameter (i.e. `email: true | false`) to control whether emails are sent.  We may need a similar flag for avoiding tweets on submission.
It may also be a good idea to have a flag in the document itself to indicate that these documents are generated from PC (something like `source: pc | author`).

I added the controls for email on doc creation and the source. However, I guess that there may not be any need for updating anything to disable tweet on submit.  I think so because the tweeting is done when the submit is done through requesting this service endpoint (https://github.com/PathwayCommons/factoid/blob/unstable/src/server/routes/api/document/index.js#L1426). In our case I just call ``doc.submit()``. Am I right?

jvwong:
> > The document creation route could have a new parameter (i.e. `email: true | false`) to control whether emails are sent.  We may need a similar flag for avoiding tweets on submission.
> > It may also be a good idea to have a flag in the document itself to indicate that these documents are generated from PC (something like `source: pc | author`).
> 
> I added the controls for email on doc creation and the source. However, I guess that there may not be any need for updating anything to disable tweet on submit. I think so because the tweeting is done when the submit is done through requesting this service endpoint (https://github.com/PathwayCommons/factoid/blob/unstable/src/server/routes/api/document/index.js#L1426). In our case I just call `doc.submit()`. Am I right?

Yes.

The editor `TaskView` coordinates what happens after the user clicks the submit button: i.e. set the doc status 'public' which leads to 1. updating related papers 2. sending thank you email and 3. tweeting it

jvwong:
> > I think it's this branch here `biopax_to_factoid`: https://github.com/PathwayCommons/factoid-converters/tree/biopax_to_factoid
> > It's got a dockerfile, so that's probably the easiest way to deploy it. Manually setting up all the java stuff on a vm would be a pain.
> 
> I've got [pathwaycommons/factoid-converters:biopax_to_factoid](https://hub.docker.com/layers/pathwaycommons/factoid-converters/biopax_to_factoid/images/sha256-4c732216cd456490e9f4c55c1e1327b99370ccdbced1d48aca5c69a5c2b0d7bb?context=repo) running. Turns out the unstable.factoid.baderlab.org instance was always pointing at a local converters via `BIOPAX_CONVERTER_URL="http://localhost:9999/convert/v2/"`

I think I figured out that your converter endpoint `/convert/v2/biopax-url-to-json` wants an archive of some sort (tar.gz)? Is this true?

jvwong:
There's no grounding information coming out of that converter: see https://github.com/PathwayCommons/factoid-converters/issues/14 



Right now, we could probably fake it, but we need some sort of object that looks like the grounding information:

Currently: 
```json
{
			"type": "protein",
			"name": "Oct1",
			"id": "e35b8c32-165e-446c-be2e-c3a4e946ffaa"
		}
```

Desired: 
```json
  {
    "namespace": "ncbi",
    "type": "ggp",
    "dbName": "NCBI Gene",
    "dbPrefix": "ncbigene",
    "id": "6580",
    "organism": "9606",
    "organismName": "Homo sapiens",
    "name": "SLC22A1",
    "synonyms": [
      "HOCT1",
      "OCT1",
      "oct1_cds",
      "solute carrier family 22 member 1",
      "organic cation transporter 1",
      "solute carrier family 22 (organic cation transporter), member 1"
    ],
    "dbXrefs": [
      {
        "db": "MIM",
        "id": "602607"
      },
      {
        "db": "HGNC",
        "id": "HGNC:10963"
      },
      {
        "db": "Ensembl",
        "id": "ENSG00000175003"
      }
    ],
    "typeOfGene": "protein-coding",
    "esScore": 10.948418,
    "defaultOrganismIndex": 1,
    "organismIndex": 0,
    "combinedOrganismIndex": 1,
    "distance": 0,
    "nameDistance": 1,
    "overallDistance": 100100
  },

```

src/server/routes/api/document/index.js
```diff
@@ -1286,15 +1298,78 @@ const tryVerify = async doc => {
...
+
+      let entityPromises = entities.map( entity => {
+        let name = entity.name();
+        let opts = {
+          limit: 1,
+          offset: 0,
+          name
+        };
+
+        return searchGrounding( opts ).then( res => {
```
jvwong, line 1340: The BioPAX source will provide an `bp:EntityReference`  with `bp:UnificationXref` , in the case of PhosphoSitePlues, some `uniprot knowledgebase`  id.

We'll have to map those rather than search.

jvwong (changes requested):
The grounding should be solved by mapping rather than search.

src/server/routes/api/document/index.js
jvwong: e.g. 
```
wget --quiet \
  --method POST \
  --header 'accept: application/json' \
  --header 'content-type: application/x-www-form-urlencoded' \
  --body-data 'from=ACC&to=P_ENTREZGENEID&query=P14873&format=tab' \
  --output-document \
  - https://www.uniprot.org/uploadlists/
```

maxkfranz:
Next steps:

1. Reenable Uniprot within the grounding service.
2. Add an option to the aggregate search (i.e. `/search`) s.t. namespaces can be blocked (i.e. uniprot namespaces aren't returned to Biofactoid UI).  This can accept an option to set the block list manually, but we will have a default value s.t. uniprot is blocked by default.
3. Within the PC import, instead of doing a `/search` (today), do a `/get` (e.g. `/get` for `uniprot:P01234`).
4. Now, you have a uniprot grounding.  So, within it, get the NCBI xref.
5. Query `/get` with the NCBI xref.
6. Use the NCBI grounding for the factoid entity in the `association`.

Once done:  Run the import on the master.factoid.baderlab.org instance.  Merge the data into

maxkfranz:
Example input for the grounding service (p53):

https://www.uniprot.org/uniprot/P04637.xml

```
<dbReference type="GeneID" id="7157"/> 
```

maxkfranz:
Re. (2), this search function should be amended.  It may be simpler than initially thought, because there is already filtering built in for one namespace at a time.  We just need to support multiple.

https://github.com/PathwayCommons/grounding-search/blob/c2b5bae9756daa8fe640ed1197af4434390da9b6/src/server/db/db.js#L289-L332

maxkfranz:
For chebi, you don't need to convert.  Just do (6) and (7).

maxkfranz:
Jeff & Max: 

- Jeff to do (1) and (3).
- Max to do (2).
- Jeff to deploy master.grounding.baderlab.org with latest code

jvwong:
@metincansiper the grounding service is up and running @ https://master.grounding.baderlab.org/get

maxkfranz:
Example:

Screen Shot 2020-10-22 at 11 04 45

JS from Insomnia:

```js
fetch("https://master.grounding.baderlab.org/get", {
  "method": "POST",
  "headers": {
    "content-type": "application/json"
  },
  "body": {
    "namespace": "uniprot",
    "id": "P01375"
  }
})
.then(response => {
  console.log(response);
})
.catch(err => {
  console.error(err);
});
```

JSON output:

```
{
  "namespace": "uniprot",
  "type": "protein",
  "dbName": "UniProt Knowledgebase",
  "dbPrefix": "uniprot",
  "id": "P01375",
  "organism": "9606",
  "name": "Tumor necrosis factor",
  "geneNames": [
    "TNF",
    "TNFA",
    "TNFSF2"
  ],
  "proteinNames": [
    "Tumor necrosis factor",
    "Cachectin",
    "TNF-alpha",
    "TNF-a"
  ],
  "synonyms": [
    "Cachectin",
    "TNF-alpha",
    "TNF-a",
    "TNF",
    "TNFA",
    "TNFSF2"
  ],
  "dbXrefs": [
    ...
    {
      "db": "GeneID", // we want this one for the ncbi id
      "id": "7124"
    },
    ...
  ]
}
```

metincansiper:
@maxkfranz I run the script that you pasted above. Here is the result:

```
Response {
  size: 0,
  timeout: 0,
  [Symbol(Body internals)]: 
   { body: 
      PassThrough {
        _readableState: [Object],
        readable: true,
        domain: null,
        _events: [Object],
        _eventsCount: 7,
        _maxListeners: undefined,
        _writableState: [Object],
        writable: true,
        allowHalfOpen: true,
        _transformState: [Object] },
     disturbed: false,
     error: null },
  [Symbol(Response internals)]: 
   { url: 'https://master.grounding.baderlab.org/get',
     status: 503,
     statusText: 'Service Unavailable',
     headers: Headers { [Symbol(map)]: [Object] },
     counter: 0 } }
```

maxkfranz:
The insomnia snippet is missing the response.json() part

metincansiper:
Looks like https://master.grounding.baderlab.org/ is giving service unavailable error independent of the query

maxkfranz:
Yes, it looks like the docker instance for that url is down.  @jvwong, would you restart it?

jvwong:
> Looks like https://master.grounding.baderlab.org/ is giving service unavailable error independent of the query
>
> Yes, it looks like the docker instance for that url is down. @jvwong, would you restart it?

The master.grounding.baderlab.org is just the nightly cron build. It's been failing with each cron build because it complains about not having fibers. I tried a bunch of things but maybe you guys will have more luck.

> cat grounding-search.screen.log

```
> grounding-search@0.3.0 boot /home/baderlab/grounding-search
> run-s refresh start


> grounding-search@0.3.0 refresh /home/baderlab/grounding-search
> run-s clear update


> grounding-search@0.3.0 clear /home/baderlab/grounding-search
> node -r esm src/server/datasource/cli clear all

/home/baderlab/grounding-search/src/server/datasource/processing.js:1
Error: Cannot find module 'fibers/future'
Require stack:
- /home/baderlab/grounding-search/src/server/datasource/processing.js
- /home/baderlab/grounding-search/src/server/datasource/uniprot.js
- /home/baderlab/grounding-search/src/server/datasource/cli.js
    at Object.<anonymous> (/home/baderlab/grounding-search/src/server/datasource/processing.js:1)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
...
```

maxkfranz:
Merging in the new lock file sorted it out

metincansiper:
>5. Now, you have a uniprot grounding. So, within it, get the NCBI xref.

@maxkfranz I thought that you mean that I must find the dbxref where the db is ``ncbi`` but I could not find such a dbxref when I made the get query for the entities of 30 random documents. So, I suspected you may mean somethjing else. What did you actually mean?

maxkfranz:
You want the `GeneID` key.  Uniprot used an unclear name.

```
  "dbXrefs": [
    ...
    {
      "db": "GeneID", // we want this one for the ncbi id
      "id": "7124"
    },
    ...
  ]
```

jvwong:
> Merging in the new lock file sorted it out

I'm not sure what time you did that but its down still/again?

maxkfranz:
There's some monkey business going on in that vm.  I did a quick spot check and none of gcc, g++, or make were installed.  There's no way it could have ever built fibers before without those.  I installed them and restarted it, and it looks OK.

Did (the other) Jeff change anything on that vm?  I don't have time to investigate this further.  If there are problems in future, then maybe that machine should only use docker.

metincansiper:
>You want the GeneID key. Uniprot used an unclear name.

Thank you! I updated this PR based on the discussion here. However, there are still being some entities where there is no grounding info. Should we keep them ungrounded? Also, I realized that the related papers UI is failing when an ungrounded entity is clicked.

jvwong:
> > You want the GeneID key. Uniprot used an unclear name.

It's on my list to create a service for this mapping between db name spaces....
> 
> Thank you! I updated this PR based on the discussion here. However, there are still being some entities where there is no grounding info. Should we keep them ungrounded? Also, I realized that the related papers UI is failing when an ungrounded entity is clicked.

That's not surprising that there are missing db links. Can you provide an example as well?

maxkfranz:
> Thank you! I updated this PR based on the discussion here. However, there are still being some entities where there is no grounding info. Should we keep them ungrounded? Also, I realized that the related papers UI is failing when an ungrounded entity is clicked.

If there's no grounding, we should just drop the document -- unless someone wants to manually curate problematic ones by settings the groundings by hand.

metincansiper:
One thing I realize is that there are other db's then uniprot knowledge base. One of them is hgnc symbol. I will check if there is even more alternatives. It maybe one of the reasons why we are missing some groundings since I was only looking for uniprot knowledge base. Should we consider other db types then uniprot knowledge as well?

metincansiper:
Here are the set of db names coming from the file:

[hgnc symbol, uniprot isoform, ena, refseq, uniprot knowledgebase, uniprot]

Uniprot has 3 different variations I guess I can get uniprot id from any of that 3. So far I was just looking for uniprot knowledgebase. Is there any thing to do for the non-uniprot names?

jvwong:
- NCBI Gene uid from RefSeq Accession:

```
curl --request POST \
  --url https://eutils.ncbi.nlm.nih.gov/entrez/eutils/elink.fcgi \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data 'id=NM_004864,NM_001126117' \
  --data cmd=neighbor \
  --data retmode=json \
  --data db=gene \
  --data dbfrom=nuccore \
  --data linkname=nuccore_gene
```

OUT
```json
{
  "header": {
    "type": "elink",
    "version": "0.3"
  },
  "linksets": [
    {
      "dbfrom": "nuccore",
      "ids": [
        "1519243108",
        "1894803104"
      ],
      "linksetdbs": [
        {
          "dbto": "gene",
          "linkname": "nuccore_gene",
          "links": [
            "9518",
            "7157"
          ]
        }
      ]
    }
  ]
}

```

- NCBI Gene uid from HGNC symbol

```
curl --request POST \
  --url https://apps.pathwaycommons.org/api/enrichment/validation \
  --header 'Accept: application/json' \
  --header 'Content-Type: application/json' \
  --data '{
	"query": [
		"TP53",
		"GDF15"
	],
	"namespace": "ncbigene"
}'
```

OUT
```json
{
  "header": {
    "type": "elink",
    "version": "0.3"
  },
  "linksets": [
    {
      "dbfrom": "nuccore",
      "ids": [
        "1519243108",
        "1894803104"
      ],
      "linksetdbs": [
        {
          "dbto": "gene",
          "linkname": "nuccore_gene",
          "links": [
            "9518",
            "7157"
          ]
        }
      ]
    }
  ]
}
```

metincansiper:
>NCBI Gene uid from RefSeq Accession:
NCBI Gene uid from HGNC symbol

Okay I considered these and now I consider any possible db name expect ena. I will search how to convert ena id to ncbi id. Would you let me know if you already know of an existing service for that?

>If there's no grounding, we should just drop the document -- unless someone wants to manually curate problematic ones by settings the groundings by hand.

I wonder if we could consider only dropping the related interaction instead of the whole document?

maxkfranz:
> I wonder if we could consider only dropping the related interaction instead of the whole document?

That's a good point, but I suspect a document may not make much sense if the removed interaction is a key one in the document.

Anyway, the point of importing this data is to get iconic (highly cited) articles.  These articles are old, and the authors probably wouldn't create factoids for them if we asked.  It's not a big deal if we drop some articles as long as we get many of the iconic ones.  We don't even need all of the iconic ones.  We don't have any of them now, so having a good number of them would be a big improvement.

We may end up just cherry-picking which documents get imported into prod -- probably far from all of them.  So I wouldn't worry about making it perfectly robust for ID edge cases.  I think it's fine to just log the problematic cases to the console, and we can reevaluate based on that data.  We can always improve things later if we want to increase our coverage.

metincansiper:
Okay, I skipped the documents including entities with no grounding. The only remaining thing is handling ``ena`` db entities. I could not find any service to convert ena id into ncbi id.

maxkfranz:
> Okay, I skipped the documents including entities with no grounding. The only remaining thing is handling `ena` db entities. I could not find any service to convert ena id into ncbi id.

How many of these `ena` IDs are there?

metincansiper:
>How many of these ena IDs are there?

There are 149 of them.

maxkfranz:
How many documents are there in total?  If the 149 are a small number and they don't include any foundational articles, then it's fine to just discard them.

metincansiper:
>How many documents are there in total? If the 149 are a small number and they don't include any foundational articles, then it's fine to just discard them.

@maxkfranz ena entities are in 147 different documents out of 16557 documents. 16557 is the number of documents remaining after excluding the ones including entities with no db xref. Before excluding them there are 16567 documents in total. I think 147 looks pretty small as compare to whole as you mentioned.

maxkfranz:
OK, that makes sense, Metin. I think it’s fine to exclude those documents for now. We can reevaluate later if needed. 

At the end of the conversion, we should have a list of excluded documents that we can review manually. Maybe a CSV that we could import into a spreadsheet (something like id, name, pc url, columns)

metincansiper:
>OK, that makes sense, Metin. I think it’s fine to exclude those documents for now. We can reevaluate later if needed.

Okay, then I think that the code is ready for review.

maxkfranz:
OK. While Jeff and I review, could you put together that spreadsheet I mentioned?  

It looks like there’s also a conflict that needs to be resolved before merging. 

Thanks

metincansiper:
>OK. While Jeff and I review, could you put together that spreadsheet I mentioned?

Okay, is this row you mentioned (id, name, pc url, columns) all for a document? What do you expect to have in columns field?

maxkfranz:
I just meant that it should include these columns:

- ID
- Name
- PC URL

If you think of other columns that would be useful, feel free to add them.

metincansiper:
Actually paxtools provide me with the pubmed ids of interactions and I make the documents by grouping interactions by their pubmed ids. I can query the document name as well by using the same way as it is done in factoid. However, I guess there may not be a PC URL for these documents/pmid's. What do you think?

maxkfranz:
The pubmed id is the main thing, so that’s fine

maxkfranz:
A list with just the pubmed ids would be fine

metincansiper:
>The pubmed id is the main thing, so that’s fine

Okay, where were you meaning to download(factoid client side, factoid server side or java server side)?

maxkfranz:
It doesn't matter to me.  Wherever it's easiest.  You could even just log them to the console and copy to Google Sheets.

We just need the list once.  The list generation probably won't be a proper, reusable feature of the system, since we don't expect to need it more than once.

Does that make sense?

metincansiper:
>It doesn't matter to me. Wherever it's easiest. You could even just log them to the console and copy to Google Sheets.
We just need the list once. The list generation probably won't be a proper, reusable feature of the system, since we don't expect to need it more than once.
Does that make sense?

Sure I created the Google Sheet with pmids: https://docs.google.com/spreadsheets/d/1Aafdwi3f9pmlPCD_zrGIrUo_s3hp9nIEdbux8pAThig/edit#gid=0

maxkfranz:
Great, Metin.  Could you add a column that just links out to pubmed for the ID? We'll need to go over the articles on Pubmed to determine whether any of them are important.

metincansiper:
>Great, Metin. Could you add a column that just links out to pubmed for the ID? We'll need to go over the articles on Pubmed to determine whether any of them are important.

I updated the document in that way.

src/server/routes/api/document/index.js
maxkfranz: @metincansiper & @jvwong, can this be marked as resolved?

src/server/routes/api/document/index.js
```diff
@@ -8,8 +8,8 @@ import cytosnap from 'cytosnap';
 import Twitter from 'twitter';
 import LRUCache from 'lru-cache';
 import emailRegex from 'email-regex';
-// import url from 'url';
 import fs from 'fs';
+import { URLSearchParams } from  'url';
 
 import { exportToZip, EXPORT_TYPES } from './export';
```
maxkfranz, line 14: This file needs the conflicts to be resolved

maxkfranz (changes requested):

src/server/routes/api/document/index.js
jvwong: Looks fine to me. We can always pop in the grounding-search `/map` here instead, now or later @metincansiper

src/server/routes/api/document/index.js
metincansiper: @maxkfranz I resolved the conflict and merged the current unstable in this branch.

maxkfranz: Great, thanks Metin

maxkfranz (approved):
Unless there are any red flags, let's merge this in for testing on the master instance.

jvwong:
@metincansiper does this depend on https://github.com/PathwayCommons/factoid-converters/tree/biopax_to_factoid ?

metincansiper:
>@metincansiper does this depend on https://github.com/PathwayCommons/factoid-converters/tree/biopax_to_factoid ?

@jvwong yes

jvwong:
> > @metincansiper does this depend on https://github.com/PathwayCommons/factoid-converters/tree/biopax_to_factoid ?
> 
> @jvwong yes

OK, then let's get that converter code in as well.

maxkfranz:
Great.  Let's merge in the convertors to the latest, and let's merge this in.  Then we can tag and release.

The email template has been updated to the latest version of the text from the Google Doc.  I verified that the emails are sent by hitting the refresh button on the admin UI, so we're ready to go.  Once we've deployed prod, we can send out the emails for one of the already submitted factoids.

metincansiper:
>OK, then let's get that converter code in as well.

I was not able to merge and push the changes into master directly. I got the following logs:

```
remote: Resolving deltas: 100% (1/1), completed with 1 local object.
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: Required status check "continuous-integration/travis-ci" is expected.
To https://github.com/PathwayCommons/factoid-converters.git
 ! [remote rejected] master -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/PathwayCommons/factoid-converters.git'
```

Therefore, I created a PR instead (https://github.com/PathwayCommons/factoid-converters/pull/18). I  am not allowed to merge it in yet. It is pending as below right now, I am not able to click somewhere to visit the travis run:

Screen Shot 2020-11-18 at 3 08 24 PM

maxkfranz:
@metincansiper, I've made you an admin for the convertors repo, so you should be able to override and merge form now on, if you need.  Travis is just really slow lately.  It looks like the builds haven't even started.

metincansiper:
Okay I merged both of this one and the PR in factoid-converters.

metincansiper:
@jvwong considering this comment (https://github.com/PathwayCommons/factoid/pull/881#issuecomment-710490695) and so PathwayCommons/factoid-converters#14: 

We are currently searching the grounding by the xrefs coming from converter and using the results as here (https://github.com/PathwayCommons/factoid/blob/unstable/src/server/routes/api/document/index.js#L1675). 

Here is the example of an xref currently coming from factoid converters (I created a PR for adding the ``org`` field it is not merged yet):

```
xref:  { db: 'hgnc symbol', id: 'EGFR', org: '9606' }
```

and this is the grounding that is retrieved in factoid side using that xref:

```
{
  namespace: 'ncbi',
  type: 'ggp',
  dbName: 'NCBI Gene',
  dbPrefix: 'ncbigene',
  id: '1956',
  organism: '9606',
  organismName: 'Homo sapiens',
  name: 'EGFR',
  synonyms: [
    'ERBB',
    'ERBB1',
    'ERRP',
    'HER1',
    'NISBD2',
    'PIG61',
    'mENA',
    'epidermal growth factor receptor',
    'avian erythroblastic leukemia viral (v-erb-b) oncogene homolog',
    'cell growth inhibiting protein 40',
    'cell proliferation-inducing protein 61',
    'epidermal growth factor receptor tyrosine kinase domain',
    'erb-b2 receptor tyrosine kinase 1',
    'proto-oncogene c-ErbB-1',
    'receptor tyrosine-protein kinase erbB-1'
  ],
  dbXrefs: [
    { db: 'MIM', id: '131550' },
    { db: 'HGNC', id: 'HGNC:3236' },
    { db: 'Ensembl', id: 'ENSG00000146648' }
  ],
  typeOfGene: 'protein-coding'
}
```

Are you suggesting to send the grounding as the second json directly from the converters instead of sending the xref from converter (and so not retrieving grounding from the factoid based on the xref as we are doing now but using the input from the converter as is)?

jvwong:
I think things have improved a lot since that comment and maybe its OK now.

I think the original problem was that there wasn't any grounding info at all.

PR #884, opened 2020-10-26
Friendlier message for editor Task warnings

jvwong:
Screenshot at Oct 26 12-46-16

Refs #861


gbader:
Looks great!

Should be "Could you double-check the following?"

Issue #858, opened 2020-10-05
Novel interaction notification


maxkfranz:
## Description

**Q: What is the name of the feature?**

A: Novel interaction notification

**Q: What does this feature enable the user to do?**

A: User X gets a notification when a novel interaction has been listed in Biofactoid and that interaction is relevant to User X's research.

Benefits:

- Biofactoid is advertised to prospective users in a relevant (non-spammy) way.  This helps to increase mindshare of Biofactoid and it helps to increase the number of users that create factoids.
- We incorporate the idea of notifications (like Emmaa), but in an automatic way:  The author does not need to explicitly sign up for the notification -- sign ups create friction.  The author also does not need to know exactly what the interaction would be a priori (e.g. using Emmaa, an author has to specify the exact participants, and an author wouldn't know what the participants would be if the interaction is novel).

**Q: What information must the user provide to use the feature?**

A: A different author, User A, creates a factoid with a novel interaction.  User X shouldn't have to provide any information.

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: 

- We need to have a way to contact User X, even though they may have not used Biofactoid before.
- We need many users, like User A, to create novel factoids.

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: Learns about new interactions that are relevant to their research.  Learns about Biofactoid organically.
- Editor: Perhaps may give an opportunity to advertise their journal (i.e. the novel interaction came from a paper in their journal).
- Computational biologist: N/A
- Curator: N/A

### Mockup

(Email to User X)

Biofactoid has found a novel interaction that we have determined to be relevant to your research using Biofactoid's advanced AI.  Biofactoid is an app that allows authors to create a digital profile of scientific discoveries in an article and connects it to related research.

You can see the new interaction here: https://biofactoid.org/document/{DOCUMENT_ID}

You can connect your own findings with other researchers by adding your articles to Biofactoid.  Get started at biofactoid.org!

### Details

- When User A submits a new factoid, get the list of related articles (as normal).
- Determine whether the article contains a novel interaction, using Indra (i.e. interaction query yields no results).  If the document does not contain any novel interactions, then bail out -- no emails are sent.
- Flag the related articles with a high correlation score (via Semantic Search).
- For each flagged article:
  - Get the corresponding author email (User X) from the Pubmed metadata.
  - Send an email template (see mockup) using Mailjet to the corresponding author.
[...]

PR #886, opened 2020-10-28
Related papers notification

maxkfranz:
Details:

- This is currently a prototype.
- A new email template type is defined as `relatedPaperNotification`.
- The emails are triggered whenever the related papers are updated for a document.
- A new metadata field is available on a document.  `doc.relatedPapersNotified()` can be used to set or get whether the notifications have been sent.  This prevents duplicates.
- Novel interactions are an optional enhancement in the emails:  An author can receive an email whether or not the originating factoid has a novel interaction.  However, if there is a novel interaction, this should be mentioned in the email.
- Currently, only notifications are sent to the terminal.  The emails are disabled until the template is set up.
- Action items are marked by `// TODO RPM`

Action items:

- [ ] Add a template to Mailjet (@jvwong), based on our email document on Google Docs.
- [ ] Reuse our Indra querying code to determine whether there is a novel interaction -- i.e. an interaction that has zero results in Indra (@metincansiper).
- [ ] Verify that our emails send correctly on unstable.  We'll temporary set unstable to send these notifications to Max and Jeff (@maxkfranz @jvwong).
- [ ] Finalise the email template text (@maxkfranz).
- [ ] Research the feasibility of the bonus 'factoidable' nudge:  Can we determine using AI whether a paper is a good candidate for Biofactoid (@cannin @JohnGiorgi @jvwong)?

Ref. : Novel interaction notification #858


maxkfranz:
Note: The linting errors will show you the main places that still need to be updated.

jvwong:
I had one refinement for the pool of papers used to select authors to contact: 

https://github.com/PathwayCommons/factoid/blob/715ede09df6219cec28ba2f54961d86a2a32bac1/src/server/routes/api/document/index.js#L1268

Document-level (`doc.relatedPapers()`) derive from NCBI's EUTILS ELINK service, using the Biofactoid document's article PMID as the query to retrieve what PubMed defines as a linked article. Now recall Biofactoid does some filtering and sorting (semantic-search) before setting the result in the `relatedPapers`.

A subset of the raw set of ELINK articles retrieved consist of: 
  - articles citing the one in Biofactoid (incoming)
  - articles referenced by the one in Biofactoid (outgoing)

So the idea is to select from these incoming and outgoing articles for authors to contact. There's a few reasons to select these:
  - It's self-evident why a researcher would be interested: Someone cited your work, here, take a look in Biofactoid
  - The article in Biofactoid was based on research/interactions in the outgoing subset, and similarly, the incoming subset  would often be building upon the Biofactoid article   
  - The ELINK service wrapper already exists, you can filter by how recently it was published, and just needs a little update 
  - An article being related to the current Factoid might not correlate well with its ability to be deposited etc

maxkfranz:
Great points, Jeff.

Re.:

> An article being related to the current Factoid might not correlate well with its ability to be deposited etc

This applies to the bonus nudge, but we can still notify authors of papers that aren't good candidates to be deposited.  Emails for those papers would exclude the nudge, but those emails would still help to increase mindshare for Biofactoid.

maxkfranz:
Template info:

```
        "TemplateID": 1871553,
        "TemplateLanguage": true,
        "Subject": "Test of related papers notification",
        "Variables": {}
```

jvwong:
So I'm setting a new document attribute `referencedPapers` and its being done alongside the `relatedPapers` so that it hooks into all of the lifecycle updates.

- Notes
  - It's entirely possible that the the references are not made available to PubMed for whatever reason.
  - Right now the `emailRelatedPaperAuthors` gets triggered on doc creation, which means the document wont have any entries much less submitted. So I guess it should be triggered post submit?

maxkfranz:
> Right now the emailRelatedPaperAuthors gets triggered on doc creation, which means the document wont have any entries much less submitted. So I guess it should be triggered post submit?

This should only get triggered at the same time as related papers are saved (or refreshed).  That should be done on submission, not on creation.  Was there a reason for changing the related papers behaviour to search on creation?

Alternatively, `emailRelatedPaperAuthors()` could bail out if the doc state isn't submitted.

maxkfranz:
Re. 

> Reuse our Indra querying code to determine whether there is a novel interaction -- i.e. an interaction that has zero results in Indra (@metincansiper).

It's not absolutely necessary to have this feature for an initial version of the notifications, but it would be great in future iterations to vary the email template based on this factor.  @metincansiper, would it be possible to use a metadata field (e.g. `intn.isNovel()`) in the interactions and just mark them as novel or not during the existing related papers queries?

src/server/routes/api/document/index.js
```diff
@@ -1220,6 +1221,59 @@ const checkApiKey = (apiKey) => {
...
+  const sendEmail = async (paper, novelIntns) => {
+    const contact = getContact(paper);
+    const email = contact.email[0];
+    const name = contact.name;
+
+    // TODO RPN enable to prevent duplicate sending
+    // doc.relatedPapersNotified(true);
+
+    const mailOpts =  await msgFactory(EMAIL_TYPE_REL_PPR_NOTIFICATION, doc, {
+      to: 'maxkfranz@gmail.com', // send to one of us for testing
```
maxkfranz, line 1243:
```suggestion
      to: 'support@biofactoid.org', // send to one of us for testing
```

@jvwong, would you prefer to use the info or support address for testing this?

jvwong:
> > Right now the emailRelatedPaperAuthors gets triggered on doc creation, which means the document wont have any entries much less submitted. So I guess it should be triggered post submit?
> 
> This should only get triggered at the same time as related papers are saved (or refreshed). That should be done on submission, not on creation. Was there a reason for changing the related papers behaviour to search on creation?

I retrieve document-level related papers on creations, because they are drawn from PubMed, and so that there was something to show in the explorer-view immediately post-submit (in fact, on create).

src/server/routes/api/document/index.js
jvwong: Maybe I should make this a config, I wouldn't want all this noise going to the support.

jvwong: like `EMAIL_ADMIN_ADDR`

maxkfranz: The idea is to test this on unstable.  It would use the user's email address on prod.  It needs to go to a shared address if more than one person is able to test it.

jvwong:
OK two updates:

1. Put the code for sending these emails in the set of post-public actions to take (tweet, updated related papers, send follow up).
2. Another config `EMAIL_RELPPRS_CONTACT` that by default, sends to the contact address unless explicitly instructed not to, in which case sends to  `EMAIL_ADMIN_ADDR`. So this won't require any changes to Docker-compose/prod at least.

maxkfranz:
> Put the code for sending these emails in the set of post-public actions to take (tweet, updated related papers, send follow up emails).

OK, I had it together with getting the related papers so that the admin UI could just refresh things with the existing refresh button.  In case of an Indra failure, for example, the related papers may not exist and the emails would never be sent.  The refresh button would allow for both of those to recover, and the related papers refreshing should already be integrated into the admin panel.

The new approach is fine, but the admin UI would now need to explicitly send these emails on refresh (if the sent flag is false).

metincansiper:
>It's not absolutely necessary to have this feature for an initial version of the notifications, but it would be great in future iterations to vary the email template based on this factor. @metincansiper, would it be possible to use a metadata field (e.g. intn.isNovel()) in the interactions and just mark them as novel or not during the existing related papers queries?

@maxkfranz I made the requested change in novel_intn branch (https://github.com/PathwayCommons/factoid/tree/novel_intn). Should I make a PR to your branch used in this PR? BTW I considered when an interaction has no result right after the query made for the interaction. Later, the interactions with not enough result are assigned with some random results from the document. I did not consider that step since I think that any interaction will not be novel in that sense as long as the document has enough results.

jvwong:
> > Put the code for sending these emails in the set of post-public actions to take (tweet, updated related papers, send follow up emails).
> 
> OK, I had it together with getting the related papers so that the admin UI could just refresh things with the existing refresh button. In case of an Indra failure, for example, the related papers may not exist and the emails would never be sent. The refresh button would allow for both of those to recover, and the related papers refreshing should already be integrated into the admin panel.
> 
> The new approach is fine, but the admin UI would now need to explicitly send these emails on refresh (if the sent flag is false).

Oh I see now. If you want it tied to the existing admin refresh, i can do that by adding it below this:

https://github.com/PathwayCommons/factoid/blob/062ee8cd52de38679d4db68c72614f30c761ee3e/src/server/routes/api/document/index.js#L1541

maxkfranz:
Sounds good

maxkfranz:
> > It's not absolutely necessary to have this feature for an initial version of the notifications, but it would be great in future iterations to vary the email template based on this factor. @metincansiper, would it be possible to use a metadata field (e.g. intn.isNovel()) in the interactions and just mark them as novel or not during the existing related papers queries?
> 
> @maxkfranz I made the requested change in novel_intn branch (https://github.com/PathwayCommons/factoid/tree/novel_intn). Should I make a PR to your branch used in this PR? BTW I considered when an interaction has no result right after the query made for the interaction. Later, the interactions with not enough result are assigned with some random results from the document. I did not consider that step since I think that any interaction will not be novel in that sense as long as the document has enough results.

@metincansiper, it's fine to just cherry-pick your commits into this branch.

maxkfranz:
This branch is now ready to be merged for testing on the unstable instance.

maxkfranz:
@jvwong, you can either merge this in now or wait for @metincansiper to add his commits re. Indra.  If you want to merge sooner, @metincansiper could add his changes in a separate PR.

jvwong:
> @jvwong, you can either merge this in now or wait for @metincansiper to add his commits re. Indra. If you want to merge sooner, @metincansiper could add his changes in a separate PR.

@metincansiper can you add your commits to this branch?

metincansiper:
@jvwong I added my commit into this branch

maxkfranz:
OK, should be good to go

jvwong:
This is throwing some error on build:

```

WARNING in ./node_modules/colors/lib/colors.js
127:29-43 Critical dependency: the request of a dependency is an expression
 @ ./node_modules/colors/lib/colors.js
 @ ./node_modules/colors/safe.js
 @ ./node_modules/winston/lib/winston/config.js
 @ ./node_modules/winston/lib/winston.js
 @ ./src/server/logger.js
 @ ./src/util/email.js
 @ ./src/util/index.js
 @ ./src/client/index.js

ERROR in ./node_modules/winston/lib/winston/common.js
Module not found: Error: Can't resolve 'fs' in '/Users/.../factoid/node_modules/winston/lib/winston'
 @ ./node_modules/winston/lib/winston/common.js 12:9-22
 @ ./node_modules/winston/lib/winston.js
 @ ./src/server/logger.js
 @ ./src/util/email.js
 @ ./src/util/index.js
 @ ./src/client/index.js

ERROR in ./node_modules/winston/lib/winston/transports/file.js
Module not found: Error: Can't resolve 'fs' in '/Users/.../factoid/node_modules/winston/lib/winston/transports'
 @ ./node_modules/winston/lib/winston/transports/file.js 10:9-22
 @ ./node_modules/winston/lib/winston/transports.js
 @ ./node_modules/winston/lib/winston.js
 @ ./src/server/logger.js
 @ ./src/util/email.js
 @ ./src/util/index.js
 @ ./src/client/index.js
^C
```

maxkfranz:
For now, you could remove winston from `email.js`.  

However, `email.js` shouldn't strictly be under `/util` in the first place, since it's server-side only.  These util files aren't universal and should be moved, or we'll have this problem again later:

- document.js - server only
- email.js - server only
- dom.js - client only

jvwong:
I was noticing that right now. I'm going to move it now.

jvwong:
Clarification: I'm moving email.js only.

PR #904, opened 2020-11-19
Enabled using uniprot ids instead of ncbi ids while exporting whole documents into zip of biopax files

metincansiper:
- Controlled by a get url parameter named ``idMapping``.


src/server/routes/api/document/index.js
```diff
@@ -518,7 +580,13 @@ http.get('/zip', function( req, res, next ){
  *             description: Download a zip file containing each Document represented as BioPAX in owl file format.
```
jvwong, line 580: Does the swagger need this (also in `/biopax/:id`)

```
/**
 * @swagger
 *
 * /api/document/zip/biopax:
 *   get:
 *     description: Download a single zip file containing every Document represented in Biological Pathway Exchange (BioPAX).
 *     parameters:
 *       - name: idMapping
 *         in: query
 *         description: ...
 *         required: false
 *         type: boolean
 *         allowEmptyValue: true
```

src/model/element/entity.js
```diff
@@ -187,10 +187,10 @@ class Entity extends Element {
     };
```
jvwong, line 187: For the return, if we could start setting `id: assoc.dbPrefix` which is the MIRIAM standard, could make life easier, e.g. index.js `mapToUniprotIds`. I don't think this breaks anything?

src/server/routes/api/document/index.js
```diff
@@ -178,6 +180,66 @@ const DEFAULT_CORRESPONDENCE = {
...
+
+    if ( db !== 'NCBI Gene' ){
+      return Promise.resolve();
+    }
+
+    const opts = {
+      id: [
+        id
+      ],
+      dbfrom: 'ncbigene',
```
jvwong, line 200: See comment above in entity.js about using `dbPrefix` in the xref.db field. That way, we we're not making up names for dbs.

jvwong (changes requested):

src/model/element/entity.js
metincansiper: @jvwong would the updates in this commit (https://github.com/PathwayCommons/factoid/pull/904/commits/12c42a885a19d568c9c2345c153730055892d0b7) covers what you mean?

src/server/routes/api/document/index.js
metincansiper: @jvwong I just made the swagger updates. While testing them out I recognized that https://master.grounding.baderlab.org/ is not available.

jvwong: I had to reboot but its fine now

Issue #907, opened 2020-11-24
Automatic connection of researcher profiles


maxkfranz:
## Description

**Q: What is the name of the feature?**

A: Automatic connection of researcher profiles with a Biofactoid article summary

**Q: What does this feature enable the user to do?**

A: The reader of an article summary can click through on an author's name to see that author's profile, with other articles etc.

**Q: What information must the user provide to use the feature?**

A: The author must provide the title of the article.

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: Not all authors may have a profile on systems like ORCID.

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: Can use an article summary to explore more thoroughly.  The article summary becomes more of a fully-fledged hub.
- Editor: N/A
- Computational biologist: N/A
- Curator: N/A

### Mockup

Currently, the author list in an article summary is not clickable.  This feature would make it so that each author that has a profile (e.g. on ORCID) would have a link on their name.  Clicking the link would open the author's external profile in a new tab.

### Details

- [ORCID allows for finding author profiles via article DOI.](https://members.orcid.org/api/resources/find-myresearchers)
- Each article summary contains the article's DOI, so author ORCID IDs can be found automatically.
- This information should be populated on document creation, where a title is specified at creation.  This information should be refreshed on submission, in case the title was not specified at creation (i.e. #901 ).
- There may be author profiles, other than ORCID, that we consider.

maxkfranz:
@metincansiper, could you investigate this issue?

metincansiper:
>@metincansiper, could you investigate this issue?

I am looking into it

maxkfranz:
@metincansiper 

See https://github.com/PathwayCommons/factoid/pull/910#issuecomment-737277779

PR #910, opened 2020-11-26
Automatic connection of researcher profiles

metincansiper:
Refs  #907

### DIFF
--- a/src/client/components/editor/info-panel.js
+++ b/src/client/components/editor/info-panel.js
@@ -1,7 +1,5 @@
 import h from 'react-hyperscript';
 import { Component } from 'react';
-import * as xmljs from 'xml-js';
-import _ from 'lodash';
 import { DOI_LINK_BASE_URL, PUBMED_LINK_BASE_URL, GOOGLE_SCHOLAR_BASE_URL } from '../../../config';
 // import { Carousel, CAROUSEL_CONTENT } from '../carousel';
 import { makeClassList } from '../../dom';
@@ -65,7 +63,7 @@ export class InfoPanel extends Component {
 
     const citation = document.citation();
     const authorProfiles = document.authorProfiles();
-    const { authors, pmid, title = 'Untitled article', reference, doi, abstract } = citation;
+    const { pmid, title = 'Untitled article', reference, doi, abstract } = citation;
 
     return h('div.editor-info-panel', [
       h('div.editor-info-title', title),
--- a/src/client/components/editor/info-panel.js
+++ b/src/client/components/editor/info-panel.js
@@ -12,45 +12,7 @@ export class InfoPanel extends Component {
   constructor(props){
     super(props);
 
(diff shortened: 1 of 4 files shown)


metincansiper:
I used the paped titled "Race, ethnicity, and NIH research awards." for testing.

maxkfranz:
@metincansiper, looks great.  Could we do the search on the server side when the document is created?  That way the data can be stored and we don't need to make queries to ORCID on each page load.

The data could be stored under the `citation` metadata.  There should already be an entry for the authors in the citation metadata, so the ORCID ID and link could be added under each author entry.

metincansiper:
>@metincansiper, looks great. Could we do the search on the server side when the document is created? That way the data can be stored and we don't need to make queries to ORCID on each page load.

The data could be stored under the citation metadata. There should already be an entry for the authors in the citation metadata, so the ORCID ID and link could be added under each author entry.

@maxkfranz I think putting it under that citation will require making a lot of sync function async. Since there is a high interaction between functions. What do you think?

maxkfranz:
Yes, it definitely makes sense to use async/await for ORCID stuff.

It should probably go just after the Pubmed stuff, since that needs to be loaded before you can query ORCID:

https://github.com/PathwayCommons/factoid/blob/53f40684ffe498b43d8c7a08c613021c5268d40c/src/server/routes/api/document/index.js#L312-L331

Once `fillDocArticle()` has completed, you can read `doc.citation()`.  

It's probably simpler to store your ORCID data separately from `doc.citation()`.  You could use a separate method, like `doc.authorProfiles()`, that returns an array of author info (like `{ name, orcid, etc... }`).  Most of the data for `authorProfiles()` would be the same as `citation().authors`, but you'd also include the ORCID data.

metincansiper:
@maxkfranz I updated the PR as you mentioned.

maxkfranz:
Looks great.  Merged

maxkfranz:
@metincansiper, I noticed a minor issue for old documents.  If a document doesn't have the profiles defined, the summary page throws an error.  Could you use the old Pubmed data as a fallback (without links) if the profiles don't exist?

Thanks

metincansiper:
@maxkfranz I could not figure out the case. I think if ``citation.authors.authorList`` (is this what you mean by the old data?) exists then the ``authorProfiles`` must also exists.

Would you provide an example pmid or document title for me to reproduce?

maxkfranz:
The problem is not document-specific (i.e. PMID).  It happens for documents that were submitted before your change was made.

Those old docs don't have your new data in them, because they were already submitted (i.e. `fillDocArticle()` has already been ran on a previous version of the app).

It should suffice to check whether your new data exists.  If it exists, show the new linked names.  If not, show the old non-clickable names.

metincansiper:
@maxkfranz Okay I created the PR #912

PR #912, opened 2020-12-02
Use old way of showing author names when the author profiles is missing in summary page.

metincansiper:
Refs https://github.com/PathwayCommons/factoid/pull/910#issuecomment-737277779

### DIFF
--- a/src/client/components/editor/info-panel.js
+++ b/src/client/components/editor/info-panel.js
@@ -62,7 +62,8 @@ export class InfoPanel extends Component {
     }
 
     const citation = document.citation();
-    const authorProfiles = document.authorProfiles();
+    // authorProfiles may not be existing for the older documents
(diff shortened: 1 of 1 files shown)


jvwong:
When  I try to view prod rethinkdb data with this PR, the author names are space-delimited. Are we OK with this? It looks a bit odd but its not a showstopper.

Screenshot_2020-12-07 The guanine nucleotide exchange factor VAV3 participates in ERBB4-mediated cancer cell migration

maxkfranz:
I suspect Metin did this so that you don’t accidentally click on one author name accidentally — just because it happened to be close to another name.  I’m agnostic to whether we use spaces or commas.  In the comma case, it could look a bit weird to have the extra space between names, e.g.:

Oliver Hardy, ___ Stan Laurel, ___ Bud Abbott, ___ Lou Costello

In any case, this can be revised with just css, e.g.  `&::after { content: ‘, ' }`

Issue #901, opened 2020-11-12
Post hoc specification of the paper in a factoid


maxkfranz:
## Description

**Q: What is the name of the feature?**

A: Post hoc specification of the paper

**Q: What does this feature enable the user to do?**

A: The author may specify which paper is associated with a factoid after the creation of the document.

**Q: What information must the user provide to use the feature?**

A: The title of the paper.

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: The paper must be specified before the document is submitted.

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: An author is able to experiment with the Biofactoid editor UI before deciding what paper his or her factoid is about.
- Editor: N/A
- Computational biologist: N/A
- Curator: N/A

### Mockup

- In the submission popover, there is the form to specify the paper.  This is shown only for documents that are created without an associated paper.
- Documents that are missing an associated paper show a title like 'Set the title of your paper'.  That title is displayed as a clickable link.  
  - By clicking that link, the author is given a popover that allows him or her to specify the paper's title -- similar to the home page.  
  - The yellow button may say something different than the home page, like 'OK'.  
  - If we already have the email address of the user, then that text field can be omitted.
  - The name of the author should be provided in the same form.  If the author's name is already stored in the system (e.g. they entered the system through an email addressed to them), then that field can be omitted.

<img width="294" alt="Screen Shot 2020-11-12 at 16 52 57" src="https://user-images.githubusercontent.com/989043/99000600-887fae80-2507-11eb-891f-b17e162fdb8f.png">

### Details

Description:

- An author may specify the paper of an associated factoid by clicking the 'Set the title of your paper' link -- which is located where the paper's title would normally be.
- An author is blocked from submitting the document until the factoid has an associated paper.  The paper association UI can be shown in the submission popover to address this.
- The author info form now contains the following fields:
  - Author name
  - Email address
  - Article title

Tasks:

- [x] Modify the existing home page author info popover (to create a new factoid).
  - [x] Add author name to the existing popover.
  - [x] Add the option to customise the yellow button text
  - [x] Add the option to customise the action on click of the yellow submit button
  - [x] Make the component reusable, outside of the home page in a separate file
- [x] Reuse the popover on the editor (show on click for empty titles)
  - [x] Add popover in the editor
  - [x] Autofill the form fields that may already be stored in the model (i.e. email case where we send the message to a particular author, but we don't know the paper yet)
    - [x] Author name
[...]

PR #913, opened 2020-12-05
Refactored RequestForm as a reusable component.

metincansiper:
The first phase of implementation for #901 as we talked with @maxkfranz.


maxkfranz:
Looks great.  The next step would be to add a 'Name' text field in the form that sets the contributing author's name.

metincansiper:
>Looks great. The next step would be to add a 'Name' text field in the form that sets the contributing author's name.

@maxkfranz I added that to this PR. I control the author name through ``doc.authorName()``. I am not sure if the function name is convenient. I could update it if desired.

src/client/components/request-form.js
```diff
@@ -0,0 +1,177 @@
...
+
+  updateForm(fields){
+    this.setState(fields);
+  }
+
+  submitRequest(){
+    const { paperId, authorEmail, authorName } = this.state;
+    const { checkIncomplete = false } = this.props;
+
+    if( !checkIncomplete && ( !paperId || !authorEmail || !authorName ) ){
```
jvwong, line 76: This would imply that author name is required in order for the form to be complete. Is this correct? 
It was always optional as it was part of the post-submit case.

src/model/document/document.js
```diff
@@ -455,6 +455,17 @@ class Document {
     }
   }
 
+  authorName( name ){
```
jvwong, line 458: So my post-submit code stashes the 'name'  under `doc: { provided: { name: "xyz" }`. If you wanna keep this as a top-level attribute doc, you'll have to update the post-submit field:

https://github.com/PathwayCommons/factoid/blob/3a993ef5b7b6a9922a8f9d4053e29bcca0fdf124/src/client/components/tasks.js#L413

jvwong (changes requested):

src/client/components/request-form.js
metincansiper: @jvwong ``checkIncomplete`` is a parameter for this component and so whether to check if any field exists is controlled by that option. Currently either all the fields can be required or optional all together controlled by ``checkIncomplete`` parameter. You mean author name should still be different then other 2 and should not be required even when ``checkIncomplete``  parameter is set?

src/model/document/document.js
metincansiper: @jvwong would this commit (https://github.com/PathwayCommons/factoid/pull/913/commits/4b2c3667d738e9d0ee8a83ca0f90c230cd517ec0) covers what you mean?

src/client/components/request-form.js
jvwong: Oh OK that's fine.

metincansiper:
I realized that I implemented the logic of ``checkIncomplete`` parameter in opposite way and fixed it (https://github.com/PathwayCommons/factoid/pull/913/commits/157b3158f5c1e4807bc75b23428d2df4c7077dd8).

@maxkfranz I guess the next thing to implement is reusing the popover on the editor (around the title)? Actually I think the title will not be visible in that case. Therefore, I can just put the edit button in place of the title?

src/model/document/document.js
maxkfranz: I agree that the data should be stored in the same way as before, ideally through the same PATCH route.

maxkfranz:
> @maxkfranz I guess the next thing to implement is reusing the popover on the editor (around the title)? Actually I think the title will not be available in that case. Therefore, I can just put the edit button in place of the title?

Yes, but in order to do that you need an easy way to create new documents without a title.  You could allow the admin interface to do this by modifying the form to allow the text boxes to be empty.  This could be an option in the component.

Once that's done, you can show some text like 'Set the article's title' where the title would normally go -- in the case where the document is empty.  Clicking that text would open the form, as you've described.

metincansiper:
>Yes, but in order to do that you need an easy way to create new documents without a title. You could allow the admin interface to do this by modifying the form to allow the text boxes to be empty. This could be an option in the component.

I think this is what the existing ``checkIncomplete`` option of component does. I am able to do it by using that option.

src/model/document/document.js
jvwong: > @jvwong would this commit ([4b2c366](https://github.com/PathwayCommons/factoid/commit/4b2c3667d738e9d0ee8a83ca0f90c230cd517ec0)) covers what you mean?

My doc attribute is `provided.name` rather than `authorName`

metincansiper: Okay sorry! I set ``provided.name`` instead of ``provided.authorName``.

jvwong (approved):

metincansiper:
@maxkfranz @jvwong I updated this PR by reusing the request for around the editor title. For now I enabled any field in the request form in the editor no matter if any of the fields is already set. I guess we may want to update the logic there. I realized that the server service for editing documents ('api/document/${id}/${secret}') checks ``document.provided()`` for paperId and authorEmail rather then the service parameter values. Therefore, I needed to update the ``document.provided()`` first and them make the service call as here (https://github.com/PathwayCommons/factoid/pull/913/commits/6d434488f756826389492a78d829b66c408258d2#diff-2e21dfae6311a798a1ba99795085d0176dba36a998c1f785963426bf83c2fbacR145). I wonder if it would be a inconsistency problem if there would be an exception between ``document.provided()``  and service calls? Also for the ``authorName`` I am just updating ``provided.name`` field. Should that one also be considered in the service like the other 2 fields?

maxkfranz:
Looks good.  I think it may be a bit simpler and less error prone for the case that you mentioned to just use the `doc` functions:

https://github.com/PathwayCommons/factoid/commit/6d434488f756826389492a78d829b66c408258d2#diff-2e21dfae6311a798a1ba99795085d0176dba36a998c1f785963426bf83c2fbacR113-R152

The extra fetch that you have at the end of that `else` is fine for now, but if the fetch/patch logic changes in future -- and it's changed many times already -- there could be a conflict.  If there are any missing setters in the `Document` class, feel free to add them to simplify the `else` block you mentioned.

So instead of making everything in the component use `fetch()`, that particular case can just use `doc` functions.  Does that make sense?

Once that's done, I think we can merge this in.

src/client/components/editor/title.js
```diff
@@ -1,28 +1,77 @@
...
+    const { document } = this.props;
+    const { citation } = this.state;
+    const { authors: { abbreviation }, title = 'Unnamed document', reference, doi } = citation;
+
+    if( !document.editable() ){
+      return null;
+    }
+
+    const getTitleContent = () => {
+       if ( title == null ) {
```
jvwong, line 42: What are the conditions under which I'm allowed to edit title? I don't think `null` is even possible?

jvwong:
Do we want the 'name' field in the homepage popup as a requirement?

src/client/components/editor/title.js
maxkfranz: It might be better as `if(title)` to handle the empty string case.  It should be possible to have an empty title as a part of the email campaign.  We send a completely empty document to a user so that they can try it out and fill in the paper details after the fact (#901).

maxkfranz:
> Do we want the 'name' field in the homepage popup as a requirement?

It helps to simplify things.  I've got some changes planned around the submission UI and this title editing popover planned after Metin's PR.

src/client/components/editor/title.js
jvwong: I'm missed this discussion  but I thought the point was to always make the info editable?

maxkfranz: The main editable case is for the when you start with an empty doc (e.g. related papers emails).  Is there another case that would benefit from editing that info after the fact?  If all the other cases have the info already set, is there a benefit of being able to change those values?

I suppose you could have an author's paper who changed the title of the paper slightly, so they want to update the factoid.  That's probably a rare edge case.

metincansiper:
>Looks good. I think it may be a bit simpler and less error prone for the case that you mentioned to just use the doc functions:
6d43448#diff-2e21dfae6311a798a1ba99795085d0176dba36a998c1f785963426bf83c2fbacR113-R152
The extra fetch that you have at the end of that else is fine for now, but if the fetch/patch logic changes in future -- and it's changed many times already -- there could be a conflict. If there are any missing setters in the Document class, feel free to add them to simplify the else block you mentioned.
So instead of making everything in the component use fetch(), that particular case can just use doc functions. Does that make sense?
Once that's done, I think we can merge this in.

@maxkfranz I see that the service that I am fetching is calling ``getPubmedArticle()`` which is also in server side. I do not know yet but there may be other similar things as well. Therefore, I think that I may not be able to put the same functionality under the ``Document`` model. What do you think?

maxkfranz:
That's a good point.  Let's go with the PR as-is, and we can reevaluate those details later if need be.

maxkfranz (approved):

src/client/components/editor/title.js
metincansiper: I actually set the default value for the title as ``null`` instead of ``Untitled`` to better understand if it is actually set (https://github.com/PathwayCommons/factoid/pull/913/files#diff-bbf58575632e23c789ad93390b6495ff736ac319e4ee5e2473416a83e1bd03caR230).

Issue #914, opened 2020-12-07
Update Cron: Fail gently


jvwong:
Background: Factoid has a cron service that dispatches signals (weekly) to retrieve/refresh information. In particular, one update refreshes PubMed metadata for documents and refreshes the related papers attribute from PubMed/INDRA etc.   

Issue: It appears that the results of the update jobs are stored, regardless of errors in information retrieval. The result is documents with no paper metadata, even those that previously had information. 

Propose: Fail gently, don't overwrite existing information if an update call fails.  

I've noticed this problem once before but wasn't sure if it was just a weird abberration.

jvwong:
Also: expose env `CRON_SCHEDULE` in docker so we can always reconfigure.

PR #916, opened 2020-12-09
Update paper metadata: Fail gently

jvwong:
Two aspects that get set/refreshed are 1. article metadata (abstract, authors etc...) 2. related papers (document-level, which includes related papers and referenced papers; network-level). 

In this PR, refresh or update can only overwrite the existing information if there a. is no error and b. the previous saved result was null or undefined. Otherwise the previous information stays put.

The trade-off here, is that if you first entered a valid PMID and all the metadata got set, you can't overwrite the paper with a non-recognized PMID/title, although it gets stashed in the `provided.paperId` field. This is a weird case anyways.

Refs #914

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -29,6 +29,7 @@ services:
       NCBI_EUTILS_API_KEY:
       NODE_OPTIONS:
       EMAIL_RELPPRS_CONTACT:
+      CRON_SCHEDULE:
   grounding:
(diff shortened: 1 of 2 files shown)


maxkfranz:
Great. 

 In future, it would be good to reevaluate (b) as we run into edge cases.  The approach in this PR is good for now, but there may be cases where you want to overwrite existing data (e.g. the article was amended because it was preprint).

This should handle the most important cases, so let's merge this in.

Issue #925, opened 2021-01-06
One-click editor link for related papers emails


maxkfranz:
## Description

**Q: What is the name of the feature?**

A: One-click editor link for related papers emails

**Q: What does this feature enable the user to do?**

A: When a prospective user receives a related papers email, they can start creating a factoid right away by clicking a link in the email.

**Q: What information must the user provide to use the feature?**

A: The user must specify the paper associated with their factoid, within the editor.

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: The submission validation must check that the paper's title is specified for the factoid.

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: A prospective user who learns about Biofactoid by email can create a new article profile by just clicking the CTA.
- Editor: N/A
- Computational biologist: N/A
- Curator: N/A

### Mockup

Here is an example of the email, as it stands today:

> Dear John Doe,
> 
> Biofactoid (biofactoid.org) is an academic project whose mission is to connect biological research. Biofactoid digitally captures the molecular interactions within articles, and it connects these articles to social media, relevant literature, and widely-used biological research apps. Biofactoid empowers researchers to discover novel findings and to connect with other researchers.
> 
> A new article profile has been published that cites your article, Lorem ipsum.
> 
> J. Schmoe et al. have published Dolor sit amet (A. Journal), where A binds to B.
> 
> Authors use Biofactoid to create interactive profiles of the discoveries in their research, and you can explore the article profile that J. Schmoe et al. have just created on Biofactoid.
> 
> Connect your own findings with other researchers by adding your own article profile to Biofactoid. It takes only a few minutes, and once added, your article will be discoverable, shareable, and app-accessible. Connect your research to the world by creating your article profile today!
> 
> The Biofactoid Team
> Bader Lab at the University of Toronto
> Sander Lab at Harvard
> The Pathway and Omics Lab at the University of Oregon

The sentence, "Connect your research to the world by creating your article profile today!", is currently a CTA that links to the Biofactoid home page.  This link would be replaced with a link to an empty document that the user can edit right away.  Each link must be unique per email.

### Details

- [x] Change the link in the email template.  The link should be passed to the template.
- [x] The server should generate a new empty document with corresponding editor links for each related papers email that is sent.

Ref.

- Novel interaction notification #858
- Post hoc specification of the paper in a factoid #901

jvwong:
[...]

PR #921, opened 2020-12-21
Post hoc titles etc

maxkfranz:
For details, refer to the commits.  The main user-facing changes include:

(1) The name is another field on the landing page CTA:

Screen Shot 2020-12-21 at 11 22 40

(2) There is a link to set the title in the editor:

Screen Shot 2020-12-21 at 11 25 24

(3) The title is hooked into the validation:

Screen Shot 2020-12-21 at 11 25 32

(4) For use cases where the author name and email are pre-filled (e.g. email campaign #858), the user just needs to set the title:

Screen Shot 2020-12-21 at 11 25 47

(5) When the title has been set, it looks like the normal editor title:

Screen Shot 2020-12-21 at 11 25 55

(6) The explore view has formatted author links to ORCID.  In the top-right corner, the contributing author is credited.

Screen Shot 2020-12-21 at 11 26 14

Builds on Metin's PR, #913.

Ref. : 

- Post hoc specification of the paper in a factoid #901
- Novel interaction notification #858

PR #932, opened 2021-02-08
One-click editor link for related papers emails

maxkfranz:
- Create empty doc when sending related papers email.
- Send editor link in template to related papers email.
- The email template needs to be updated to use the `editorUrl` variable.

Ref. : One-click editor link for related papers emails #925

### DIFF
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -1450,19 +1450,35 @@ const emailRelatedPaperAuthors = async doc => {
 
   const hasContact = paper => getContact(paper) != null;
 
+  const createEmptyDoc = async (authorName, authorEmail) => {
+    const provided = {
+      authorEmail,
+      authorName
+    };
+
+    const doc = await postDoc(provided);
+
+    return doc;
+  };
+
   const sendEmail = async (paper, novelIntns) => {
     const contact = getContact(paper);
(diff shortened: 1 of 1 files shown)


src/server/routes/api/document/index.js
```diff
@@ -1450,19 +1450,35 @@ const emailRelatedPaperAuthors = async doc => {
 
   const hasContact = paper => getContact(paper) != null;
 
+  const createEmptyDoc = async (authorName, authorEmail) => {
+    const provided = {
+      authorEmail,
```
jvwong, line 1455: We need the `paperId` set here or else `postDoc` complains threre's not PubMed article found.

jvwong (changes requested):

src/server/routes/api/document/index.js
maxkfranz: I think hat `postDoc()`, if it doesn't already, should be allowed to fail in part or in whole.  There is no paper ID for this empty doc as per: 

- One-click editor link for related papers emails #925
- Post hoc specification of the paper in a factoid #901
- Post hoc titles etc #921
- Refactored RequestForm as a reusable component. #913

jvwong: Ah yes, empty doc my bad

jvwong (approved):
OK works for me

Issue #836, opened 2020-09-21
PC-2-factoid: Supplementing the Factoid database with curated interactions


jvwong:
## Description

**Q: What is the name of the feature?**

A: Add to Factoid database documents, created from an external (existing) database 

**Q: What does this feature enable the user to do?**

A: Access more (expert-)curated interactions supported by literature via Factoid 

**Q: What are the applicable constraints, e.g. compatibility or performance?**

- General criteria for selecting external datasource to create Factoid documents: 
  - Interactions tied directly to publication (pmid)
  - Interactions supported by (low-throughput) evidence
  - Interaction types that are in-line with the Factoid model 
  - Data available, in usable format

**Q: How does this feature affect each class of user (persona)?**

There are a few benefits to introducing external data into Factoid database:

  a) 'Frame-out' the Factoid app with useful, quality, curated data
  b) A lot is data for older articles that wouldn’t otherwise be curated
  c) Can serve as pre-made Factoids that we can ask the authors to ‘verify’
  d) Begins the idea of unifying PC and Factoid (search/import)
 
## Specification

- Input
  - Datasource: PhosphositePlus looks like it might fit the general selection criteria (above)
- Output
  - For each PMID, a representation of a Factoid document 
  - Compatible with batch entry into Factoid

### Details

1. Fetch data file

My first thought is to use [PhosphositePlus (PSP) PC BioPAX](https://www.pathwaycommons.org/archives/PC2/v12/PathwayCommons12.psp.BIOPAX.owl.gz)

2. Filter Interactions

- By experimental evidence
  - PSP provides information about the type of experimental procedure used to verify the interaction and we would want only keep those referencing with 'Low-throughput papers (LTP)' approaches
    - i.e. `bp:Evidence` with `bp:evidenceCode` one of
      - mutation analysis (MI:0074)
      - identification by antibody (MI:0421)
      - western blot (MI:0113)

3. Group interactions

Factoid documents are one-to-one with articles, and interactions are added to an article/document. So in this case, interactions would have to be grouped under the same PMID. Not sure how many PSP interactions fall under the same PMID, and I wouldn't merge genes/interactions in output

4. Map to document JSON

- Use the document API for import: `doc.fromJson(obj)`
- Refer to the JSON format from the files in the zip download as a reference: https://biofactoid.org/api/document/zip
- Main fields are
  - `id` : new uuid
  - `elements`: new uuid
  - `secret` (might be generated for you)
- Create a new document with POST /api/document
[...]

PR #940, opened 2021-02-22
Supplement db improvements

metincansiper:
Refs #836 and works together with https://github.com/PathwayCommons/factoid-converters/pull/23
- Added new error handlings
- Created a ``from-url`` end point in server side and directly called it from client side
- When an entity does not have a valid organism/grounding ignored it along with its interaction and the entity that is at the other end of the interaction
- Make the queries for pubmed ids one by one
- Limits the number of elements for which related papers will be queried at a time
- Use queues to handle related paper updates (both for admin and pc url cases)


maxkfranz:
@metincansiper.  Thanks for the update.  I'll review in more depth over the next couple of days.  In the meanwhile, would you be able to look into the merge conflict?

metincansiper:
@maxkfranz I resolved the merge conflicts.

src/server/routes/api/document/index.js
```diff
@@ -1734,6 +1762,75 @@ const postDoc = provided => {
...
+  const { url } = provided;
+
+  tryPromise( () => getJsonFromBiopaxUrl( url ) )
+    .then( response => response.json() )
+    .then( docsJSON => new Promise( () => {
+      let pmids = Object.keys( docsJSON );
+      pmids = pmids.filter( pmid => pmid != null );
+      pmids = _.slice( pmids, 0, 2 );
+      // TODO: which email address?
+      let authorEmail = 'pc@biofactoid.com';
```
maxkfranz, line 1802: Maybe use an env var here.  That way we can decide on the email address later without code changes

metincansiper: @maxkfranz okay I set it through an env variable

metincansiper:
>When an entity does not have a valid organism/grounding ignored it along with its interaction and the entity that is at the other end of the interaction

This is one of the things that I mentioned in the PR description. It may be important to decide whether to handle it in thw way I did or another way.

jvwong:
I see an error when I try input `https://www.pathwaycommons.org/archives/PC2/v12/PathwayCommons12.psp.BIOPAX.owl.gz`:
```
error:    FetchError: request to http://localhost:3011/get failed, reason: socket hang up
...
```
 which points to my local grounding service at port 3011

metincansiper:
>I see an error when I try input https://www.pathwaycommons.org/archives/PC2/v12/PathwayCommons12.psp.BIOPAX.owl.gz:

@jvwong would you provide the pmid of document for which that happens? Also, I wonder if that error was caught properly or blocked the server?

jvwong:
> > I see an error when I try input https://www.pathwaycommons.org/archives/PC2/v12/PathwayCommons12.psp.BIOPAX.owl.gz:
> 
> @jvwong would you provide the pmid of document for which that happens? Also, I wonder if that error was caught properly or blocked the server?

I tried again this morning and now I'm getting some 404 GATEWAY TIMEOUT from INDRA very early on. 

```
error:   Error for https://db.indra.bio/latest/statements/from_agents?format=json-js&agent0=CAMK2A%40TEXT&agent1=CD44%40TEXT
error:   504: Gateway Timeout
```

metincansiper:
@maxkfranz @jvwong I see that this line (https://github.com/PathwayCommons/factoid/blob/unstable/src/server/routes/api/document/index.js#L1748) makes server not waiting for updating related papers on new document requests. I wanted to make sure that it is done on purpose. And if so should it work like that while creating new documents from a biopax file as well?

metincansiper:
>@maxkfranz @jvwong I see that this line (https://github.com/PathwayCommons/factoid/blob/unstable/src/server/routes/api/document/index.js#L1748) makes server not waiting for updating related papers on new document requests. I wanted to make sure that it is done on purpose. And if so should it work like that while creating new documents from a biopax file as well?

To give more info about this: the early 404 GATEWAY TIMEOUT error that @jvwong get must be because I did not realize that the related function is not waiting for updating related papers before returning. It made a lot of related papers search being on work simultaneously and caused a lot of indra queries being active at a time. I assume that it is done on purpose in the regular case of updating related papers when a single document is created by a user.

I wonder is there any preference on case where multiple documents are created through biopax file. I think it will affect whether the papers must be updated or not before the operation is accepted to be done.

I think I can handle any of these cases based on preference.

maxkfranz:
One approach would be to handle the BioPAX documents separately in a queue:

- Add document from BioPAX.  Keep the document state as non-submitted.
- Add the new document to a queue.
- On the server, in the background, a document is taken out of the queue at a regular interval (e.g. every minute).  Calculate the related papers info for the dequeued doc.
- Once the related papers calculation has completed, mark the document as submitted.
- When marking the document as submitted, make sure that the doc does not generate any emails.  BioPAX papers aren't submitted by individual authors, so no emails are necessary.  Further, generating emails for BioPAX papers significantly increases the volume of emails we send:  This could cause problems with having our emails marked as spam.

metincansiper:
>One approach would be to handle the BioPAX documents separately in a queue:

@maxkfranz queueing was one of the options I was considering in case where we do not need to wait related paper updates to point the user that the process is done. Therefore, I guess you are preferring not waiting for related paper updates as a part of process. Am I right?

However, one thing confusing me is that why we want to make the submissions after related papers are updated. Currently it is done before requesting the related papers updates. If we do it after related papers updates when we signal the client about that the db supplementation is done the documents would not actually be submitted. Also normally we are triggering related paper calls after submission of a paper by a user. It may be contradicting. Am I wrong?

Also, since we are thinking about the querying of related papers jobs we can extend it to single document submission cases as well. I think it would be needed since there may be cases where a user want to submit a document while there are related papers works done in server because of a db supplementation work in process or because of other users document submissions (We may still want to prioritize the user submissions though). How does that sound?

maxkfranz:
The cases of an automatically-generated doc (from BioPAX) and a manually-created doc (from an author) are distinct in terms of UX:

For an author's doc, we want to get the related papers right away.  We don't want the user to be waiting a long time for the related papers to be populated.  The user may want to share their article's summary (explorer view), and having a long spinner is unsatisfying.  At the same time, we don't want to block the user from seeing their article summary, so the related papers querying happens in the background.  The user can see their summary right away, even though the related papers can take 10-30 seconds to complete.

For a BioPAX doc, there is no user.  Nobody needs to see the doc right away, since there is no user pressing the submit button -- expecting to see the summary right after hitting the button.  So in this case, you can wait to publish/submit the doc until the related papers are done.  This isn't a strict requirement, but it's a nicer UX to see these docs fully completed.

You could put all of the related papers requests into a single queue, as you mentioned.  However, you also need to prioritise author-created docs.  So either you need two queues or a single priority queue -- with author docs always top priority.

maxkfranz:
Another consideration is how BIoPAX imports will run:  If they run offline, as they may initially, then you don't care as much about the timing of the submission status and the related papers querying.  You do all of the importing in a local instance, and then manually import those docs into the production DB.  You still need queuing in order to avoid API limits.

It's still nice to have the submission status changed only after the related papers are complete, because you can more easily perform the manual import from the offline instance.  You probably don't want to import docs into production if they have missing/broken related papers lists.  Further, this approach makes it a smoother transition to a future online approach where you just point a production instance at some BioPAX source and it queues everything up in place.

metincansiper:
@maxkfranz @jvwong I made the discussed updates.

metincansiper:
It may be hard to run it for every document in my local server but I can run it for 100 documents and let you know about the results. How about that?

metincansiper:
@maxkfranz @jvwong I see that it works fine for the first 100 documents from pc file. I realized the following thigs:

- I have seen a few cases where an entity ended up with no grounding. In the cases where there is no valid xref coming from biopax file (either no xref or xref with unsupported organism) I omitted the entity along with its interaction and the entity at the other end of interaction. I think that the cases where I see no xref at the end must be because grounding search found no result for the entity. I guess I may need to apply the same logic for such cases as well. Am I right? Would it make more sense to remove the entity along with the interaction and the other end of interaction (as I am currently doing for the case of invalid or no grounding from pc) or should I just remove the enitity itself (and so interaction as well but not the other end of interaction)?
- I see that there are some documents where there is no entity is shown. I suspect that it must be because of that all the entities are removed because of grounding issues. Should I skip such documents?

maxkfranz:
What proportion of documents do you expect would have excluded entities?  If the percent is small, then maybe those documents should just be excluded altogether.

jvwong:
@metincansiper how many documents are in that psp biopax?

metincansiper:
>@metincansiper how many documents are in that psp biopax?
@jvwong there are 16418 documents.

metincansiper:
>What proportion of documents do you expect would have excluded entities? If the percent is small, then maybe those documents should just be excluded altogether.

@maxkfranz I will try to figure it out. They may be excluded because of grounding info or because the grounding service just did not give any result back. Considering the no result cases it would be a little harder to detect.

metincansiper:
>What proportion of documents do you expect would have excluded entities? If the percent is small, then maybe those documents should just be excluded altogether.

@maxkfranz I made some updates on code to detect it faster and still running it. However I think I can tell that it is too much. So far it is 1109 over 3056 (I can write the final result as well later).

Also, I detected some new things:

- I already mentioned that in some cases grounding results are not found even though the xref is valid. To be clearer in such cases the first grounding search call that we make to find the corresponding ncbi id yields no ncbi id.
- I again see some exceptions from grounding search service for the entities whose organisms are supported by grounding search service. Examples are A9UF07 (human) and P34152-2 (Mouse) for uniprot. I also see one case (P49137 for uniprot) that did not work when the server was querying that but worked for my independent query later. I suspect some queries may be failing by chance/time. 
- Weirdly an entity with rabbit organism passed through my filter. If I am not wrong rabbit is not one of the supported organisms. I will take a look at why that happened.

jvwong:
>     * I already mentioned that in some cases grounding results are not found even though the xref is valid. To be clearer in such cases the first grounding search call that we make to find the corresponding ncbi id yields no ncbi id.

Can you provide a couple of examples? This sounds interesting..

metincansiper:
>I made some updates on code to detect it faster and still running it. However I think I can tell that it is too much. So far it is 1109 over 3056 (I can write the final result as well later).

My battery run out at some point last night so it did not run for all docs yet but it is already 3675/10262 which I think high enough so I am killing the process.

metincansiper:
>Can you provide a couple of examples? This sounds interesting..

@jvwong some examples are below:

```
{ db: 'uniprot', id: 'P27791', org: '10116' }
{ db: 'uniprot', id: 'P08414', org: '10090' }
{ db: 'uniprot', id: 'P06494', org: '10116' }
{ db: 'uniprot', id: 'P05696', org: '10116' }
```

Actually the grounding service returns a result with some ``dbXrefs`` but for none of them ``xref.db`` is ``GeneID``.

metincansiper:
>What proportion of documents do you expect would have excluded entities? If the percent is small, then maybe those documents should just be excluded altogether.

>My battery run out at some point last night so it did not run for all docs yet but it is already 3675/10262 which I think high enough so I am killing the process.

Based on this I think we can eliminate the option of totally excluding the documents (Am I right?). I think we must choose one of the other options.

jvwong:
> > Can you provide a couple of examples? This sounds interesting..
> 
> @jvwong some examples are below:
> 
> ```
> { db: 'uniprot', id: 'P27791', org: '10116' }
> { db: 'uniprot', id: 'P08414', org: '10090' }
> { db: 'uniprot', id: 'P06494', org: '10116' }
> { db: 'uniprot', id: 'P05696', org: '10116' }
> ```
> 
> Actually the grounding service returns a result with some `dbXrefs` but for none of them `xref.db` is `GeneID`.

Strangely UniProt doesn't have those mappings. 

I tried a mapping app here called [Synergizer](http://llama.mshri.on.ca/synergizer/translate/). It seems to work pretty good but you have to indicate the species in english words:, e.g. 

POST to `http://llama.mshri.on.ca/cgi/synergizer/serv`
```json
{
	"method":"translate",
	"params":[{
		"authority":"ncbi2015",
		"species":"Rattus norvegicus",
		"domain":"uniprot",
		"range":"entrezgene",
		"ids":[
			"P27791",
			"P06494",
			"P05696"
			]
	}],
	"id":0
}
```

and the response is:

```json
{
  "error": null,
  "id": 0,
  "result": [
    [
      "P27791",
      "25636"
    ],
    [
      "P06494",
      "24337"
    ],
    [
      "P05696",
      "24680"
    ]
  ]
}
```

jvwong:
Just an aside for future reference: 

I suppose we should view our 'map' service as a quick and easy way to try to get an ID. It relies on UniProt and NCBI curating those links, and there are clearly many cases (thousands) where it fails. I tried some other tools (g:Convert, Synergizer) and looked around at NCBI files. There's no perfect solutions.

It's still preferable that we have the NCBI Gene ID because it gives us so many links between the databases.

metincansiper:
>I suppose we should view our 'map' service as a quick and easy way to try to get an ID. It relies on UniProt and NCBI curating those links, and there are clearly many cases (thousands) where it fails. I tried some other tools (g:Convert, Synergizer) and looked around at NCBI files. There's no perfect solutions.

I know the failing cases you mentioned are not the ones that I mentioned here. However, just to make sure we are at the same page I want to tell that the failing cases are reported are not all about this problem. Some of them are the cases where entity does not have a valid xref. However, I did not check the distribution of them.

metincansiper:
>I suppose we should view our 'map' service as a quick and easy way to try to get an ID. It relies on UniProt and NCBI curating those links, and there are clearly many cases (thousands) where it fails. I tried some other tools (g:Convert, Synergizer) and looked around at NCBI files. There's no perfect solutions.

I know the failing cases you mentioned are not the ones that I mentioned here. However, just to make sure we are at the same page I want to tell that the failing cases are reported are not all about this problem. Some of them are the cases where entity does not have a valid xref. However, I did not check the distribution of them.

metincansiper:
>Just an aside for future reference:

I suppose we should view our 'map' service as a quick and easy way to try to get an ID. It relies on UniProt and NCBI curating those links, and there are clearly many cases (thousands) where it fails. I tried some other tools (g:Convert, Synergizer) and looked around at NCBI files. There's no perfect solutions.

It's still preferable that we have the NCBI Gene ID because it gives us so many links between the databases.

@jvwong I guess the example that you provided converts uniprot ids into ncbi ids. Am I right?

jvwong:
> It's still preferable that we have the NCBI Gene ID because it gives us so many links between the databases.
> 
> @jvwong I guess the example that you provided converts uniprot ids into ncbi ids. Am I right?

Yes, the service was created by a lab now at U of T. I don't know how often they update it but it seems to work. However, you need to supply their specific parameters in the HTTP POST e.g. species is "Rattus norvegicus" but I guess its in their docs

metincansiper:
>Yes, the service was created by a lab now at U of T. I don't know how often they update it but it seems to work. However, you need to supply their specific parameters in the HTTP POST e.g. species is "Rattus norvegicus" but I guess its in their docs

Okay I will use it instead of grounding search service while getting ncbi id from uniprot id. 
I think the input for that parameter can be accessed from the Organism file (https://github.com/PathwayCommons/factoid/blob/unstable/src/model/organism.js#L421)

jvwong:
> > Yes, the service was created by a lab now at U of T. I don't know how often they update it but it seems to work. However, you need to supply their specific parameters in the HTTP POST e.g. species is "Rattus norvegicus" but I guess its in their docs
> 
> Okay I will use it instead of grounding search service while getting ncbi id from uniprot id.
> I think the input for that parameter can be accessed from the Organism file (https://github.com/PathwayCommons/factoid/blob/unstable/src/model/organism.js#L421)

Do you have any idea how many cases the grounding-search `/map` is failing to return a valid UniProt --> NCBI Gene id?

metincansiper:
>Do you have any idea how many cases the grounding-search /map is failing to return a valid UniProt --> NCBI Gene id?

@jvwong I am running something to detect it. For now the numbers are as below:
```
numProblems:  425
numEntities:  24986
numUniqueProblems:  269
numUniqueEntities:  9588
```
There are repeating xrefs therefore I counted number of total (``numProblems``, ``numEntities``) and unique entities (``numUniqueProblems``, ``numUniqueEntities``). By "problem" I mean the ones yielding no result.

jvwong:
> > Do you have any idea how many cases the grounding-search /map is failing to return a valid UniProt --> NCBI Gene id?
> 
> @jvwong I am running something to detect it. For now the numbers are as below:
> 
> ```
> numProblems:  425
> numEntities:  24986
> numUniqueProblems:  269
> numUniqueEntities:  9588
> ```

269 / 9588 ~= 2.2% which is pretty good i think? But you're saying above ...

> Some of them are the cases where entity does not have a valid xref. 

If missing xref is the problem then there's not much we can do - just drop the interaction.  If that leads to any empty doc then just drop the entire doc I guess?

metincansiper:
>269 / 9588 ~= 2.8% which is pretty good i think? But you're saying above ...

@jvwong I did not include the missing xrefs in this calculations. Currently the numers are:

```
numProblems:  1219
numEntities:  81405
numUniqueProblems:  716
numUniqueEntities:  32789
```
Then should we stick on grounding search service or switch to the one that you mention? Or maybe try grounding search first and try the other one when grounding search yields no result?

jvwong:
> > 269 / 9588 ~= 2.8% which is pretty good i think? But you're saying above ...
> 
> @jvwong I did not include the missing xrefs in this calculations. Currently the numers are:
> 
> ```
> numProblems:  1219
> numEntities:  81405
> numUniqueProblems:  716
> numUniqueEntities:  32789
> ```
> 
> Then should we stick on grounding search service or switch to the one that you mention? Or maybe try grounding search first and try the other one when grounding search yields no result?

If I understand you correctly, isn't a 'missing xref' impossible to assign to an id?

metincansiper:
>If I understand you correctly, isn't a 'missing xref' impossible to assign to an id?

Yes, I cannot query when xref is missing. If you mean that yes I already cannot call grounding service when the xref is missing.

jvwong:
> > If I understand you correctly, isn't a 'missing xref' impossible to assign to an id?
> 
> Yes, I cannot query when xref is missing. If you mean that yes I already cannot call grounding service when the xref is missing.

If the failures of the `/map` are only ~2.8%, and the case where an xref is missing is a 'dead end', then I think we should just keep the system as is and:
-  drop any interactions with an ungrounded participant
- drop any empty document

metincansiper:
>If the failures of the /map are only ~2.8%, and the case where an xref is missing is a 'dead end', then I think we should just keep the system as is and:

Okay that make sense. I do not think it would effect our approach on that but I am telling just to remind there is also one case where xref exists but not valid/supported.

metincansiper:
When I stopped the running process the latest numbers were (just to update):

```
numProblems:  2211
numEntities:  172661
numUniqueProblems:  1307
numUniqueEntities:  69663
```

metincansiper:
>drop any interactions with an ungrounded participant
>drop any empty document

@jvwong @maxkfranz When I looked into the grounding search code I remembered that in the past we decided to drop the whole document when an entity has invalid xref. However, I as far as I can remember (not so sure) by that time we only checked the entities where there was no xref at all or just has unhandled xref db. As far as I remember that number was more negligible (again not sure).

Right now we figured out a lot more cases and the number of problematic situations (like unsupported org etc.) that increased the number of number of problematic entities. Therefore, I think the previous rule is no more valid and I must apply the one that @jvwong mentioned:

```
drop any interactions with an ungrounded participant
drop any empty document
```

metincansiper:
>drop any empty document

@maxkfranz @jvwong while dropping such documents I am using ``doc.trash()`` is that a convenient way? If not is there any way I can totally remove a document in Factoid? If really needed I can try to run and store grounding search results earlier so that I would be able to ignore such docs before creating them but it does not look ideal considering implementation.

jvwong:
> > drop any empty document
> 
> @maxkfranz @jvwong while dropping such documents I am using `doc.trash()`. Is that a convenient way? If not is there any way I can totally remove a document in Factoid? If really needed I can try to run and store grounding search results earlier so that I would be able to ignore such docs before creating them but it does not look ideal considering implementation.

There is an endpoint `/api/document`

https://github.com/PathwayCommons/factoid/blob/4aed3ee7bdcdea0dd1dd5c34e8c5877b4cd902d0/src/server/routes/api/document/index.js#L1576

caveat: `deleteTableRows` doesn't look like it drop the secret from the factoid.secret table

metincansiper:
>There is an endpoint /api/document

Okay, I reused ``deleteTableRows`` function and then called another function I implemented to delete the document secret from secret table. I think it can be reviewed now.

metincansiper:
I forgot to mention it would be better to run the latest code here with the updates from https://github.com/PathwayCommons/factoid-converters/pull/25

jvwong:
> I forgot to mention it would be better to run the latest code here with the updates from [PathwayCommons/factoid-converters#25](https://github.com/PathwayCommons/factoid-converters/pull/25)

OK , I'm running it on master, anyways, so far so good....

metincansiper:
>OK , I'm running it on master, anyways, so far so good....

I think running with master must be okay.

jvwong (approved):
This works for me and I think it should be merged in. 

The caveats are:
  - Should always run this on a non-production instance, 'offline'
  - Lots of edge cases I'm seeing to be taken care of in subsequent PRs

Issue #951, opened 2021-03-17
Docker: Static assets reflect the environment at build-time, not run time 


jvwong:
The Dockerfile instructs to execute an `npm run build` which takes into account the default config variables at that time. When the container is run, it ingests environmental variables that overwrite these defaults, at least on the server. However, the client-side assets are already fixed and do not reflect these new values.

Example of how this manifests:

The Explore button link in the editor submit-success popup always links to `http://biofactoid.org/<id>`. It should be configurable via environmental variable. The email I get has the correct link.

Reproduce:

I notice this on Docker Compose:
- set BASE_URL=http://master.factoid.baderlab.org in the docker-compose.yml file (i.e the .env file)

e.g. try https://master.factoid.baderlab.org

jvwong:
Source of the problem: Image at build-time uses default env variables for client-side static assets 

https://github.com/PathwayCommons/factoid/blob/4aed3ee7bdcdea0dd1dd5c34e8c5877b4cd902d0/Dockerfile#L36

PR #952, opened 2021-03-22
Docker: environment variables in client-side assets

jvwong:
Move the `npm build command` to the entrypoint.sh file, so that the app assets are built under the provided environment.

Refs #951

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,10 +31,6 @@ RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key
 # Note: NODE_ENV is development so that dev deps are installed
 RUN NODE_ENV=development npm install
 
-# Build project
-RUN npm run clean
-RUN npm run build
-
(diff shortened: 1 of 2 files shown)


maxkfranz (approved):
If this is blocking your efforts, then let's just merge this as-is.

This approach results in the built image lacking dependencies, which means that the deployment is less reproducible.  The risk is slight, but it does go against the purpose of docker a bit.  We should note this detail in the deployment documentation in case it becomes a problem later on.

jvwong:
The only obvious alternative to me is we fix those variable (e.g. 'master.factoid.baderlab.org') at build using GitHub runner and push that to the Docker Hub as a branch. The trade-off is we're fixing the URL and we gotta build these images.

maxkfranz:
Fair enough.  The perfect is the enemy of the good.  I think as long as we log the limitations in our docker repo, then it's fine.  It's not worth the effort for now as long as things are working.

Issue #969, opened 2021-05-05
Grounding: The data source name should be made clear to user


jvwong:
This issue relates to the editor node tool-tip association information that is displayed. We provide this information to help users editing their document decide if we have correctly matched the top database hit from the grounding search.   

Background: Currently, the user input is reused as the entity name of the top hit. The reasoning behind this decision is that it can sometimes be confusing to users if the grounding database provides an odd or not commonly used 'name' (which we use in the entity search results 2 onwards). One example is estrogen receptor, almost universally referred to as 'ER' has an offiical name 'ESR1' ([NCBI Gene 2099](https://identifiers.org/ncbigene:2099)).

<img width="350" alt="image" src="https://user-images.githubusercontent.com/4706307/117188070-a3dac080-adaa-11eb-94cc-45075e99f39f.png">

Issue: This is a case where the ground is clearly incorrect, but the database name is informative, but nonetheless discarded. 

<img width="350" alt="image" src="https://user-images.githubusercontent.com/4706307/117188223-d1276e80-adaa-11eb-892d-3ebc636943f3.png">
 
In the latter case, the ChEBI name is [`Soman`](https://www.ebi.ac.uk/chebi/searchId.do?chebiId=CHEBI:9195) which would be helpful for the user to have, or would at least give them pause.

jvwong:
Simple approach is to include Official Name/CheEBI Name in the top hit under 'synonyms'.

<img src="https://user-images.githubusercontent.com/4706307/117193129-7ee94c00-adb0-11eb-9560-b507e020c69f.png"  width="350px" />

maxkfranz:
Yeah, I think we discussed that option at some point, but it wasn’t a priority since we didn’t have compelling example entities.

More specifically, the rule may be: If the typed name is not the same as the official name, then we add the official name to the start of the synonym list.  This avoids duplication when there’s an exact match.

PR #976, opened 2021-05-11
Retain the data source's 'name' in grounding tooltip entry

jvwong:
If the user label is different than the search hit selected, stash the data provider 'name' (e.g. Official, ChEBI) in the synonyms list.

For example, below, the name 'Carotol' is kept as a synonym.

image

Refs #969

### DIFF
--- a/src/client/components/element-info/entity-info.js
+++ b/src/client/components/element-info/entity-info.js
@@ -148,6 +148,21 @@ class EntityInfo extends DataComponent {
       delete match.typeOfGene; // n.b. the model overrides via this field
     }
 
+    if( s.name !== match.name ) {
+      // Add 'name' to shortSynonyms
+      let shortSyns = _.get( match, 'shortSynonyms', [] );
+      let newShortSyns = _.uniq( _.concat( match.name, shortSyns ) );
+      _.set( match, 'shortSynonyms', newShortSyns );
+    }
+
+    const assoc = el.association();
+    if( assoc ) {
+      // Drop 'name' from shortSynonyms
+      let shortSyns = _.get( assoc, 'shortSynonyms', [] );
+      let newShortSyns = _.without( shortSyns, assoc.name );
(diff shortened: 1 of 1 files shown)


src/client/components/element-info/entity-info.js
```diff
@@ -148,6 +148,23 @@ class EntityInfo extends DataComponent {
       delete match.typeOfGene; // n.b. the model overrides via this field
     }
 
+    // If label is different from match 'name', add latter to shortSynonyms
+    const label = s.name;
+    let matchShortSyns = _.get( match, 'shortSynonyms', [] );
+    let inputNotMatchName = label !== match.name;
```
maxkfranz, line 154: Should this be case insensitive?

src/client/components/element-info/entity-info.js
```diff
@@ -148,6 +148,23 @@ class EntityInfo extends DataComponent {
...
+    if ( inputNotMatchName ) {
+      let newShortSyns = _.uniq( _.concat( match.name, matchShortSyns ) );
+      _.set( match, 'shortSynonyms', newShortSyns );
+    }
+
+    // Remove the match 'name' from the current association, if exists
+    const assoc = el.association();
```
maxkfranz, line 161: Since the association will be replaced on line `el.associate( match );`, should this instead modify `match`?

maxkfranz (commented):
Looks good.  There are some edge cases to consider:

src/client/components/element-info/entity-info.js
jvwong: My thinking was: `assoc` will be unassociated when the `match` is associated with `el`, so you need to reset it to its original state (drop the name from short synonyms).

maxkfranz: That makes sense.  I don't think an association is ever reused, though, so it probably doesn't matter.  As long as this wasn't intended for the incoming association, then I think it's good to merge.

maxkfranz (approved):

Issue #863, opened 2020-10-07
Browsing: Beyond the carousel


jvwong:
## Description

**Q: What is the name of the feature?**

A: Browse (beyond the carousel)

**Q: What does this feature enable the user to do?**

A: Access any public document via homepage

**Q: What information must the user provide to use the feature?**

A: N/A

**Q: How does this feature affect each class of user (persona)?**

A:  All would benefit from browsing and exploration 

### Mockup

This is a common pattern now.  An popular example is [Pinterest](https://www.pinterest.ca/search/pins/?q=french%20toast&rs=typed) -- an infinite scroll grid and a search bar.

### Details

- Grid UI
  - [react-infinite-scroll](https://www.npmjs.com/package/react-infinite-scroll-component)
  - CSS grid or CSS flex box
- Search has several options for the back end:
  - [Google custom search JSON API](https://developers.google.com/custom-search/v1/introduction)
  - [rethinkdb log stash](logstash-input-rethinkdb)
  - [rethinkdb elastic search sync](https://rethinkdb.com/docs/elasticsearch/)
  - Use elastic search manually
    - Rebuild the index when you start the factoid server
    - Add to index on submit

maxkfranz:
I've fleshed out the issue description more, including links for reference.

maxkfranz:
Note that the ideal version of this would be to delegate to Google rather than to have our own page.  Google search results are much more discoverable because most people start from Google.  Only users who already know about Biofactoid would be able to use a search page that we host  ourselves.

The CTA button here could be replaced with a button that starts a Google search: 

![Screen Shot 2020-10-13 at 10 20 46 AM](https://user-images.githubusercontent.com/989043/95873336-d43f0c80-0d3d-11eb-990d-8ec14af46b9c.png)

A search link could be added to the navigation items at the top of the page (i.e. beside 'contact', 'twitter').  Further, a search button could be added alongside the existing 'add my article' button:

![Screen Shot 2020-10-13 at 10 22 54 AM](https://user-images.githubusercontent.com/989043/95873637-1c5e2f00-0d3e-11eb-899c-9e7a55441283.png)

jvwong:
Yeah I agree. This is related to 'Index images via sitemap' in https://github.com/PathwayCommons/factoid/issues/842#issuecomment-707850543. 

Also we should be mindful of when this sort of thing should be handled by [PC SIFGraph](https://www.pathwaycommons.org/sifgraph/swagger-ui.html)

[...]

PR #984, opened 2021-06-08
A class for client-side Document search

jvwong:
A DocumentSearch class that allows you to index and search Document objects.

Doc fields indexed:
 - text
 - citation title, authors abbreviation, reference, abstract
 - entity names and synonyms

Uses minisearch under the hood.

Refs #863


jvwong:
FYI: https://lucaong.github.io/minisearch/classes/_minisearch_.minisearch.html

maxkfranz:
Do you think it would be useful to also include the interaction text, so you can search for 'A phosphorylates B'?

jvwong:
> Do you think it would be useful to also include the interaction text, so you can search for 'A phosphorylates B'?

I index the document `text` field - isn't that it? I don't think interactions have this, but they do have a `type`/`association` (e.g. binding).

src/client/document-search.js
```diff
@@ -0,0 +1,87 @@
...
+      storeFields: [
+        'id',
+        'citation'
+      ]
+    };
+
+    const docOpts = _.defaults( opts, DEFAULT_DOCUMENT_OPTS );
+
+    super( docOpts );
+  }
```
maxkfranz, line 84: It would be nice to have an async `fetch` function in the class that loads in the documents from the REST API and calls `index()` on them.

It would be useful to store the array of documents as a property in the class, alongside a map of the doc ID to the full doc JSON.  That map could be used in this subclasses' `search()` function, where it returns the full doc JSON instead of just the minisearch objects.

maxkfranz:
> > Do you think it would be useful to also include the interaction text, so you can search for 'A phosphorylates B'?
> 
> I index the document `text` field - isn't that it? I don't think interactions have this, but they do have a `type`/`association` (e.g. binding).

That sounds right.  Missed that.  Thanks

Issue #909, opened 2020-11-24
Caching of recent factoids


maxkfranz:
- The query to load the carousel on the homepage is slow.  It would be better UX to show the result right away.
- The cache should apply only to the new documents list, as shown on the homepage in the carousel (i.e. newest 20).
- Cache the result of  GET [`/api/document`](https://biofactoid.org/api/#/Document/get_api_document) in memory on the server.
- Invalidate the cache on PATCH [`/api/document/{id}/{secret}`](https://biofactoid.org/api/#/Document/patch_api_document__id___secret_), i.e. on submission.
- The cache may contain only the list of document IDs to keep things simple.  This avoids cases where document data changes in the interim, which may be difficult to monitor.  The lookups by ID should be much faster than the existing query with sorting and filtering.
- Queries that specify non-default parameters (e.g. not just the newest 20 docs) bypass caching entirely.  We just need the homepage to be fast.

jvwong:
@maxkfranz @gbader 

I have been testing a staging instance of factoid (beta.biofactoid.org; Digital Ocean) that includes caching for document pathway images on the homepage/search.

I noticed that the initial load of the search results is very memory intensive, to the point where it attempts to use up more memory than is available on the Droplet (maximum 8 GB, also the size of the production Droplet). The result is that the OS kills off other containers, namely the  Elasticsearch container and often the factoid container itself. The image below shows the stats at this point (16:10). Aside: Testing locally and on unstable didn't reveal this as the hosts both had a ton of RAM (>32 GB).

<img width="1002" alt="beta-biofactoid-DO-stats-memory-pre-post-resize" src="https://user-images.githubusercontent.com/4706307/161341130-38e40032-4e21-46a3-b5d4-25052d895690.png">

I tried an experiment where I resized the staging Droplet from 8GB to 16GB of memory, everything else the same (took about 4 min - this is the break point in the chart above). When I relaunched the factoid container and loaded the search, the memory was sufficient (little after 16:50 in the chart) and the factoid container peak usage was around 8 GB RAM.

After caching, the factoid container settles down and uses only around 600 MB of RAM.

So I think we might need to consider upping the production Droplet RAM (16 GB = $80/mo whereas our 8GB = $40/month), otherwise this version of the search won't work well. Alternatives ?  

PS. Initially I  tried varying the env var `DOCUMENT_IMAGE_CACHE_SIZE` but this didn't have any effect.

maxkfranz:
[...]

PR #986, opened 2021-06-09
Search UI on the home page

maxkfranz:
- Integrated with the home page.  No looking around to find another page or waiting for a new page to load.
- Shows all docs when search box is focussed and empty.  You can just scroll through the list.
- Shows top 10 search results when you type a query.
- Fixes DocumentSearch returning unsubmitted docs.
- Carousels can be refreshed to be used in search.  May be useful in future for categorised search results.

Ref. :

- Browsing: Beyond the carousel #863
- A class for client-side Document search #984


maxkfranz:
@jvwong, would you make sure that this is on a public instance for demonstration tomorrow?

jvwong:
Hey this looks great. I had a couple of initial thoughts:

- When I first loaded the page it was difficult to find the search feature, I guess most people have the magnifying glass icon which helps
- When you are searching and clear the text with the (x) icon, the page is empty. The alternative is to go back to browse mode.

I noticed a bug on mobile where the clicking the close (x) shifts the input over rather than close:

maxkfranz:
Let's leave this in a branch until we finalise the search strategy

jvwong:
This branch is useful in the admin, at the very least. I've been coming up with some ridiculously long and complicated Rethinkdb queries to achieve what this is going so maybe at least I will try to incorporate this into the admin.

maxkfranz:
How about we merge this in as-is for now?  We can disable the homepage search before putting this on prod if we want.

jvwong:
This one now has most of the missing pieces. This makes me think that #909 is worth taking a stab at whether we use this or not publically.

jvwong:
FYI I have the latest version of this branch (merged with #909) up on https://test.biofactoid.org/ which auto-builds a Docker image based on the `test` branch.

maxkfranz:
OK, I think it's good to merge

Issue #937, opened 2021-02-17
Network Explorer: Interaction information


jvwong:
## Description

**Q: What is the name of the feature?**

A: Network Explorer: Interaction information

**Q: What does this feature enable the user to do?**

A: Identify  "at a glance" what relevant interaction evidence an article contains; provide more context. Enable users to triage articles.

**Q: What information must the user provide to use the feature?**

A: Select element within the Explorer network.

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: We've got more experience with INDRA, so look out for edge cases:

- A large number of interactions (INDRA 'statements') are returned and must be triaged
- No interactions (statements) are returned
- A large amount of papers (evidence) is returned and must be triaged 
- How to deal with two types of papers/evidence:
  -  No text snippet available
  - Text snippet available

**Q: How does this feature affect each class of user (persona)?**

A: I considered this question from a biologists point of view. It was never clear to me why articles were being displayed for a selected interaction or participant. To find out exactly why, at least in a couple of cases I followed up on, would require perhaps a few hours. It also helps to know the type of article (review, research), for greater context. 

From a wider perspective, its clear that we should strive to offer information that is unique and 'sticky'. There's really interesting stuff from INDRA we're already retrieving, its just a matter of using it. By the same token, we'll be able to judge more easily when the items we're offering are of poor quality (and so will users).

### Mockup

I'm using Ben/John's entry (Sorger) as a worked example: [Encoding Growth Factor Identity in the Temporal Dynamics of FOXO3 under the Combinatorial Control of ERK and AKT Kinases. Cell Syst 6 2018](https://biofactoid.org/document/3d2a77ba-55c1-463b-aff3-9acaa0307b62)

#### Case I: Evidence for interaction 

![Evidence available](https://user-images.githubusercontent.com/4706307/108256979-2d73de00-712c-11eb-88ce-c92ac226834d.png)

Notes:
- Tell me that its a review is important and can help contextualize a text snippet

#### Case II: No evidence

The point here really is that if its a truly novel interaction, we should be explicit about this, rather than "back-fill" other papers into this section or view it as some failure of the approach. Just say "nothing here to see".

### Details

Getting data from INDRA is mostly the same, but the changes are downstream:

- Retrieve statements from INDRA by search via participant Uniprot ID (`/statements/from_agents`) 
- Sort statements:
  - Interaction text (template, e.g. "AKT1 inhibits FOXO3 via phosphorylation") vs.
[...]

Issue #985, opened 2021-06-09
Network-level related papers: Fallback when external service fails


jvwong:
Background: Related to [Regulation of mRNA export through API5 and nuclear FGF2 interaction. Nucleic Acids Res 48 2020](https://biofactoid.org/document/8e17bc30-b426-4ee1-addc-33067caabc8f) explorer view, network-level related papers.

- Issue: Some of the entities (e.g. eIF4, LRPPRC) and edges have no related papers associated with them.

- Reproducing this bug: I can reproduce locally with database dump. The offending items have no `relatedPapers` attribute set in the db.

Logs:

```
info:    Updating network-level related papers for doc 8e17bc30-b426-4ee1-addc-33067caabc8f
info:    PATCH /api/document/8e17bc30-b426-4ee1-addc-33067caabc8f/<secret> - - ms - -
error:   Null response for http://api.indra.bio:8000/assemblers/english
info:    Retrying, attempt 1
info:    Updating document-level related papers for doc 8e17bc30-b426-4ee1-addc-33067caabc8f
info:    Finished updating document-level related papers
info:    Updating network-level related papers for doc 8e17bc30-b426-4ee1-addc-33067caabc8f
error:   Null response for http://api.indra.bio:8000/assemblers/english
info:    Retrying, attempt 2
info:    PATCH /api/document/8e17bc30-b426-4ee1-addc-33067caabc8f/<secret> - - ms - -

error:   Null response for http://api.indra.bio:8000/assemblers/english
info:    Retrying, attempt 1
error:   Null response for http://api.indra.bio:8000/assemblers/english
info:    Retrying, attempt 3
....

```

jvwong:
related to #937

jvwong:
Doesn't look like the INDRA endpoint (http://api.indra.bio:8000) is reachable. Should have some sane fallback.

jvwong:
INDRA endpoint came back up, related papers OK now --- but this issue is still valid: Should have fallback when services fail.

maxkfranz:
Isn't the related papers refresh cron job turned on?  It should just fix itself on the next run, assuming that Indra is up.

jvwong:
The cron job is off because I wanted to include a fallback (https://github.com/PathwayCommons/factoid/pull/916) - so I can turn it back on. 

I'd like to turn all these things back on in the Digital Ocean version (either staging(master) / prod).

jvwong:
[...]

Issue #780, opened 2020-07-16
INDRA: Queries randomly fail


maxkfranz:
Sometimes INDRA queries randomly fail.  E.g., see @jvwong's example document: https://github.com/PathwayCommons/factoid/pull/777#issuecomment-658975666

Sometimes the queries go through just fine.  Other times, INDRA dies: 

```
Unable to retrieve the indra staments for the entity 19102@HGNC
```

jvwong:
Yes, this morning it was giving off a server error for https://db.indra.bio/latest/statements/from_agents?agent0=CD21@TEXT&agent1=IL-6@TEXT&format=json-js but now it worked

maxkfranz:
Just now I'm getting it failing about half of the time.

pagreene:
@maxkfranz What errors are you getting, specifically?

maxkfranz:
@pagreene, I don't see any details about the error, but it looks like the request might not be getting a response.   @metincansiper, would you be able to track down the specific error?

metincansiper:
No actually I have not coincided the time it was failing.

metincansiper:
@pagreene we are considering to switch to stable version of Indra as it is suggested over Slack. Our main concern is whether there is a big difference in the data for the stable version versus latest version. Do you have any idea about that?

pagreene:
The data will be different, and in particular the content you get from stable is at this point many months old. On the other hand, we are (a) looking to update stable in the near future, but (b) latest should generally be _pretty_ stable generally.

maxkfranz:
At this point, we're prioritising new articles for our related articles feature.  So for now, it's preferable to have fresh data.

jvwong:
This one ([`agent0=19102@HGNC`](https://db.indra.bio/latest/statements/from_agents?agent0=19102@HGNC&format=json-js)) is failing consistently:

```sh
> GET /latest/statements/from_agents?agent0=19102%40HGNC&format=json-js HTTP/2
> Host: db.indra.bio
> user-agent: insomnia/2020.3.3
> accept: */*

* Connection state changed (MAX_CONCURRENT_STREAMS == 128)!

< HTTP/2 502 
< content-type: application/json
< content-length: 36
< date: Mon, 20 Jul 2020 19:53:45 GMT
< x-amzn-requestid: ec7f8c61-f46b-45d7-b296-5efb9fa823ff
< x-amzn-errortype: InternalServerErrorException
[...]

PR #988, opened 2021-06-15
Exploring at the network-level: Related Interactions

jvwong:
This is a prototype. Displays more detail of INDRA search results upon interacting with elements of the network in our Explorer.

The original motivation for this feature was to provide more explicit information about why the papers were being displayed for a given network element.

- Notes
  - It is backwards compatible in the sense that it displays 'There were no interactions found for the selected item' for `element.relatedPapers` data with the existing schema, or if in fact none were found
  - 'Show more articles' isn't implemented, not sure if this should be punted til later and how it should be implemented (routing?)
  - I'm showing 'Source' (e.g. Pathway Commons; Human Protein Reference Database) when available but not sure if this is worth including or how to include it  (link to PC apps?).

Refs #937


metincansiper (approved):

maxkfranz:
@jvwong, would you merge this in and link to a live example of the results on the unstable instance?

The code looks OK, and I'd like to have a live example to point to in order to comment on the UI.

jvwong:
OKey, I'll put in on master.factoid.baderlab.org - ping you when its done

jvwong:
@maxkfranz  I updated the site, and kicked off an example: https://master.factoid.baderlab.org/document/e5dc3eef-6af0-4c61-8a42-ddeb4a4285b5

maxkfranz:
Looks good.  Minor feedback:

- It would be nice if there were just a bit of spacing between the title, author list, and journal name.  It's close, making it a bit harder to make things out.
- The journal name wouldn't blend in so much with the author list if it were italicised.

jvwong:
We've tries to mitigate failures and provide fall-backs upon 
 - retrieving network related papers from INDRA (#785, #780) 
 - refreshing papers metadata post-submission (#985)

Another existing issue is that when semantic-search returns an error, for whatever reason (https://github.com/PathwayCommons/semantic-search/issues/98) there is no reasonable fallback herein, which ultimately leads to no related papers - its 'all or none'. 

All of these points of failure should be considered in this PR since the relevant code is being updated.

Issue #351, opened 2018-06-29
DEMO button on landing page


sacdallago:
## Description of new feature

On the landing page, have a button (similar to _Try factoid_) that brings the users to the /new/seed page with a populated text-box (populated using the same abstract that is used for the _Try factoid_ button).

## Motivation for new feature

Showcase the whole process: insert text, extract annotations, select editor, edit.

jvwong:
From the paper review:

> 6. Please consider providing a demo to try out the editing interface in a 'sandbox' without having to submit any information (article title, name, etc.). This can help users decide whether they can contribute and which articles may be most appropriate for them to work on.

jvwong:
There's a bit of a lag on the demo (#995) when clicking the DEMO from homepage: anywhere from 3 - 6 seconds. May give the impression something is broken...

PR #995, opened 2021-07-08
Editor demo on homepage

jvwong:
Provides a demo for the editor on the homepage. Each time the `/demo` route is hit a new, live document is created.

- Details
  - Status immediately set to 'trashed' so submit is disabled
  - bypasses related paper retrieval and any emailing 
  - Using a real but generic article for the doc.article 

- Possible issues?
 - Not sure if swapping in a demo doc in client-editor  is going to cause problems?
 - Creates real docs but with the shared secret 'demo' - makes it easier to delete en masse though

Screenshot at Jul 08 12-23-22

Screenshot at Jul 08 12-24-22

Refs #351

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ npm-debug.log
 *.iml
 /download/*
 !/download/README
+.vscode
--- a/src/client/components/editor/index.js
+++ b/src/client/components/editor/index.js
@@ -204,9 +204,22 @@ class Editor extends DataComponent {
               authorEmail: DEMO_AUTHOR_EMAIL
             })
           });
+          let toJSON = res => res.json();
           let reloadDoc = () => doc.load();
+          let updateDoc = docJSON => {
+            const { id, secret } = docJSON;
+            doc = new Document({
+              socket: docSocket,
+              factoryOptions: { socket: eleSocket },
+              data: { id, secret }
+            });
+            this.setData('document', doc);
+          };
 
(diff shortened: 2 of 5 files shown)


maxkfranz:
> Status immediately set to 'trashed' so submit is disabled

This may be an issue if a user expects to be able to test the demo from beginning to end.  Some options:

(1) Allow the user to submit demos.  This may risk junk demo data being confused with real article data.

(2) Add a special tooltip on the submit button for the demo case, explaining to the user why they can't submit.

Option (2) seems simpler and less risky.

jvwong:
> (2) Add a special tooltip on the submit button for the demo case, explaining to the user why they can't submit.
> 
> Option (2) seems simpler and less risky.

Are you thinking to keep the Submit button enabled (just swap popup message)?

jvwong:
I left the demo document status alone, and just put a Submit button popup message

image

maxkfranz:
> I left the demo document status alone, and just put a Submit button popup message
> 
> image

If you think the status is useful, it's fine to preset the status on demos -- e.g. they need to be filtered out of data exports somehow.  

The main thing is that the user gets some feedback so they're not confused.

jvwong:
Then I won't make anymore updates: The demo docs can be filtered and removed via secret === `DEMO_SECRET`.

Also I should put up a ticket about garbage collecting these ones....

Issue #994, opened 2021-06-28
Handling documents that are 'removed' from Biofactoid data


jvwong:
> 4. Please address whether there is a process for removing a submission altogether (e.g. if a submission is found to be incorrect or if a paper is being retracted).

> - Easy to update in the text: TODO -- points to consider:
> - When DOI: Marked as retracted - could be a box tick in the admin interface, email our support address
> - Currently: support email used on case-by-case
> - Could also automatically get this information from PubMed - erratum and retractions. Probably a future feature. (e.g. https://pubmed.ncbi.nlm.nih.gov/20829833/)

jvwong:
I'm thinking a 'removed' doc.status that can be accessed via admin.

jvwong:
Actually why not just set to 'trashed' - these should be backed up anyways...

jvwong:
Another more realistic option is to keep the document as is, and rely on the cron to pick up the citation information:

```json
{
  "id": "42d45f8d-892f-4a75-bccc-7b1510732557",
  ...
"citation": {
     ...
    "pubTypes": [
      {
        "UI": "D016428",
        "value": "Journal Article"
      },
     ...
      {
        "UI": "D016441",
        "value": "Retracted Publication"
      }
    ],
    "ISODate": "2008-12-09T00:00:00.000Z"
  },
  ...
}
```

PR #998, opened 2021-07-14
Insert a "Retracted Publication" flag in Explorer

jvwong:
Associated article is flagged (by Pubmed) as retracted, Explorer displays a banner:

Screenshot at Jul 14 15-10-48

Refs #994


maxkfranz:
It would look a bit more consistent with other article sites if it looked more like a tag, like your review tag -- but red:

Screen Shot 2021-07-15 at 14 58 06

Issue #1000, opened 2021-08-03
Normalize document time stamps


jvwong:
It's NoSQL, so it depends on the value you send to the DB.  Regardless of where the inconsistency is, the API (probably the `Document`) should simply enforce a normalisation.

_Originally posted by @maxkfranz in https://github.com/PathwayCommons/factoid/issues/997#issuecomment-892030677_

jvwong:
I'm stumped on this one. 

Not sure how Date set in the model / browser can be interpreted correctly on server.

maxkfranz:
Here's a search: https://github.com/PathwayCommons/factoid/search?q=lastEditedDate

It looks like it's always set as a `Date` type in the `Document` already.  So then, could it be that these problematic docs were generated from an older version of the code?  If that true and the code is fine now, then we just need to amend the problematic docs manually.

Another option is to change the code so it's always a string type.  We'd have to be consistent about the format used (probably just use ISO), but it would guarantee that there aren't any mismatches.  The downside with that option is that you can't do DB queries on the time anymore, since it's not a `Date` type.

jvwong:
> It looks like it's always set as a `Date` type in the `Document` already. So then, could it be that these problematic docs were generated from an older version of the code? If that true and the code is fine now, then we just need to amend the problematic docs manually.
> 
> Another option is to change the code so it's always a string type. We'd have to be consistent about the format used (probably just use ISO), but it would guarantee that there aren't any mismatches. The downside with that option is that you can't do DB queries on the time anymore, since it's not a `Date` type.

This isn't an issue of old docs. Currently, at doc initialization, the server (nodeJS) sets `lastEditedDate` and its interpreted correctly as a rethinkdb `time` type. I think the problem is that, upon doc editing, a Date object in the browser gets inserted into  the `model/Document/Syncher` in the browser, and  the server / rethinkdb, interprets this as String. So my best guess is that `Date` needs to be set directly on the nodeJS server for rethinkdb to correctly interpret it as a `time` object.   

I guess we can just enforce a string and remember to use my little hack:

https://github.com/PathwayCommons/factoid/blob/065b82d184435c9060dadd1b4617d9ed1de645bd/src/server/routes/api/document/update.js#L91-L94

maxkfranz:
It may be that it's not being serialised across the socket properly.  Dates aren't proper JSON values; they're JS objects (from classes).

[...]

PR #1002, opened 2021-08-06
Normalize document dates

jvwong:
- Document model
  - set dates as UNIX time (seconds since epoch) 
  - get dates as JS Date objects in getters and via .json() method
- other
  - provide `toTime` patch that always returns a rethinkdb `PTYPE<TIME>` object (backwards compatibility)

The downside here is that if you use the rethinkdb admin console, you'll get a grab bag of data types in the `lastEditedDate` field, so its harder to sort.

Refs #1000

### DIFF
--- a/src/model/document/document.js
+++ b/src/model/document/document.js
@@ -18,7 +18,7 @@ const DEFAULTS = Object.freeze({
   verified: false
 });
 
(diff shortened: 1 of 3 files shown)


maxkfranz:
Looks good.  In terms of consistency, we could patch the existing data in future so it's all using the number format.

Issue #1032, opened 2022-01-03
Zip bulk download: Recreates new bulk file on each request / Gateway Timeout


jvwong:
We [advertise](https://github.com/PathwayCommons/factoid#getting-the-data) bulk data downloads via `/api/document/zip`. However, I noticed that:

1. Browser responds with a Gateway Timeout error when visiting the link 
2. Looking at the cached file in the factoid container at `downloads/factoid_bulk` it appears to be recreated each time a request is made

This creates a barrier to data download.

- Cannot reproduce on unstable instance (works fine here)
- Able to reproduce on beta instance: Gateway Time-out (1); cannot reproduce (2)

---

Aside: 

- I wonder if this endpoint is being hit by bots/Google
- We should probably not generate this file on the fly, other optimizations

jvwong:
With regards to the recreation, the following

https://github.com/PathwayCommons/factoid/blob/3e5c23028a12780706d04e11f8ae83d1b2337282/src/server/routes/api/document/index.js#L631

should actually reference the time the file was modified (`mtime`) rather than the file creation time (`birthtimeMs`), which is a fixed time (in the past).

jvwong:
Aside from above bug regarding recreate time stamp, the `/api/document/zip` endpoint doesn't provide an HTTP response until it's finished dumping the bulk data - even with some optimizations, this takes a while (couple of minutes). 

One idea is to have this endpoint instead:
  - serve existing files 
  - ensure the files  
    - exist on boot 
    - are updated on some regular interval 'offline' (e.g. cron), not lazily

PR #1034, opened 2022-01-06
Create Biofactoid data downloads 'offline' 

jvwong:
This update attempts to make creating files with Biofactoid data and downloading them more efficient.

- File downloading  
  - The `/api/document/zip` and `/api/document/biopax/zip` endpoints just serve file(s) that are already created
- File creation
  - Listens for database events that either create or modify documents with status 'public'
  - Events fire off a request to `exportToZip` as before to create the required file
  - Requests for file creation are delayed ad batched within a specified time window (`EXPORT_BULK_DELAY_HOURS`) 

Refs #1032


jvwong:
To consider: File creation on app reboot

src/server/routes/api/document/export.js
```diff
@@ -99,4 +109,105 @@ const exportToZip = (baseUrl, zipPath, types, biopaxIdMapping) => {
...
+
+/**
+* initExportTasks
+* Initialize the export tasks
+*/
+const initExportTasks = async ( exportOnInit = true ) => {
+  const MS_PER_SEC = 1000;
+  const SEC_PER_MIN = 60;
+  const MIN_PER_HOUR = 60;
+  const baseUrl = `http://localhost:${PORT}`; // ever not localhost?
```
maxkfranz, line 181: Not really needed anyway, since you could call the corresponding function directly instead of going through the route.  It works either way, though going the routes could slow things down.

src/server/routes/api/document/index.js
```diff
@@ -619,32 +620,8 @@ const generateSitemap = () => {
...
-      return exportToZip(addr, filePath);
-    }
-
-    return Promise.resolve();
-  };
-
-  tryPromise( lazyExport )
-    .then( () => res.download( filePath ))
+http.get('/zip', async function( req, res, next ){
+  tryPromise( () => res.download( BULK_DOWNLOADS_PATH ) )
```
maxkfranz, line 624: It could be useful to log errors here in case the generation of the zip somehow failed

maxkfranz (commented):
Some notes

src/server/routes/api/document/export.js
jvwong: Makes sense - this requires extracting the logic from the route handlers:
  - `/api/document/` and also
  - `/api/document/biopax`, 
  - `/api/document/sbgn`
  - `/api/document/text`

I'll try, but if it gets hairy I'll just leave it.

jvwong: - Refactoring converter functions
  - Stripped out the logic from the endpoints and reused them directly, avoiding the web service routes.
- Fixed up the behaviour on boot: 
  - When `NODE_ENV` is `production`, always create exports
  - When `NODE_ENV` is not `production` (e.g. dev), create the export zips if any are missing (so the web service routes work), otherwise do nothing

metincansiper (approved):

jvwong:
> Looks good to me. Just wondered one thing why there is a delay shift for biopax exports (https://github.com/PathwayCommons/factoid/pull/1034/files#diff-b0e82ba0bf6ce7365597b983e42f14dcca6f038f4297df4c1f811a28f88d035dR190)? Is there a specific reason of why you want to start biopax exports after bulk exports?

I will get rid of that, probably overthinking things.

Issue #1064, opened 2022-04-04
Retrieve ORCID information accurately and more comprehensively


jvwong:
Disambiguating authors via "Open Researcher and Contributor ID" (ORCID) can be useful in associating a researcher's various contributions and crediting them, as well as offering infrastructure to authenticate. 

Currently, we use the ORCID web service to perform a general search, querying with the DOI and filtering on name. However, this approach is a bit fuzzy (i.e. matching strings) and does not use the valuable information indexed by PubMed, resulting in incomplete sets of ORCIDs. 

The goal here is to combine the exact information indexed by PubMed and ORCID, and normalize this information within a document (i.e. `authorProfiles`). 

---

Some stats about documents from the database:

|  Document Category      |     Count     |
|-------------------------|------------|
| Has Contributor name    |     85         |
| Has >=1 ORCID               |     70         |
| Has ORCID for Contributor name   |     42      |

Related to #1056 #1017

PR #1065, opened 2022-04-11
Refactor ORCID disambiguation to be more precise and comprehensive 

jvwong:
- Search the ORCID public API using [fielded search: DOI, 'self'](https://info.orcid.org/documentation/api-tutorials/api-tutorial-searching-the-orcid-registry/#easy-faq-2707)
  - For unique last names, match the ORCID search result on last alone, else resort to first name
  - Store the ORCID accession (local ID) rather than the url moving forward
  - Lots of weird edge cases like
    - PubMed putting ORCIDs in the [wrong format](https://pubmed.ncbi.nlm.nih.gov/25086665/) 
    - PubMed can store the ORCID URL or the accession 

Refs #1064


jvwong:
@metincansiper do you remember why the Document `provided` field has a `name` attribute rather than just using the `authorName`?

https://github.com/PathwayCommons/factoid/blob/94586ccfb62fe4c7b65a1080d7fd96e40e7e6bf3/src/server/routes/api/document/index.js#L1911-L1917

metincansiper:
> @metincansiper do you remember why the Document `provided` field has a `name` attribute rather than just using the `authorName`?
> 
> https://github.com/PathwayCommons/factoid/blob/94586ccfb62fe4c7b65a1080d7fd96e40e7e6bf3/src/server/routes/api/document/index.js#L1911-L1917

@jvwong I actually could not remember any specific reason for that. I suspect it may just be a confusion of me at that time. Actually, with that it seems like that `name` is used instead of `authorName` but both of them are being available because `authorName` is used as is during the doc creation and `name` is added later as a new field. 

See this line (https://github.com/PathwayCommons/factoid/blob/unstable/src/client/components/editor/title.js#L60) and line 43 above that `authorName` is retrieved from `document.provided()` and it seems to be working since `authorName` is also available (BTW I guess in line 60 it may make sense to use the already defined `provided` variable).


There may be 2 things we must consider before removing `name` field:

- There may be a very little possibility that there would be reason behind setting `name` filed that I cannot remember now.
- Because of some possible dynamic assignment of field names etc. we may end up with having a bug while renaming that field (as it is used by different parts of the code now, probably implemented by different people I may miss something).

Considering the 2 thing above and that both of the fields (`name` and `authorName`) are available, maybe we can consider keeping the `name` field but replacing any explicit usage of it with `authorName` and add a comment to where `name` field is introduced so that we would encourage usage of `authorName` instead of it.

What do you think?

metincansiper (approved):

jvwong:
> There may be 2 things we must consider before removing name field:

I agree that's its probably redundant.

> Considering the 2 thing above and that both of the fields (name and authorName) are available, maybe we can consider keeping the name field but replacing any explicit usage of it with authorName and add a comment to where name field is introduced so that we would encourage usage of authorName instead of it.

I won't change anything right now. I think these changes, if made, could be part of another issue so don't worry about making any changes.

Thanks!

Issue #1041, opened 2022-01-26
Add direct link for article when sharing (e.g. explorer)


jvwong:
Add the direct link when posting to Twitter or elsewhere - Why ?

[re: Altmetric 'Sources of Attention'](https://help.altmetric.com/support/solutions/articles/6000235926-twitter)
> Altmetric tracks Twitter attention in real-time via an API. We automatically collect tweets, retweets, and quoted tweets that contain a direct link to a scholarly output. Without the direct link to an output, our system will not be able to match the tweet with the appropriate output.
> Altmetric does not track tweets that contain links to news stories or blog posts about scholarly outputs (e.g a tweet of a news story about a journal article). These are considered "second-order citations", as they are mentions of attention about the output and rather than the output itself.

maxkfranz:
We'll have to consider how this may affect our cards if we use a different link

jvwong:
Easiest solution is to append link ('Read the paper: https://doi.org/<doi>') to the auto-generated text of interactions.

jvwong:
Basically, last url posted with a card wins (this is opposite what the [actual docs](https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started#multiple) say)

PR #1078, opened 2022-09-01
Reply to Tweet with paper link

jvwong:
On successful Tweet, reply to that status with a link to the paper. This is cleaner than jamming it all in one tweet.

Sample result below:

Screen Shot 2022-09-01 at 3 04 48 PM

Refs #1041


maxkfranz:
Great.  Many people use threads like this, so it's a good fit

Issue #1053, opened 2022-03-10
Event metrics


maxkfranz:
## Description

**Q: What is the name of the feature?**

A: Event metrics service

**Q: What does this feature enable the user to do?**

A: Provide implicit feedback

**Q: What information must the user provide to use the feature?**

A: Just use the interface

**Q: What are the applicable constraints, e.g. compatibility or performance?**

A: Must be a server-driven API

**Q: How does this feature affect each class of user (persona)?**

A:

- Biologist: Provide implicit feedback for which features are best
- Editor: "
- Computational biologist: "
- Curator: "

### Mockup

- A user performs event X, which is something that is valuable for us to know w.r.t. design.
- Action X is logged.

### Details

- Provide an endpoint, e.g. `/api/event`, that we POST to for relevant events.
- The payload should be enough to send a request for this:
  - Docs: https://developers.google.com/analytics/devguides/collection/protocol/v1
  - Request: https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide
  - Params: https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters
  - Note: May need some sort of anonymous user ID depending on what we want to measure

jvwong:
Looking at what this might entail:

- Client 
  - general data
    - client ID (UUID)
  - type: event
    - category (String, e.g. 'video')
    - action (String, e.g. 'play')
    - label (String, e.g. 'holiday')
    - value (Int, e.g. '300')
- Server
  - id (uuid) via [cookie-session](http://expressjs.com/en/resources/middleware/cookie-session.html)

jvwong:
[Universal Analytics will be going away](https://support.google.com/analytics/answer/11583528)

jvwong:
- TODO
  - [ ] Get GTM code into production 
  - [ ] *Consider events to track

* straightforward video on event tracking with [Google Tag Manager](https://www.youtube.com/watch?v=ghUwerH1fsw)

jvwong:
Events to track (many are already tracked by GTM)

- [ ] Click CTA
- [ ] Search terms

PR #1109, opened 2023-02-03
Migrating to Google Analytics 4.

jvwong:
Google analytics 4 tagging.

Refs #1053


maxkfranz:
Great, will the new ID carry over all the old stats in the panel or will the stats start over?

jvwong:
Sorry hold on - I need to update this. 
Google analytics 4 is a zoo of new jargon...

> Great, will the new ID carry over all the old stats in the panel or will the stats start over?
It will start over, but maybe there's a way to import.

jvwong:
OK this Google Tag Manager approach gets 90% of the stuff we care about. Let's pull it in and I'll figure out any other stuff later.

Issue #1085, opened 2022-10-04
Document requirements: python and node version


jvwong:
You need python and the node version is restricted (in .nvmrc). Otherwise the errors are cryptic.

jvwong:
Version 10 could not support chrome - see  #1111

PR #1117, opened 2023-02-07
Use version 16 LTS

jvwong:
Refs #1085 #1111


maxkfranz:
Maybe package.json too if we're bumping everything up

PR #1121, opened 2023-02-16
Neo4j Testing

lindajiawenli:
For writing Neo4j Tests

### Possible Tests (work in progress)

- [x] 1. Verify DB connection with Neo4j driver
- [ ] 2. Verify 1 node is being made correctly
- [ ] 3. Verify 1 edge is being made correctly
- [ ] 4. Verify that attempting to make a duplicate node fails
- [ ] 5. Verify that attempting to make a duplicate edge fails
- [ ] 6. Verify that searchByGeneId function/read database is correct
- [ ] 7. Verify that all the nodes in a document are made (no duplication or missing nodes)
- [ ] 8. Verify that all the edges in a document are made (no duplication or missing nodes)
- [ ] 9. Verify that all the nodes in multiple documents are made (no duplication or missing nodes)
- [ ] 10. Verify that all the nodes in multiple documents are made (no duplication or missing nodes)
- [ ] 11. Verify that a node deletion is successful (function exists for testing, not sure if it is useful in our case)

refs #1122

PR #1123, opened 2023-02-16
GraphDB configuration variables

jvwong:
- Configurable variables for the graph database (Neo4j)
- ~~Driver code initializes if doesn't exist~~
- tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,12 @@ AppSignal:
 - `APPSIGNAL_APP_NAME` : name of this app (e.g. 'Biofactoid')
 - `APPSIGNAL_APP_ENV` : used to indicate which instance is running (e.g 'master', 'production', 'unstable')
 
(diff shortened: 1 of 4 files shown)


lindajiawenli:
Hi @jvwong, it looks good to me! Definitely an improvement over the `connect-neo4j.js` edits [I made over at my pr.](https://github.com/PathwayCommons/factoid/pull/1121/commits) 
Good call making the username and password "undefined" instead of just ignoring it like I did! 

Is this ready to be merged?

jvwong:
One more thing @lindajiawenli - the docs say that Driver.verifyConnectivity is deprecated (but only for some return type). So anyways that's why I got rid of it.

lindajiawenli:
Yeah I noticed that too! Thanks for working around it.

PR #1129, opened 2023-02-21
Neo4j Testing (Part 2)

lindajiawenli:
For writing Neo4j Tests (Part 2, after [accidentally merging the first one](https://github.com/PathwayCommons/factoid/pull/1121))

Possible Tests (work in progress)

 
- [x] 1. Verify DB connection with Neo4j driver
- [x]  2. Verify 1 node is being made correctly
- [x]  3. Verify 1 edge is being made correctly
- [x]  4. Verify that attempting to make a duplicate node fails
- [x]  5. Verify that attempting to make a duplicate edge fails
- [x]  6. Verify that searchByGeneId function/read database is correct for an non-existing node
- [x]  7. Verify that searchByGeneId function/read database is correct for an existing node
- [ ]  8. Verify that all the nodes in a document are made (no duplication or missing nodes)
- [ ]  9. Verify that all the edges in a document are made (no duplication or missing nodes)
- [ ]  10. Verify that all the nodes in multiple documents are made (no duplication or missing nodes)
- [ ]  11. Verify that all the nodes in multiple documents are made (no duplication or missing nodes)
- [ ]  12. Verify that a node deletion is successful (function exists for testing, not sure if it is useful in for our user)


lindajiawenli:
I have a couple tests in the example.js file now, that I would like to receive feedback on! The next couple tests need the Document API, I believe.

I would also be interested in hearing about more basic edge cases I can make tests for . Right now all I have to test the basic addNode, addEdge and searchByGeneId functions and I am open to making more low level functions. 

I do have some functions that I use only for testing in the src/neo4j/test-functions.js file

neo4j-test/example.js
```diff
@@ -11,26 +11,116 @@ describe('Example set of tests', function () {
...
+    expect(await getNumEdges()).equal(1);
   });
 
-  it('Ensure searchGeneById works as expected', function () {
+  it('Ensure searchGeneById works as expected for MAPK6', async function () {
+    let mapk6 = await searchByGeneId('ncbigene:5597');
+
+    let mapk6Relationships = mapk6.map(row => {
+      return row.get('r');
+    });
```
maxkfranz, line 77: It would be useful to have some slightly higher level functions here, e.g. `searchByGeneID()` could directly return an array of edge JSONs.

maxkfranz (commented):
Great start.  I've attached some suggestions for next steps on the low level functions before moving on.

neo4j-test/example.js
lindajiawenli: I've made two functions building off of `searchByGeneId`:

`getNeighbouringNodes( id )`: returns the array of nodes connected to/from the specified gene

`getInteractions( id )`: returns the array of edges/relationships leading from/towards the specified gene

Let me know what you think and please take a look at my other (rather small) amendments when you get the chance! Thank you so much for the feedback

jvwong:
Re Docker: You can run both databases with `docker-compose up -d`. The data should persist between runs as it is stashed in a 'volume' somewhere on your machine.

lindajiawenli:
I believe I'm done with this PR for the time being, if you could give it another once-over, I would appreciate it!

jvwong (approved):
Looks good to merge in - just one question: Is `addEdge` unique by relationship `id`? Biofactoid allows one to add an arbitrary number of edges between nodes (identical or otherwise).

lindajiawenli:
> Looks good to merge in - just one question: Is `addEdge` unique by relationship `id`? Biofactoid allows one to add an arbitrary number of edges between nodes (identical or otherwise).

Yes, right now the grounding id is the unique UUID every interaction in a factoid document has. I couldn't think of a more suitable one. 

If it was the two nodes the edge is connected to- then there can only be one edge between those two nodes ever, which does not make sense. That was the other grounding id(s) I was thinking about

Issue #1133, opened 2023-03-07
Document: Accept article in doc instance  


jvwong:
The `Document` model can accept a JSON representation to populate an instance via `fromJson()`. Enable this method to accept article information. This will allow it to output a `citation` field. 

Related to [#1132 ](https://github.com/PathwayCommons/factoid/pull/1132#issuecomment-1456912488)

PR #1134, opened 2023-03-07
Document from JSON: article

jvwong:
Enable the article info to be included in Document creation from JSON. 

Refs #1133


lindajiawenli:
Unfortunately it still doesn't seem to be working for me- I switched to this branch, copy-pasted some relevant test, and it doesn't pass/I can see from the breakpoints that I still don't see citations. Can we work on this tomorrow?

Screen Shot 2023-03-07 at 6 41 55 PM

lindajiawenli (approved):

Issue #1135, opened 2023-03-08
RethinkDB for testing purposes: Cleanly delete before each test


lindajiawenli:
The current way of clearing the testing RDB sometimes does not complete fully, resulting in errors when the test runs again ("Duplicate primary id found").

New Tentative Approach: Cleanly delete database before each test runs, and create a new empty one

jvwong:
I found an NPM library `rethinkdb-fixtures` that lets you create database test fixtures directly from file (e.g. a direct dump of the database), and looks simpler.

 Let me put it up as PR and we can discuss.

https://rethinkdb.com/docs/frameworks-and-libraries/

PR #1136, opened 2023-03-08
Add article to test document JSON 

jvwong:
This updates the document-level tests so that all necessary JSON info (i.e. article) is incorporated into a Document instance using fromJson.

I commented out the breaking tests.  Also need to address #135.


lindajiawenli (approved):
Looks good to me! Citation field is consistently generating now

Note: Issue with database not being properly cleaned persists (see https://github.com/PathwayCommons/factoid/issues/1135)

PR #1140, opened 2023-03-09
addDocumentToNeo4j now uses Document instance

lindajiawenli:
Previously used to take in a plain object as its parameter.

I have made the appropriate adjustments so that the function uses the Document model instance as required.

### DIFF
--- a/neo4j-test/doc-tests.js
+++ b/neo4j-test/doc-tests.js
@@ -4,7 +4,7 @@ import rdbFix from 'rethinkdb-fixtures';
 import fixture from './testDoc.json';
 import { loadDoc } from '../src/server/routes/api/document/index.js';
 import { initDriver, closeDriver } from '../src/neo4j/neo4j-driver.js';
-import { addDocumentToNeo4j } from '../src/neo4j/neo4j-document.js';
+import { addDocumentToNeo4j, convertUUIDtoId } from '../src/neo4j/neo4j-document.js';
 import { deleteAllNodesAndEdges, getGeneName, getNumNodes, getNumEdges, getEdge } from '../src/neo4j/test-functions.js';
 import r from 'rethinkdb';
 
@@ -37,7 +37,7 @@ describe('Tests for Documents', function () {
     await rdbConn.close();
   });
 
-  beforeEach('Delete nodes and edges', async function () {
+  beforeEach('Delete nodes and edges from Neo4j', async function () {
     await deleteAllNodesAndEdges();
   });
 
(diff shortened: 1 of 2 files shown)


lindajiawenli:
However, I believe I can:
1. Make a more elegant way to determine whether an element is an Entity (Node) or Interaction (Edge). Current implementation uses an array called `nodeTypes` in `src/neo4j/neo4j-document.js`
2. Find a better way of converting an entity's UUID to it's neo4j id form (ex. "ncbigene:207"). See current implementation with `converUUIDtoId` function in `src/neo4j/neo4j-document.js`

Let me know what you think @maxkfranz @jvwong 
Should I do the above? And if so, should I make new PRs/Issues or just add commits to this PR

neo4j-test/doc-tests.js
```diff
@@ -62,21 +62,20 @@ describe('Tests for Documents', function () {
...
-    // let edge = await getEdge('01ef22cc-2a8e-46d4-9060-6bf1c273869b');
-    // expect(edge.type).equal('INTERACTION');
-    // expect(edge.properties.type).equal('phosphorylation');
-    // expect(edge.properties.sourceId).equal('ncbigene:5597');
-    // expect(edge.properties.targetId).equal('ncbigene:207');
-    // expect(edge.properties.xref).equal(myDoc.id);
-    // expect(edge.properties.doi).equal('10.1126/sciadv.abi6439');
-    // expect(edge.properties.pmid).equal('34767444');
-    // expect(edge.properties.articleTitle).equal('MAPK6-AKT signaling promotes tumor growth and resistance to mTOR kinase blockade.');
+    expect(await getNumEdges()).equal(1);
```
jvwong, line 69: Might want to consider in the future comparing to the document instance rather than hardcoding, just thinking if wen end up testing a bunch of different doc cases (e.g. complexes). 

e.g. 
```js
...
   expect(edge.properties.xref).equal(myDoc.id());
    expect(edge.properties.doi).equal(myDoc.citation().doi);
    expect(edge.properties.pmid).equal(myDoc.citation().pmid);
    expect(edge.properties.articleTitle).equal(myDoc.citation().title);
...
```

jvwong (approved):

neo4j-test/doc-tests.js
lindajiawenli: Hmm, that's a good point. 
What about when we have documents to test that aren't binary? Ex. 3 nodes and 2 edges. Should I make different tests for that, or should I have some kind of loop

```
for (let n = 0; n < numNodes; n++) {
  // test this node
}
```

lindajiawenli: @jvwong could you take another look at my test now? I've soft coded all the parts and utilized loops so that the test can handle an arbitrary number of nodes/edges. Let me know what you think!

I was taught in school that tests should be hard coded and I guess that's good for the assignments they gave us, but not for this scenario

jvwong: Works for me. Whatever is easiest to read.

Issue #1146, opened 2023-03-17
Neo4j: Handling Complexes


lindajiawenli:
Some more details can be found in this pdf: [Hairball Complex Notes.pdf](https://github.com/PathwayCommons/factoid/files/11001831/Hairball.Complex.Notes.pdf)

### Basic Idea
A complex will be represented by a "hairball" of nodes and edges.
Suppose n is the number of elements in the complex. 
We will be making n nodes (if they do not yet exist) and n(n-1) _Complex edges_.

![Screen Shot 2023-03-17 at 9 38 42 AM](https://user-images.githubusercontent.com/66929920/225920715-6a35b16e-2f40-4047-977b-6593ffe6f903.png)

### What will a _Complex edge_ look like in Neo4j?
`type`: COMPLEX
`id:` UUID from factoid (this is the grounding id)
`complexParticipants`: an array of all the node ids (ex. 'ncbigene:201')  that are in the complex

### What will a Complex interacting with a non-complex look like?

![Screen Shot 2023-03-17 at 9 41 56 AM](https://user-images.githubusercontent.com/66929920/225921601-a505a214-82ad-463a-ab7d-5a9c1df7c0c4.png)

### What will a Complex interacting with a Complex look like?

![Screen Shot 2023-03-17 at 9 42 46 AM](https://user-images.githubusercontent.com/66929920/225921799-d4435163-3796-482a-9617-504d247066f5.png)

lindajiawenli:
@jvwong Here's a TL:DR version of the notes I sent you

maxkfranz:
Make sure to consider the ability to reconstruct the left side from the right side, if you decide to store the data this way (e.g. for visualisation use cases).

A ‘complex edge’, as you’ve defined it, lets you know there is a complex, but it is not sufficient on its own to account for, for example, the inter-complex edges in your second example.  How do you know the edges are between the two complexes and not between the individual proteins?

lindajiawenli:
@maxkfranz You're right, thank you-that's a good point.

I'm thinking I can add a property in all edges: `participant-types` that is:

- `'noncomplex-to-noncomplex'` by default
- `'complex-to-complex'` when both participants are complexes
- `'complex-to-noncomplex'` when the source is a complex
- `'noncomplex-complex'` when the source is a non-complex


[...]

PR #1147, opened 2023-03-17
Iss1146: Neo4j Handling Complexes

lindajiawenli:
See #1146 for more detail

Making changes to code for handling Complexes via the "hairball" visualization

The following tasks (non-exhaustive) must be completed:

### OLD Plan:
- ~~Make a function similar to `addEdge()` that makes a COMPLEX edge rather than a regular INTERACTION edge. Call this `addComplexEdge()`. Should have fields for: `id`, `targetId`, `sourceId` and `allParticipants` (`allParticipants` is an array)~~
- ~~Make a Neo4j Cypher query string for `addComplexEdge()`~~
- ~~Update `addDocumentstoNeo4j()` to account for making complexes via `addComplexEdge()`~~
- ~~Add a `participantTypes` field to ALL interaction edges (can be `'noncomplex-to-noncomplex'`, `'complex-to-complex'`, `'noncomplex-to-complex'` or `'complex-to-noncomplex'`)~~
- ~~Update `addDocumentstoNeo4j()` to account for the `participantTypes` field so it can properly handle making INTERACTION edges in all 4 cases (see above)~~

### New Plan:

- [x] Add the following additional fields to edges: `component` (this is an array. [ ] by default), `targetComplex` (null by default), `sourceComplex` (null by default).
- [x] Handle making a complex (special considerations: `id` of the edges will be the factoid UUID for the complex, `type` will be "complex")
- [x] Handle making an interaction between a complex and a non-complex
- [x] Handle making an interaction between two complexes

### DIFF
--- a/src/neo4j/neo4j-document.js
+++ b/src/neo4j/neo4j-document.js
@@ -148,6 +148,22 @@ export async function addDocumentToNeo4j(doc) {
         const sourceComplex = doc.get(sourceUUId);
         const targetComplex = doc.get(targetUUId);
 
+        for (let i = 0; i < sourceComplex.elements().length - 1; i++) {
+          for (let j = 0; j < targetComplex.elements().length - 1; j++) {
+            const complexElementSourceUUId = sourceComplex.elements()[i].id();
+            const targetElementSourceUUId = targetComplex.elements()[i].id();
+            const edgeInfo = {
+              id: e.id(),
+              type: e.type(),
+              component: [],
+              sourceId: convertUUIDtoId(doc, complexElementSourceUUId),
+              targetId: convertUUIDtoId(doc, targetElementSourceUUId),
+              sourceComplex: sourceComplex.id(),
+              targetComplex: targetComplex.id()
+            };
+            arrEdges.push(edgeInfo);
+          }
+        }
       }
       arrEdges.push(edgeInfo);
     }
--- a/src/neo4j/neo4j-document.js
+++ b/src/neo4j/neo4j-document.js
@@ -38,9 +38,9 @@ function complexEdgesMaker(i, doc, complex, arrOfEdges) {
 
   const component = makeComponent(complex, doc);
 
-  while (i < complex.elements().length - 2) {
+  for (let j = i; j < complex.elements().length - 1; j++) {
     const sourceUUId = complex.elements()[i].id();
-    const targetUUId = complex.elements()[i + 1].id();
+    const targetUUId = complex.elements()[j + 1].id();
     const edgeInfo = {
       id: complex.id(),
       type: complex.type(),
--- a/src/neo4j/neo4j-document.js
+++ b/src/neo4j/neo4j-document.js
@@ -48,7 +48,7 @@ export async function addDocumentToNeo4j(doc) {
       // Untested
       const complex = e;
       const component = makeComponent(complex, doc);
-      for (let i = 0; i < complex.elements().length - 2; i++) {
+      for (let i = 0; i < complex.elements().length - 1; i++) {
         for (let j = i + 1; j < complex.elements().length - 1; j++) {
           const sourceUUId = complex.elements()[i].id();
           const targetUUId = complex.elements()[j].id();
--- a/src/neo4j/neo4j-document.js
+++ b/src/neo4j/neo4j-document.js
@@ -23,40 +23,6 @@ function makeComponent(complex, doc) {
 }
 
 /**
- * This is a recursive function
- * @param { Int } i 
- * @param { Document } doc 
- * @param { Document element } complex 
- * @param { Array } arrOfEdges 
- * @returns an array with all the info for the edges needed to represent complex
- */
-function complexEdgesMaker(i, doc, complex, arrOfEdges) {
-  if (i == complex.elements().length - 1) {
-    return arrOfEdges;
-  }
-
-  const component = makeComponent(complex, doc);
-
-  for (let j = i + 1; j < complex.elements().length - 1; j++) {
-    const sourceUUId = complex.elements()[i].id();
-    const targetUUId = complex.elements()[j].id();
-    const edgeInfo = {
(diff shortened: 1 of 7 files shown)


lindajiawenli:
Note to reviewers:

The test files `add-nodes-edges.js` and `doc-tests.js` have been modified to include the new fields of `component`, `sourceComplex` and `targetComplex`. 

I also changed a lot of lines to be .to.equal instead of just .equal because I think that's better form.

There's not much else that is different in those files.

src/neo4j/neo4j-document.js
```diff
@@ -29,12 +37,33 @@ export async function addDocumentToNeo4j(doc) {
...
         name: e.association().name
       };
       arrNodes.push(nodeInfo);
+    } else if (e.isEntity && e.isComplex()) {
+      // TODO: Push on the edges that will represent this complex e in arrEdges
+      // Untested
+      const complex = e;
+      const component = makeComponent(complex, doc);
+      for (let i = 0; i < complex.elements.length - 2; i++) {
+        for (let j = i + 1; j < complex.elements().length - 1; j++) {
```
jvwong, line 52: Have `.elements` and `.elements()`. Maybe the same?

src/neo4j/neo4j-document.js
```diff
@@ -29,12 +37,33 @@ export async function addDocumentToNeo4j(doc) {
...
         id: `${e.association().dbPrefix}:${e.association().id}`,
         name: e.association().name
       };
       arrNodes.push(nodeInfo);
+    } else if (e.isEntity && e.isComplex()) {
+      // TODO: Push on the edges that will represent this complex e in arrEdges
+      // Untested
+      const complex = e;
+      const component = makeComponent(complex, doc);
+      for (let i = 0; i < complex.elements.length - 2; i++) {
```
jvwong, line 51: Is this creating the in-complex edges? If so will these indices work?

jvwong:
Also, does the mapping from factoid to neo4j capture the kind of edge (aka `group` in `entries` of an `element`)?

src/neo4j/neo4j-document.js
```diff
@@ -29,12 +37,33 @@ export async function addDocumentToNeo4j(doc) {
...
+        for (let j = i + 1; j < complex.elements().length - 1; j++) {
+          const sourceUUId = complex.elements()[i].id();
+          const targetUUId = complex.elements()[j].id();
+          const edgeInfo = {
+            id: complex.id(),
+            type: 'binding',
+            component: component,
+            sourceId: convertUUIDtoId(doc, sourceUUId),
+            targetId: convertUUIDtoId(doc, targetUUId),
+            sourceComplex: 'N/A',
```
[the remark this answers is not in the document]
lindajiawenli: Yeah, it's an issue with neo4j- if it's null, the property just never gets created and my tests fail (though I could just delete those cases)
Screen Shot 2023-03-22 at 1 46 16 PM

src/neo4j/neo4j-document.js
lindajiawenli: No you're right- that's a careless typo, it should be .elements(), I'll Cmd-F through the file to clean up those mistakes

src/neo4j/neo4j-document.js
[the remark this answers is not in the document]
jvwong: Is it ok to use empty string? That's falsy in JS

lindajiawenli:
@jvwong 
> Also, does the mapping from factoid to neo4j capture the kind of edge (aka `group` in `entries` of an `element`)?

Actually- no, I don't think I have a property for that yet! I think I originally wanted to replace the INTERACTION edge label with ACTIVATION or similar, but then I forgot about it. Should I implement it that way or should I just add another property? If so, what should I call it?

src/neo4j/neo4j-document.js
lindajiawenli: Yes- so complex has a field of entries. Suppose `entries: [A, B, C, D]`. 

We take the first entry, `i=0 entries[ i ] = A`. We make an edge between `A` and entry that follows it in `entries` (this is what the inner loop/the`j` index is for). So `A` is connected to `B`, `C`, `D`.
After that is done, increment `i, i=1` and do the same for `B`. B is connected to `C`, `D` (and from the previous iteration, `A` is also connected to `B`).

By the time we get to D, the last element of entries, we know that `A`, `B` and `C` are already connected to `D` so there is nothing to do.

I could also just let the outer loop's conditional be `complex.elements().length - 1` to make things simpler. The inner loop's conditional won't be satisfied anyway. 

What do you think?

jvwong:
> @jvwong
> 
> > Also, does the mapping from factoid to neo4j capture the kind of edge (aka `group` in `entries` of an `element`)?
> 
> Actually- no, I don't think I have a property for that yet! I think I originally wanted to replace the `INTERACTION` edge label with `ACTIVATION` or `INHIBIT` or something similar, but then I forgot about it. Should I implement it that way or should I just add another property? If so, what should I call it?

Just curious - just leave it for now and note it somewhere...

src/neo4j/neo4j-document.js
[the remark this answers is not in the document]
lindajiawenli: Yeah, empty string works. Do you prefer `''` or `""`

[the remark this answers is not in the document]
jvwong: Single quotes is good.

src/neo4j/neo4j-document.js
lindajiawenli: I have altered `line 51`'s condition to be `complex.elements().length - 1`

lindajiawenli: Note to self: the `complex.elements().length - 2` condition would not work (because of case where there are 2 nodes- both loops would never run in that case). 
It must be `complex.elements().length - 1`

Thank you for catching that error @jvwong!

Issue #1153, opened 2023-03-27
Neo4j: Develop more search functions, filters


lindajiawenli:
### Currently:
We have one search function called `neighbourhood( id )` that takes the `id` of an entity (ex. `'ncbigene:207'`) as a parameter and returns `null` if the entity does not exist in the database, otherwise it will return an object with two fields: 
- list of all edges connected to that entity
- list all entities that are connected to that entity via one edge

### Goals:
- [x] A search function like `neighbourhood` that filters out all complexes from results
- [ ] `get( id )` where uuid is the factoid document uuid. And this will return an object that contains two fields for the nodes/edges relevant to that document. In the Neo4j Browser (`localhost:7687`) this would show the desired graph in ball-and-stick form.

~~A search function that returns top 10, 20, etc. neighbouring nodes based on how much 'weight' a node has. This might be called something like topNeighbouringNodes( id, lim )~~

@jvwong @maxkfranz Are there any other search functions or filters you think would be useful to the user? Or let me know if neighbourhood is enough and I should move on to something else

jvwong:
Just thought of one more: Get a graph for a particular document

lindajiawenli:
That sounds good! I'll work on that now. I'm thinking of something like:

`findGraphbyUUID( uuid )` where uuid is the factoid document uuid. And this will return an object that contains two fields for the nodes/edges relevant to that document. In the Neo4j Browser (`localhost:7687`) this would show the desired graph in ball-and-stick form. 

Does that sound good?

jvwong:
> `findGraphbyUUID( uuid )` where uuid is the factoid document uuid. And this will return an object that contains two fields for the nodes/edges relevant to that document. In the Neo4j Browser (`localhost:7687`) this would show the desired graph in ball-and-stick form.
> 
> Does that sound good?

Or just `get( id )`

lindajiawenli:
Sure, should I continue on https://github.com/PathwayCommons/factoid/pull/1154 or make a new branch/pr?

jvwong:
> Sure, should I continue on #1154 or make a new branch/pr?

Merge that other PR in if you're happy then start a new PR so they don't interfere.

PR #1154, opened 2023-03-27
Neo4j: Complex filter for neighbourhood()

lindajiawenli:
#1153 

### Currently:

We have one search function called `neighbourhood( id )` that takes the `id` of an entity (ex. 'ncbigene:207') as a parameter and returns `null` if the entity does not exist in the database, otherwise it will return an object with two fields:

- list of all edges connected to that entity
- list all entities that are connected to that entity via one edge

### Task:
Make a function like `neighbourhood`, except it has an extra parameter, `includeComplexes`. If `includeComplexes` is  true, return as usual. Otherwise, do NOT include: 
- edges that represent complexes
- neighbouring nodes that are only neighbours because they are apart of a complex
- interactions that would be returned because the node that is being searched is apart of a complex

### DIFF
--- a/neo4j-test/doc-tests-complex.js
+++ b/neo4j-test/doc-tests-complex.js
@@ -6,6 +6,7 @@ import { loadDoc } from '../src/server/routes/api/document/index.js';
 import { initDriver, closeDriver } from '../src/neo4j/neo4j-driver.js';
 import { addDocumentToNeo4j, convertUUIDtoId } from '../src/neo4j/neo4j-document.js';
 import { deleteAllNodesAndEdges, getGeneName, getNumNodes, getNumEdges, getEdge, getEdgeByIdAndEndpoints } from '../src/neo4j/test-functions.js';
+import { neighbourhoodWithoutComplexes } from '../src/neo4j/neo4j-functions.js';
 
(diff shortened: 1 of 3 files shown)


lindajiawenli:
### Test Cases:

1. [Complex with 2 nodes and nothing else](https://biofactoid.org/document/71478983-b540-43b3-a552-2085dd6b0659). Result should be `null`
2. [Complex with 3 nodes, 3 edges between them](https://biofactoid.org/document/e3ad1fe0-7cc9-4075-b9a7-32626ce0c2ff)
3. [Complex with 2 nodes, one interacting with one non-complex](https://biofactoid.org/document/c164dd79-2351-4a88-b7f4-7a48245735ef)
4. [Two complexes with 2 nodes each, individual nodes interacting with another node from a different complex](https://biofactoid.org/document/8d0977b5-8839-4383-ab51-5198369d3d72) 
5. [Complex interacting with a single non-complex](https://biofactoid.org/document/4638c40d-200d-4fa6-906b-2f567a3cbe06). Result should be `null`
6. [2 complexes interacting with each other](https://biofactoid.org/document/27a7e755-3b47-44aa-96e1-19e4858f395c)

Issue #1163, opened 2023-04-04
Neo4j: Add helper functions for search functions


lindajiawenli:
### Currently:

The main search function, neighbourhood( id ), returns an array of 'records', each record has three values, `n`, `m` which are nodes, and `r` which is the edge. To make things simpler for testing purposes, I created 2 associated functions, `getNeighbouringNodes( id )` and `getInteractions( id )`. 

`getNeighbouringNodes( id )` returns the `m` values (without duplicates).
`getInteractions( id )` returns the `r` values. 
The `n` values are all the same (the node whose `id` we searched) so we disregard. 

But `neighbourhoodWithoutComplexes` and `get` currently do not not have any such helper functions.

### Question:

Should I make the equivalent functions for `neighbourhoodWithoutComplexes` and `get` to make my test suites more readable?

@jvwong

jvwong:
> ### Question:
> Should I make the equivalent functions for `neighbourhoodWithoutComplexes` and `get` to make my test suites more readable?

I feel like `neighbourhoodWithoutComplexes` could be merged with `neighbourhood`, with the default being return everything. If `get` returns nodes and edges I think that's fine.

PR #1165, opened 2023-04-05
Neo4j: Update the search functions

lindajiawenli:
#1163 

- [x] Merge `neighbourhood` and `neighbourhoodWithoutComplexes`
- [x] Make `neighbourhoodReadable`, a function that wraps around `neighbourhood` that returns data in a more user-friendly way (that is, return object with 2 fields: neighbouringNodes and edges)
- [x] Update `get` to return object with 2 fields: nodes and edges, instead of an array of objects that has three fields

### DIFF
--- a/src/neo4j/neo4j-functions.js
+++ b/src/neo4j/neo4j-functions.js
@@ -72,14 +72,18 @@ export async function addEdge(id, type, component, sourceId, targetId, sourceCom
  * @returns null if node not in database, array of objects with
  * fields for the nodes and edges otherwise
  */
-export async function neighbourhood(id) {
+export async function neighbourhood(id, withComplexes = true) {
     const driver = getDriver();
     let session;
     let record;
     try {
         session = driver.session({ database: dbName });
         let result = await session.executeRead(tx => {
-            return tx.run(giveConnectedInfoByGeneId, { id: id });
+            if (withComplexes) {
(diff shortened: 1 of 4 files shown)


src/neo4j/neo4j-functions.js
```diff
@@ -94,59 +99,58 @@ export async function neighbourhood(id) {
...
+ * @param {*} id in the form of "dbName:dbId", ex: "ncbigene:207"
+ * @param { Boolean } withComplexes default true to show complexes
+ * @returns null if id not found in database, object with 2 fields (one for neighbouring nodes and one for
+ * edges) otherwise
+ */
+export async function neighbourhoodReadable(id, withComplexes = true) {
+    let record = await neighbourhood(id, withComplexes);
+    if (record) {
+        return {
+            neighbouringNodes: _.uniqBy(record.map(row => { return row.get('m').properties; }), edge => edge.id),
```
jvwong, line 140: Why not just `nodes`?

jvwong (approved):

src/neo4j/neo4j-functions.js
lindajiawenli: I was probably thinking about my helper function, `getNeighbouringNodes`, but `nodes` make more sense. I'll go ahead and change it

PR #1169, opened 2023-04-06
neo4j driver refactoring

jvwong:
A couple of observations and cosmetic changes:
  - `guaranteeSession` convenience func so you don't have to think about initializing, config etc
  -  The default community edition database is `neo4j`  so set that in config. You can't create one in the driver and I think the CE only lets you access one database anyways.


lindajiawenli (approved):
Looks good to me, `guaranteeSession` is a nice addition, thank you!

I will say that I'm thinking about changing `initDriver()` to `initNeo4jDriver()` and etc. in the future, because sometimes it's hard to tell what driver exactly is being initialized. But I can save that for the last two weeks of my work term

PR #1170, opened 2023-04-06
Submit documents to neo4j on public

jvwong:
The addition of a document to the neo4j database. 
Here, Rethinkdb watches the `document` table and when document `status` gets set to `public` it calls the neo4j add documnent code.

---

@lindajiawenli Seeing weird issue: When I draw and submit a new document via the Biofactoid user interface, the nodes and edges in neo4j are duplicated.

Screen Shot 2023-04-06 at 2 04 11 PM

Screen Shot 2023-04-06 at 1 58 35 PM

Q: If  I call `addDocumentToneo4j` twice with the same doc instance, would you expect the edges and nodes to be added once? In the tests it only adds them once.

### DIFF
--- a/neo4j-test/add-nodes-edges.js
+++ b/neo4j-test/add-nodes-edges.js
@@ -3,7 +3,7 @@ import { initDriver, closeDriver } from '../src/neo4j/neo4j-driver.js';
 import { addEdge, addNode, getInteractions, getNeighbouringNodes, neighbourhood } from '../src/neo4j/neo4j-functions';
 import { deleteAllNodesAndEdges, getGeneName, getNumNodes, getNumEdges, getEdge } from '../src/neo4j/test-functions.js';
 
-describe('02. Tests for addNode, addEdge and seachByGeneId', function () {
+describe('02. Tests for addNode, addEdge and neighbourhood', function () {
 
   before('Should create a driver instance and connect to server', async function () {
     await initDriver();
@@ -116,7 +116,7 @@ describe('02. Tests for addNode, addEdge and seachByGeneId', function () {
     expect(edge.properties.articleTitle).to.equal('MAPK6-AKT signaling promotes tumor growth and resistance to mTOR kinase blockade.');
   });
 
-  it('Ensure searchGeneById works as expected for MAPK6', async function () {
+  it('Ensure neighbourhood works as expected for MAPK6', async function () {
     await addNode('ncbigene:207', 'AKT');
     await addNode('ncbigene:5597', 'MAPK6');
     await addEdge('01ef22cc-2a8e-46d4-9060-6bf1c273869b',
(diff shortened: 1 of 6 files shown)


lindajiawenli:
I tried adjusting the tests so that `addDocumentToNeo4j` is called twice, or even 3 or 4 times and it seems to be okay (code confirms the correct number of nodes/edges are added, and I verified it visually on Neo4j Browser).

The cypher queries that `addNode` and `addEdge` are meant to safeguard against duplicates (I have tests for that in `neo4j-test/add-nodes-edges.js`.

I will take a closer look to see what's going on!

lindajiawenli:
I'm seeing the same issue too
Screen Shot 2023-04-06 at 4 21 58 PM

jvwong:
https://neo4j.com/docs/cypher-manual/current/clauses/merge/#query-merge-using-unique-constraints?

lindajiawenli:
> https://neo4j.com/docs/cypher-manual/current/clauses/merge/#query-merge-using-unique-constraints?

I will try calling
```
CREATE CONSTRAINT FOR (n:Entity) REQUIRE n.id IS UNIQUE;
```

Before the database writes, however I am wondering how this test passes without it:

Screen Shot 2023-04-06 at 4 58 08 PM

Also the ids of edges are not unique in the case of complexes, so I will also need to think about that before creating a constraint

jvwong:
Something else is up:

(a) When I create a breakpoint when `addDocumentToNeo4j` is called (`graphdb.js`) and step through the debugger it creates the correct graph
(b) When I just run the code as is it duplicates the nodes and edges

jvwong:
That's what i was suspecting - do you know why the tests don't seem to reproduce this? I just put two function calls in a test

lindajiawenli:
My theory is that when you call 
```
await addDocumentToNeo4j( doc )
await addDocumentToNeo4j( doc )
```
one after another, it works as expected because the first call finishes, and then the second call finishes and the merge statements handle duplication.

But when it runs after the document becomes 'public', they run at the same time/write to the database at the exact same time(?)

lindajiawenli:
I would like to write tests for this behaviour, but I'm not sure how to replicate it.

Right now I'm:

1. going to my Biofactoid `localhost:3000`
2. manually making a 'new article`
3. submitting it and refreshing the webpage so I see it appear in the carousel on the homepage
4. navigate to my Neo4j Browser (`localhost:7687`) and run `MATCH (n) RETURN n` to see what happened

lindajiawenli (approved):
Seems to be working now! I've tried a couple other examples (with complexes as well) and I haven't been seeing the duplications

jvwong:
> Seems to be working now! I've tried a couple other examples (with complexes as well) and I haven't been seeing the duplications

OK - perhaps the rethinkdb is emitting some (duplicate) events simultaneously.

Issue #1156, opened 2023-03-29
Neo4j: addAllDocumentsToNeo4j


lindajiawenli:
### Currently:
The function `addDocumentToNeo4j( doc )` takes in one document instance `(doc)` and makes the appropriate nodes and edges (works for complexes as well now) in a Neo4j database.

It is currently used in test suites, with a hard-coded RDB fixture called `factoid-neo4j-test`

### Goal:
Create a function `addAllDocumentsToNeo4j()` that makes the appropriate nodes and edges in Neo4j for *ALL the documents in the "real" `factoid` database. 

*with appropriate filters so that only public documents are used. Do not put demos, drafts, etc. into Neo4j.

lindajiawenli:
> add all (public) docs: cron job could initiate (nightly) 
>add one: when an author hits "submit" on their document

Oh, I am not sure how to do that. That sounds more practical than what I have now though, which is just adding all public documents to neo4j when `node -r esm src/neo4j/add-all-docs.js` is run from the command line

lindajiawenli:
Ok, sounds good to me. At least we've discovered that there aren't any errors (that I can immediately see) that come from adding all 194 public documents to Neo4j

jvwong:
There's an error thrown on restoring an empty database (which only ever happens in dev mode):

1. clear the database
2. restore db from file

```
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

Error: No response from database for ID c34266eb-0e76-4b8e-a91d-47fcb77bc03e
```

jvwong:
The above bug arises because the restore adds to `document` and `element` tables, but of course, not all at once. So when a document is added, it looks for the associated data in other tables, which haven't been added yet (this is also a hidden bug of the database dump, but its set on a time delay).

jvwong:
I still think it would be nice to "add all" under the following circumstances:

- [ ] Refresh at regular interval (e.g. nightly)
- [ ] Populate graphDB when a we spin up a new factoid instance (for convenience)

PR #1172, opened 2023-04-11
Refactor: To Neo4j on submit

jvwong:
This version does not listen to events around restoring data to Biofactoid database. 
Also, swallow errors so the factoid app doesn't get killed.

Refs https://github.com/PathwayCommons/factoid/issues/1156#issuecomment-1503474549

### DIFF
--- a/src/server/routes/api/document/graphdb.js
+++ b/src/server/routes/api/document/graphdb.js
@@ -12,16 +12,15 @@ const handleDocChange = async (err, item) => {
     return;
   }
 
-  const new_val = _.get( item, ['new_val'] );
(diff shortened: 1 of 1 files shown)


lindajiawenli (approved):
I have tested this with three different documents (2 of which contain complexes) and it still works as desired in that sense.

From #1156: 
>The above bug arises because the restore adds to document and element tables, but of course, not all at once. So when a document is added, it looks for the associated data in other tables, which haven't been added yet (this is also a hidden bug of the database dump, but its set on a time delay).

I'm not entirely sure what the 'restore' is

PR #1175, opened 2023-04-13
Refresh Graph DB data

jvwong:
This feature enables factoid to trigger refresh (and initial restore) of the data into Neo4j instance. All graph db elements are wiped prior to restore. There is a configurable delay on the period between successive refreshes. 

- `config.js` variables of interest:
  - `CRON_SCHEDULE: 0 * * * *`:  The "global" cron clock (set to run every hour)
  - `GRAPHDB_CRON_REFRESH_PERIOD_MINUTES: 1440`: The minimum time between Graph DB refreshes (daily)
  - `LOG_LEVEL: debug` 

Refs #1156

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -39,12 +39,12 @@ export const APPSIGNAL_APP_NAME = env('APPSIGNAL_APP_NAME', 'Biofactoid');
 export const APPSIGNAL_APP_ENV = env('APPSIGNAL_APP_ENV', 'development');
 
 // CRON jobs
-export const CRON_SCHEDULE = env('CRON_SCHEDULE', '0 1 * * *');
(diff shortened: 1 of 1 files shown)


jvwong:
One way to test this is to run:

```sh
$ LOG_LEVEL="debug" CRON_SCHEDULE="* * * * *"  GRAPHDB_CRON_REFRESH_PERIOD_MINUTES="5" npm run watch
```

Which will ask the cron to fire an event every minute, and the refresh to wait 5 minutes between refreshes

lindajiawenli (approved):
Refreshes perfectly for me! Also, good move putting `deleteAllNodesAndEdges` into `index` and `neo4j-functions.js`

Issue #1173, opened 2023-04-13
Neo4j: Put apoc.lock on relationships


lindajiawenli:
Currently, there is a lock (`apoc.lock.nodes`) on all nodes of `id` when `addNode( id )` is called.

Add a lock on relationships that have endpoints `sourceId` and `targetId` as well, for when `addEdge` is called.
This will guard against the duplication of edges/relationships (@jvwong once briefly observed this, so it is best to be careful).

See https://neo4j.com/labs/apoc/4.1/graph-updates/locking/

PR #1176, opened 2023-04-13
Neo4j: Lock on edges to prevent duplication

lindajiawenli:
#1173 

- [x] Place apoc.lock.rels on relationships before calling `addEdge`/`addDocumentToNeo4j`
- ~~Place constraint on relationships as well for extra security~~

The goal is to prevent duplicate edges that may occur (as MERGE is not enough due to the possibility that the add functions run in parallel with each other/at the same time)


lindajiawenli:
Edit: It appears that asserting that node ids are unique is available for non-Enterprise/paid users, but other constraints are not. Additionally, constraints on relationships are currently limited to making sure a relationship has X property

Source: https://neo4j.com/docs/cypher-manual/current/constraints/

lindajiawenli:
My testing seems fine, but then again I didn't experience any edge duplication after I locked/added the constraint on the nodes

jvwong (approved):

Issue #1151, opened 2023-03-23
Image endpoint: Avoid checking updates per request


jvwong:
A reasonable estimate is that 99.9% of the time, submitted pathways / documents are never updated by the author after the fact. A very minor case (handful) is when it is updated, by us during post-curation or the author does do it post-submission. 

This suggests that we can prioritize the case where the network image stays the same on submit. Importantly, checking the time of last update on each document on image fetch is unnecessary and introduces a performance burden on each webpage load (home, search).

---

_Originally posted by @jvwong in https://github.com/PathwayCommons/factoid/issues/1143#issuecomment-1480158389_

> I'm finding the `getDoc` call to be a major offender following an initial load:
> 
> https://github.com/PathwayCommons/factoid/blob/eca68a131632c259a52f895741a9f8b406755cbb/src/server/routes/api/document/index.js#L1453-L1466
>
>Alternative to handle `imgCache` staleness is:
> - Fetch the JSON results from `/api/document/` (ttl is 24 hours)
>  - Find the matching document and it's `lastEditedDate`
>  - Hit the cache or regenerate as usual
>
>So the downside is authors won't see updates to their submitted docs within 24 hours. The p-limit code is still valid on initial load, the rest of the time the entire search browse page loads almost instantly.
>

maxkfranz:
Other notes:

- If the docs are rarely changed, then don't even bother with cache invalidation.  Even when we do correct an author's document, do we change it visually?  It's probably just the grounding or interaction type, which doesn't make a visual change.
- Store images in the DB.
  - Pro: Simple.
  - Pro: Would make social cards faster and more reliable.  I think Twitter may bail out if the image loads too slowly.
  - Con: Bigger DB.

jvwong:
> Other notes:
> 
> * If the docs are rarely changed, then don't even bother with cache invalidation.  Even when we do correct an author's document, do we change it visually?  It's probably just the grounding or interaction type, which doesn't make a visual change.

For whatever reason, the cache is returning true on every `has(id)`, regardless of the ttl, which in fact, is being set as expected. This means it behaves like you said above.

PR #1179, opened 2023-04-24
Use positional argument 'maxAge' for cache

jvwong:
To set an expiration time for a key/value in lru-cache, use the positional argument `maxAge`. 

Problem was I was viewing the most updated API docs (9.1) rather than our [installed version (4.1.1)](https://github.com/isaacs/node-lru-cache/releases/tag/v4.1.1)

Refs: https://github.com/PathwayCommons/factoid/issues/1151#issuecomment-1503363300

Relevant sections of 4.1.1 Docs:

> ## Options
> ...
> * `maxAge` Maximum age in ms.  Items are not pro-actively pruned out
  as they age, but if you try to get an item that is too old, it'll
  drop it and return undefined instead of giving it to you.
>
> ## API
>
> * `set(key, value, maxAge)`

### DIFF
--- a/src/server/routes/api/document/index.js
+++ b/src/server/routes/api/document/index.js
@@ -1460,9 +1460,9 @@ http.get('/(:id).png', function( req, res, next ){
   const fillCache = async doc => {
     const img = await getDocumentImageBufferRatedLimited(doc);
     const cache = { img };
-    const ttl = calcTtl(doc);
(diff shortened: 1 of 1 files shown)


maxkfranz:
I’ll take a look at the PRs tomorrow

maxkfranz:
Simple change so merging this in...

PR #1180, opened 2023-04-25
Support an 'entity' type 

jvwong:
Support grounding to the generic `entity` type.

Related to grounding search support for Famplex (families and complexes):
   - [PR: Support FamPlex](https://github.com/PathwayCommons/grounding-search/pull/131)
   - [Issue: Expand entity support: complexes & families](https://github.com/PathwayCommons/grounding-search/issues/130)

### DIFF
--- a/src/client/components/element-info/entity-assoc-display.js
+++ b/src/client/components/element-info/entity-assoc-display.js
@@ -33,6 +33,27 @@ let protein = (m, searchTerms, includeOrganism = true) => {
   ];
 };
 
+let entity = (m, searchTerms, includeOrganism = true) => {
+  return [
+    includeOrganism ? h('div.entity-info-section', [
+      h('span.entity-info-title', 'Organism'),
+      h('span', m.organismName)
+    ]) : null,
+    m.summary ?
+      h('div.entity-info-section', [
+        h('span.entity-info-title', 'Summary'),
+        h('span', m.summary)
+      ]): null,
+    h('div.entity-info-section', !m.shortSynonyms ? [] : [
+      h('span.entity-info-title', 'Synonyms'),
+      ...m.shortSynonyms.map( name => h('span.entity-info-alt-name', [
+        h(Highlighter, { text: name, terms: searchTerms })
+      ])),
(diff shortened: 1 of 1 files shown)


maxkfranz:
@jvwong, is there a particular reason to use a general entity type for this as opposed to something more specific, like 'complex'?

Families feels a bit generic, but on the other hand, we could have two specific types (i.e. 'family' & 'complex').  In any case, we'll have to live with whatever decision we make for a long time.  Are there BioPAX conversion advantages and disadvantages to consider for the generic approach versus the two-type-specific approach?

What do you think?

jvwong:
> @jvwong, is there a particular reason to use a general entity type for this as opposed to something more specific, like 'complex'?
> 
> Families feels a bit generic, but on the other hand, we could have two specific types (i.e. 'family' & 'complex'). In any case, we'll have to live with whatever decision we make for a long time. Are there BioPAX conversion advantages and disadvantages to consider for the generic approach versus the two-type-specific approach?
> 
> What do you think?

yeah good question. There's a few issues to sort out: 

1. What's a family vs a complex? Indeed, you can have a family of complexes (e.g NF-κB, Figure 1). Also, it's completely reasonable to have a complex consist of one or more components that are themselves family members (TGF-beta receptor is a dimer of two families of proteins). 
  - Ben and I talked about this, there *could* be a way to process famplex data to figure out what the family members are (i.e. a `relations.csv` file has various `is-a` and `partof` between entities); the remaining would be complexes. **I think this is worth looking into in the grounding-search PR.**
  
2. Models. Factoid (and likely BioPAX) are modelled such that complexes are "component-centric", that is , the individual components define the complex. Indeed, the rationale behind Famplex is resources didn't have a good system for grounding complexes *per se* (the parent). Consequently, my guess is that a lot of the code will try to recurse into a complex. 

---

 **Figure 1. NF-κB family members.** 

'NF-κB' is a dimeric (two component) protein complex. 5 unique components (genes) can be present to generate (at least) 11 distinct NF-κB forms.

jvwong:
I've figured out a way to set famplex items to either `complex` or `family`.

Issue #1181, opened 2023-04-25
Supporting Complexes and Family members via Famplex


jvwong:
To include [grounding support for Famplex complexes and family members in grounding-search]() may involve updates herein. In particular:

- Curation tool support for Famplex types:
  1. [x]  `protein` (aka 'family')
  2. [x] ~~`complex`~~ `namedComplex`
- Updates to the BioPAX export  to handle the  Famplex entity types
  1. [x] `protein` (aka 'family') 
  2. [x] *~~`complex`~~ `namedComplex`

- *BioPAX and Complexes:
  - BioPAX L3 does not specify an EntityReference for a complex, merely it's components
  - A Complex may have 0 components aka "black-box"
  - A complex may declare an Xref (e.g. UnificationXref)

jvwong:
Some more considerations around FamPlex complexes:

- Type 
  - A Biofactoid `Complex`  is created within the editor via drag and drop (top-down)
  - No (clear) way to swap a Biofactoid `Entity` for a `Complex` when a grounding search top hit is a FamPlex 'complex' (bottom-up). 
    - Although, the latter would be more extensible (e.g. adding GO terms to grounding search ) 
- Nesting
  - Should we (dis)allow a (compound) complex to have a (FamPlex) complex entity inside it?

PR #1187, opened 2023-06-19
Famplex: Named complex

jvwong:
Support the entity type `namedComplex`, which represent FamPlex entities with components. In the curation tool, entity information for named complexes resembles existing (compound node) complexes and do not display a reconfigurable type or organism. Instead, a user must simply select another record. 

Representative examples:

Screen Shot 2023-06-19 at 4 00 25 PM
Screen Shot 2023-06-19 at 4 04 32 PM

- More items to consider 
  - [x] **gene/Famplex clashes, refinement**
  - [x] search template
  - [x] BioPAX/SBGN template
  - [x] factoid-converters
  - [x] neo4j mapping
  
Refs #1181, https://github.com/PathwayCommons/grounding-search/issues/130, https://github.com/PathwayCommons/grounding-search/pull/141

### DIFF
--- a/src/client/components/element-info/entity-assoc-display.js
+++ b/src/client/components/element-info/entity-assoc-display.js
@@ -7,6 +7,11 @@ const { IDENTIFIERS_ORG_ID_BASE_URL } = require('../../../config');
 
 let protein = (m, searchTerms, includeOrganism = true) => {
   return [
+    m.summary ?
+    h('div.entity-info-section', [
+      h('span.entity-info-title', 'Summary'),
+      h('span', m.summary)
+    ]): null,
     includeOrganism ? h('div.entity-info-section', [
       h('span.entity-info-title', 'Organism'),
       h('span', m.organismName)
@@ -36,6 +41,7 @@ let protein = (m, searchTerms, includeOrganism = true) => {
 let ggp = protein;
 let dna = protein;
 let rna = protein;
+let namedComplex = protein;
 
 let chemical = (m, searchTerms) => {
   return [
@@ -96,6 +102,6 @@ let modification = (mod, onEdit) => h('div.entity-info-section.entity-info-mod-s
   ])
 ]);
 
-export const assocDisp = { ggp, dna, rna, protein, modification, chemical, complex, link };
+export const assocDisp = { ggp, dna, rna, protein, modification, chemical, complex, link, namedComplex };
 
 export default assocDisp;
--- a/src/client/components/element-info/entity-info.js
+++ b/src/client/components/element-info/entity-info.js
@@ -365,10 +365,10 @@ class EntityInfo extends DataComponent {
     let doc = p.document;
     let children = [];
 
-    const citation = doc.citation();
-    const { pmid } = citation;
+    // const citation = doc.citation();
+    // const { pmid } = citation;
 
-    const hasPubmedMetadata = pmid != null;
+    // const hasPubmedMetadata = pmid != null;
 
     let Loader = ({ loading = true }) => h('div.entity-info-matches-loading' + (loading ? '.entity-info-matches-loading-active' : ''), [
       loading ? h('i.icon.icon-spinner') : h('i.material-icons', 'remove')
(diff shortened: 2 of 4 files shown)


maxkfranz:
Other items to consider:  

(1) When we add support for automatic organism hints, it may be sensible to take the hint into account when suggesting Famplex entries.  Whereas previously only NCBI entries would have an organism which could use the hint for weighting, now we also have Famplex entries which are all implicitly human.  How we approach that may take into account how the Famplex entries are structured in the grounding service.

(2) Are there any prominent collisions with NCBI entries.  If so, would the use of the gene sub-selection UI be useful here?

jvwong:
> Other items to consider:
> 
> (1) When we add support for automatic organism hints, it may be sensible to take the hint into account when suggesting Famplex entries. Whereas previously only NCBI entries would have an organism which could use the hint for weighting, now we also have Famplex entries which are all implicitly human. How we approach that may take into account how the Famplex entries are structured in the grounding service.

The Famplex entities (`protein` and `namedComplex`) coming through the grounding-search are explicitly set to have:

```json
...
		"organism": "9606",
		"organismName": "Homo sapiens",
...

```

> 
> (2) Are there any prominent collisions with NCBI entries. If so, would the use of the gene sub-selection UI be useful here?

Good point, I'll add to list. For instance, someone could want to say `NFKappaB` (`namedComplex`) or `NF-kappaB` (`gene`), such as in the following [Sanchez-Lopez et al. Cell Death Differ 27 2020(https://biofactoid.org/document/999205de-8768-483b-9203-6a775f533099):

Screen Shot 2023-06-22 at 11 43 50 AM

jvwong:
> Other items to consider:
> 
> (1) When we add support for automatic organism hints, it may be sensible to take the hint into account when suggesting Famplex entries. Whereas previously only NCBI entries would have an organism which could use the hint for weighting, now we also have Famplex entries which are all implicitly human. How we approach that may take into account how the Famplex entries are structured in the grounding service.

 I realize my description regarding this was unclear. The organism is explicit, it is currently just not 'refinable'.

maxkfranz:
For (1), I think we're in agreement.  The main thing here is that there's an explicit organism field in the data for the hints.

For (2), that may be a good test case.  Another example is ARP.

Screenshot 2023-06-23 at 13 45 22

It may feel more natural to select a new "Family" option after "Protein", rather than hunting under "Which arp" -- where it's not immediately obvious that it's a family.

Similarly, if there's a collision with another type (e.g. drug), then it may be worth considering adding yet another option after "Protein" (e.g. "Chemical").

jvwong:
> For (1), I think we're in agreement. The main thing here is that there's an explicit organism field in the data for the hints.
> 
> For (2), that may be a good test case. Another example is ARP.
> 
> Screenshot 2023-06-23 at 13 45 22
> It may feel more natural to select a new "Family" option after "Protein", rather than hunting under "Which arp" -- where it's not immediately obvious that it's a family.
> 
> Similarly, if there's a collision with another type (e.g. drug), then it may be worth considering adding yet another option after "Protein" (e.g. "Chemical").

I think its worth taking some time to review and clarify (for myself as well) the refinement logic in the entity grounding popup, and when it should apply:
 - Material 'type' selection 
 - Organism selection
 - 'Which {name}' selection

To be continued...

jvwong:
Table I summarizes the entity grounding refinement made available in an entity tooltip within the curation tool. 

The criteria for refinement is a gene or gene product. In practice, this means that any records from NCBI Gene, and `protein` entities from FamPlex (aka a family) are subject to refinement. An example of a FamPlex protein is PKC.

**Table I. Refinement attributes available in entity grounding search UI**
| Property     | Purpose                                    | Restrictions (Domain)               | Range                              | Comments                  |
| ------------ | ------------------------------------------ | ----------------------------------- | ---------------------------------- | ------------------------- |
| Type         | Set material type                          | \-                                  | Gene or product, DNA, RNA, Protein |                           |
| Organism     | Group exact search hit matches by organism for quick access | Entity record with exact match to name or synonym      | Organism                           | Not related to homologues |
| Which {name} | Select exact matches among those from an organism       | Multiple records with exact name or synonym match in a given organism | Grounding record                   |

Notes:
- In the current version, FamPlex `protein` entities (aka family) are refinable.  `complex` and `namedComplex` (from FamPlex)  aren't refinable, as I didn't think it makes sense to do anything at the parent-level. 
- In general, I thought it would be weird to expose to users that an entity is a protein from a family (PKC) vs a singular gene (p53) which is why I don't think we should distinguish the two during sub-selection. I think this is the messy, fluid way that biologists refer to things. I'm not saying its right or ideal.

jvwong:
Latest iteration:

- `namedComplex` are entities with a FamPlex ID representing complexes
- Regarding export (BioPAX or SBGN) , a `namedComplex` is considered a black-box complex
  - Compatible with factoid-converters with minor update (tag complex with UnificationXref for ID)

PR #1200, opened 2023-08-31
CrossRef modules for that support search 

jvwong:
A `works` module exposes `find` that:
 - (a) guesses what the input query is (DOI, text term) 
 - (b) retrieves it from CrossRef web service 
 - (c) makes sure the top match is sane (e.g. substring of title, exact DOI) 
 - (d) the right 'type' (restricted to preprints right now).

The `map` module tries to convert the CrossRef record into a consistent schema that we store - in this case, a 'PubMed' article. 

None of the additions here are called  by the server - will save for next PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ Services:
 - `ORCID_BASE_URL` : url for [ORCID](https://orcid.org/) website
 - `ORCID_PUBLIC_API_BASE_URL` : url for version of [ORCID](https://orcid.org/) public API
 - `NO_ABSTRACT_HANDLING` : labels directing how to sort documents missing query text. 'text' (default): autogenerate text from templates; 'date': sort by date and ignore text.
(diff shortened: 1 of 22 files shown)


maxkfranz:
Looks great.

Are Crossref-originated entries marked somehow (e.g. flag or enum) so that we know they came from Crossref rather than Pubmed?

jvwong:
> Looks great.
> 
> Are Crossref-originated entries marked somehow (e.g. flag or enum) so that we know they came from Crossref rather than Pubmed?

Not right now - it's implicit that if the record has only a DOI it is from CrossRef. Having a PubMed ID doesn't necessarily tell you anything about where it came from, as these records may or may not possess a DOI.

maxkfranz:
Sounds like the options are open in future then. Great

PR #1212, opened 2023-10-12
Extract string tests for items that resemble DOIs and PMIDs

jvwong:
Checking for a DOI-like string comes up more often, especially when extracting from plain text fields.


maxkfranz (approved):
Looks good for Crossref. 

We may have to revisit for other systems in future. For now, good enough.

Issue #1286, opened 2024-07-18
Handle case where no grounding information is returned


jvwong:
### Background
As discussed previously (https://github.com/PathwayCommons/grounding-search/issues/159) there is desire to reduce grounding errors that arise from poor matches. 

This can happen because author node label is:
  1.  an out of scope (aka out of dictionary or 'OOD') concept 
  2. synonym that isn't listed in a entity database record
  3. gibberish 
 
An underlying issue is that the grounding search is configured to always return a list of search hits from which a "match" (i.e. top ranked item) is assigned. Based on previous work (see  https://github.com/PathwayCommons/grounding-search/discussions/162) we can configure Elasticsearch to reduce these errors, returning empty results for many of these cases.

### Problem
A (prior) assumption in Biofactoid is that users understand the scope of curation, and therefore, would only enter a name for an entity that is supported and for which there is a corresponding match.

**a.** Node grounding popup message: The existing warning message for no results (which really only occurred in case of a network error) implies that they should "try again" (i.e. case 2 above, just find the right synonym). This is a problem if the entity doesn't exist (cases 1 also 3). This should be handled more generically.

<img width="334" alt="Screenshot 2024-07-22 at 3 10 18 PM" src="https://github.com/user-attachments/assets/99f111e9-3568-4050-beba-dbbf6e3399c4">

**b.** Ungrounded nodes (that aren't complexes):  When a node is never labelled or not ungrounded (e.g. if no grounding search results are returned) the input label is not shown. The problem with the latter that this could be confusing for a user - "what happened to the name I just typed for this node?" Also, at submit time, the presence of blank nodes presents warnings, but does not block submission. Rather, it blocks the document from becoming "public". This should handle the case more generically.

<img width="245" alt="Screenshot 2024-07-22 at 3 14 38 PM" src="https://github.com/user-attachments/assets/097c9b18-08c2-4376-8ce7-a14fe8c78b41">
    
**c.** Admin: The admin doesn't flag the existence of ungrounded (not complex) nodes in Documents. 

### Implementation

Options for curation tool:

(a) (Blocking) Most users understand grounding: Remind them about scope and suggest they may want to use another synonym. Perhaps they will correct it on their own (best). Downside is that users may ignore this, become confused or frustrated - this tool doesn't "work"

[...]

Issue #1240, opened 2024-01-18
Admin: GET document(s)


jvwong:
The admin page is used to access documents to view and update pathway, article and user information etc. There are a growing  number of documents and currently the admin only allows paging through docs by creation date. The need is to access specific documents by: Document ID and/or the article ID (i.e. DOI, PMID).

PR #1287, opened 2024-07-24
Admin: Flag entity-related issues

jvwong:
Here I augment the admin/model to flag common entity-level issues:
  - ungrounded (non-complex) entities
  - mixed organisms (could signal orthologue issues)

Question: For convenience, I added an entity `normalized` method, but I can't recall what the difference is between `associated` and `completed`?

Refs #1286, part 'c'

### DIFF
--- a/src/model/element/entity.js
+++ b/src/model/element/entity.js
@@ -165,6 +165,25 @@ class Entity extends Element {
     return update;
   }
 
+  normalized(){
+    // normalized := has a name and is associated
+    const name = this.syncher.get('name');
+    return name && this.associated();
+  }
+
+  issues(){
+    const items = [];
+    if( !this.normalized() ){
+      items.push({
+        error: {
+          name: 'Normalization'
+        },
+        message: `Entity with name "${this.name()}" is not normalized`
+      });
+    }
+    return items;
+  }
+
   organism(){
     const assoc = this.association();
 
--- a/src/model/element/entity.js
+++ b/src/model/element/entity.js
@@ -167,7 +167,7 @@ class Entity extends Element {
 
   normalized(){
     // normalized := has a name and is associated
-    const name = this.syncher.get('name');
+    const name = this.name();
     return name && this.associated() && this.completed();
   }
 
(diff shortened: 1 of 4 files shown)


src/model/element/entity.js
```diff
@@ -165,6 +165,25 @@ class Entity extends Element {
     return update;
   }
 
+  normalized(){
+    // normalized := has a name and is associated
+    const name = this.syncher.get('name');
+    return name && this.associated() && this.completed();
```
jvwong, line 171: Not sure what 'completed' was supposed to indicate?

jvwong:
Admin page examples for (top) unnormalized entity (blocks public status) (bottom) mixed organisms (warning)
Screenshot 2024-07-24 at 11 21 47 AM

src/model/element/entity.js
maxkfranz: associated true (entity/node only) => grounded

completed (any element / node / edge) => the user has finished specifying the element

E.g. an interaction doesn't have a grounding but it can still be completed by specifying it as phosphorylation.

maxkfranz (approved):
Looks great.

Suggestion for future:  A filter at the top for these flagged, problematic documents would be nice so you don't have to hunt for them.

I suppose a submitted/not-public would be the same, but it's not as clear.

351774702-e181f6f8-d5ae-41ab-94c7-bf666c3cc6b8

jvwong:
> Looks great.
> 
> Suggestion for future: A filter at the top for these flagged, problematic documents would be nice so you don't have to hunt for them.

Good point, there are some blocked (submitted) contributions that I never reviewed, missed.
 
Add this to a few other admin features required, based on things I'm doing  manually:

- GET Document(s) (#1240) 
  - By ID 
  - Nice to have: By DOI, PMID, email, title etc...
- Download as CSV
  - For re-analysis, QC purposes
  - Visually inspecting Documents is no sufficient to detect a lot of weird errors, bugs
- Filter by flag (issue, warning)
  - This

PR #1288, opened 2024-07-24
More specific message for empty/no match.

jvwong:
When there is no match to a node name, keep the error message but add more detail about model scope. As before, nodes are still left blank and submission will not be blocked but Document will not proceed to public.

I went back and forth on this, but in the end, should rely on user to correct/delete this node. 

Screenshot 2024-07-24 at 2 00 58 PM

Refs #1286


maxkfranz (approved):
Fair enough

PR #1296, opened 2024-08-12
Fetch hints

fileoy:
### Key changes on the hint service:
- PubTator API path is now configurable
- get a BioCDocument from the PubTator
- A simple wrapper to retrieve Hints from PubTator

### DIFF
--- a/test/pubtator/pubtator.js
+++ b/test/pubtator/pubtator.js
@@ -14,6 +14,7 @@ import pubtator_6 from './10.1016_j.molcel.2024.01.007.json';
 import pubtator_7 from './10.15252_embj.2023113616.json';
 import pubtator_8 from './pubtator_8.json';
 import { NCBI_BASE_URL } from '../../src/config.js';
+import { HTTPStatusError } from '../../src/util/fetch.js';
 
 /**
  * bioCDocuments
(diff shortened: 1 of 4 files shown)


test/pubtator/pubtator.js
```diff
@@ -216,4 +218,43 @@ describe('pubtator', function () {
       }); // BioC Document
     }); // forEach
   }); // map
+  
+  describe('get', function(){
+    it('Should be null when no body returned', async function(){
+      let pmid = '38496625';
```
jvwong, line 224: I don't think this test is useful as is. e.g. If you try `39131402` (PMID exists, no pubtator data) the server returns a HTTP status 400, your code will throw an error.

test/pubtator/pubtator.js
```diff
@@ -216,4 +218,43 @@ describe('pubtator', function () {
...
+      expect( bioCDocument ).to.be.null;
+    });
+
+    it('Should exist when a body is returned', async function(){
+      let pmid = '28041912';
+      var path = new RegExp( pmid );
+      nock( NCBI_BASE_URL )
+        .get( path )
+        .reply( 200, { foo: 'bar' } );
+      let bioCDocument = await pubtator.get( pmid );
```
jvwong, line 256: Isn't this testing `get`?

jvwong (changes requested):

test/pubtator/pubtator.js
fileoy: Yes it tests `get` ,  I took it as it is from branch `iss1148_pubtator-service`

jvwong (approved):

PR #1319, opened 2025-01-29
Fix Sciwheel from breaking formatting of entity synonyms

maxkfranz:
With Sciwheel installed, the synonyms are broken.  This patch fixes it.

See screenshot:

Screenshot 2025-01-29 at 16 45 08


jvwong (approved):
Great thanks!

